PR #29, opened 2015-08-28
Add Yahoo provider

jerriep:
Added Yahoo provider. Please note that you will need to follow [this guide](https://www.twilio.com/blog/2014/03/configure-windows-for-local-webhook-testing-using-ngrok.html) to be able to test the Yahoo provider. It is a tricky one...

Oh and the location of the IIS config has changed in VS 2015, so that article is incorrect. It is now in the `.vs` folder of your project, so in other words `AspNet.Security.OAuth.Providers\.vs\config\applicationhost.config`

Because you will get a different URL in ngrok than what I did, you will also need to register own app in the Yahoo developer portal at https://developer.yahoo.com/apps/, and use your ngrok URL as the callback URL.

Even when you did everything correctly, you will most probably still run into strange issues with weird errors. The Yahoo infrastructure seems quite unstable and will give you random errors. Usually just pressing F5 to refresh the page will resolve the errors. Other times you have to press F5 10 times.... :(

Good luck ;)

### DIFF
--- a/src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
@@ -1,0 +1,84 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Security.Claims;
+using System.Text;
+using System.Threading.Tasks;
+using AspNet.Security.OAuth.Extensions;
+using Microsoft.AspNet.Authentication;
+using Microsoft.AspNet.Authentication.OAuth;
+using Microsoft.AspNet.Http.Authentication;
+using Microsoft.Framework.Internal;
+using Newtonsoft.Json.Linq;
+
+namespace AspNet.Security.OAuth.Yahoo {
+    public class YahooAuthenticationHandler : OAuthAuthenticationHandler<YahooAuthenticationOptions> {
+        public YahooAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {}
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+            var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+
+            var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            response.EnsureSuccessStatusCode();
+
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+
+            identity.AddOptionalClaim(ClaimTypes.NameIdentifier, YahooAuthenticationHelper.GetIdentifier(payload), Options.ClaimsIssuer)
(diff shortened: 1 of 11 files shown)


src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            response.EnsureSuccessStatusCode();
+
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+
+            // Extract the profile object from the result payload
+            var profile = payload.Value<JObject>("profile");
+
+            // Extract all the claims from the profile
+            if (profile != null) {
```
PinpointTownes: Have you tried moving this null check directly to the `YahooAuthenticationHelper` class?
Something like:

```csharp
public static string GetFamilyName([NotNull] JObject user) => user.Value<string>("familyName"); 
```

->


```csharp
public static string GetFamilyName([NotNull] JObject user) => user.Value<JObject>("profile")
                                                                 ?.Value<string>("familyName"); 
```

This way, one can use the helper without having to manually extract the profile node first :+1:

src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            return new AuthenticationTicket(
+                notification.Principal, notification.Properties,
+                notification.Options.AuthenticationScheme);
+        }
+
+        protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(string code, string redirectUri) {
+            var tokenRequestParameters = new Dictionary<string, string>() {
+                {"client_id", Options.ClientId},
+                {"redirect_uri", redirectUri},
+                {"client_secret", Options.ClientSecret},
```
PinpointTownes: Sounds like `client_id` and `client secret` are flowed both in the request form and in the `Authorization` header. Do you need both methods?

PinpointTownes:
Woo, really nice, thanks @jerriep! :+1: 
I'll give it a try when I'm back from holidays (Sunday).

Note: CI failed, but I'm pretty sure that's unrelated to your PR.

src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
jerriep: Sure. I was in 2 minds on how to do this. Will change as per your suggestion

src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationHandler.cs
jerriep: Yeah it is a bit confusing. See this:
https://developer.yahoo.com/oauth2/guide/flows_authcode/#step-4-exchange-authorization-code-for-access-token

The state that it should be passed in, but in their example it isn't. Will try it without and see what happens...

jerriep: Turns out you are right. I removed it. Well spotted :)

src/AspNet.Security.OAuth.Yahoo/project.json
```diff
@@ -0,0 +1,20 @@
...
+
+    "dependencies": {
+        "AspNet.Security.OAuth.Extensions": { "type": "build", "version": "1.0.0-*" },
+        "Microsoft.AspNet.Authentication.OAuth": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Internal": { "type": "build", "version": "1.0.0-*" }
+    },
+
+    "frameworks": {
+        "dnx451": { },
+        "dnxcore50": { }
```
PinpointTownes: There's a bug with dependency resolution in the latest DNX. Could you please add this workaround to make CI happy? :thought_balloon: 

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/project.json#L27-L31

Thanks!

jerriep:
All seems ok now, thanks for solution

PinpointTownes:
Merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/237509d99d731c3067913317dfdb236331e309db

Thanks Jerrie :+1:

Issue #2, opened 2015-05-12
Add BattleNet provider


jerriep:

PinpointTownes:
Done: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/c241e55e71f6c07f1f1b2c0bdf728c14cf9b66c4

@Gekctek thank you! :+1:

PR #42, opened 2015-09-10
Battle.net provider

Gekctek:
aspnet-contrib/AspNet.Security.OAuth.Providers#2 Adds Battle.Net provider


samples/Mvc.Client/Startup.cs
```diff
@@ -64,6 +65,13 @@ public class Startup {
                 options.ClientSecret = "bf9bacc22391a01977f8e7d57daadd564809e6a1";
             });
 
+			app.UseBattleNetAuthentication(options =>
```
PinpointTownes: Tabs? :scream:

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.CallbackPath"/>.
+        /// </summary>
+        public const string CallbackPath = "/signin-battlenet";
+
+		/// <summary>
+		/// Default value for <see cref="OAuthAuthenticationOptions.AuthorizationEndpoint"/> for US servers.
+		/// </summary>
+		public const string AuthorizationEndpointUS = "https://us.battle.net/oauth/authorize";
```
PinpointTownes: Please replace tabs with spaces :+1:

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace AspNet.Security.OAuth.BattleNet {
+    public class BattleNetAuthenticationHandler : OAuthAuthenticationHandler<BattleNetAuthenticationOptions> {
+        public BattleNetAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {
+        }
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+
+			string userInfoUri = Options.UserInformationEndpoint + "?access_token=" + tokens.AccessToken;
```
PinpointTownes: Question: do you know if BattleNet supported flowing the access token in the `Authorization` header?

(also, please fix the tabs :smile:)

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+            var request = new HttpRequestMessage(HttpMethod.Get, userInfoUri);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+            var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            response.EnsureSuccessStatusCode();
+
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+
+
+			string id = BattleNetAuthenticationHelper.GetIdentifier(payload);
```
PinpointTownes: Please use `var` or remove these intermediary variables.

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+            CallbackPath = new PathString(BattleNetAuthenticationDefaults.CallbackPath);
+
+			this.SetServerRegion(ServerRegion.US); //Default to US
+			
+            UserInformationEndpoint = BattleNetAuthenticationDefaults.UserInformationEndpoint;
+
+            SaveTokensAsClaims = false;
+        }
+
+		public void SetServerRegion(ServerRegion serverRegion)
```
PinpointTownes: I'm not sure I really like this pattern. I'll think a bit more about it and get back to you ASAP! :+1: 

/cc @jerriep

PinpointTownes:
Thanks for this PR, it looks nice! :+1: 

Before merging, we'll have to fix an issue we're experiencing with our build system. I'm pretty confident it will be fixed tomorrow. In the meantime, feel free to fix the minor code style remarks I mentioned.

/cc @jerriep 

Thanks!

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationHandler.cs
Gekctek: I tried and it didnt work. Also i didnt see it in their docs

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationHandler.cs
Gekctek: Is there a reason that one would use var over explict types besides for line length or dynamic like objects? I get the remove the variable altogether, which ill do, but I usually do it because it is much more readable

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationOptions.cs
Gekctek: The only other way I considered was a method that just returns the server url and you pass in the enum but i figured just having the constants made sense and a optional helper method

jerriep: Yeah, Battle.NET was a nasty looking one last time around as well...

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+            CallbackPath = new PathString(BattleNetAuthenticationDefaults.CallbackPath);
+
+            this.SetServerRegion(ServerRegion.US); //Default to US
+
+            UserInformationEndpoint = BattleNetAuthenticationDefaults.UserInformationEndpoint;
+
+            SaveTokensAsClaims = false;
+        }
+
+        public void SetServerRegion(ServerRegion serverRegion) {
```
jerriep: Would it not be better to have a property called `ServerRegion` and then automatically set the correct endpoints in the setter of that property, rather than having the user call a method?

The normal pattern for setting options is always to just set a property, but now we are deviating from this by having the user calling a method.

Gekctek: I would argue that it would be confusing to have a property change other properties. There is no good way for people to easily see that they would have to either set one property (ServerRegion) or set two properties (the endpoints) and not all 3. We wouldn't want to just leave the one property (ServerRegion) because that wouldnt be consistent with the other ones and takes away some customization.  Also not having a helper method there might be ok so infact to be less confusing and more consistant I can just remove it all together.  I just want a confusing senario where a user is setting 3 properties for no purpose.

jerriep: Yes, you are correct.

PinpointTownes: I agree with @jerriep, a `Region` property is clearly the way to go.

If you really think developers may forget to set this property, leave it to null by default and add a check throwing an exception if it is not set AND if the endpoints have not been explicitly set.

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.TokenEndpoint"/> for CN servers.
+        /// </summary>
+        public const string TokenEndpointCN = "https://www.battlenet.com.cn/oauth/token";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.UserInformationEndpoint"/>.
+        /// </summary>
+        public const string UserInformationEndpoint = "https://us.api.battle.net/account/user";
```
PinpointTownes: Is the userinfo endpoint the same for all regions?

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.CallbackPath"/>.
+        /// </summary>
+        public const string CallbackPath = "/signin-battlenet";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.AuthorizationEndpoint"/> for US servers.
+        /// </summary>
+        public const string AuthorizationEndpointUS = "https://us.battle.net/oauth/authorize";
```
PinpointTownes: Consider creating nested classes to avoid the region suffix: `BattleNetAuthenticationDefaults.US.AuthorizationEndpoint`

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
Gekctek: Good point. sorry i have been moving apartments so ive been busy the last week

PinpointTownes:
Looks good. Don't forget the `Region` property, as suggested by @jerriep: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/42#discussion_r39918990

Gekctek:
Fixed Region property

PinpointTownes:
Rebased and merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/c241e55e71f6c07f1f1b2c0bdf728c14cf9b66c4 :tada: 

Thanks for your contribution! :+1:

PR #43, opened 2015-09-10
Update to Sake 0.2.2 to fix build failures

pranavkm:


PinpointTownes:
Thanks for this PR @pranavkm! :+1: 

Dumb question: as reported by @davidfowl, @kichalla committed a sightly different change on aspnet/Universe: https://github.com/aspnet/Universe/commit/a7bebaae83b0540a6e6b37b9be2a20d912eac39a. Unlike yours, it is supposed to download the latest Sake version from NuGet.org. What is the recommended option?

Also, I'm not sure why the PR is not building. I tried restarting the AppVeyor build to make sure it was not a transient bug, but it's still failing.

PinpointTownes:
Merged, thanks @pranavkm! :+1: 

OAuth2: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/e32a75910797eb2f6f9c3a3119a440cc9f4dd331 ; OpenID2: https://github.com/aspnet-contrib/AspNet.Security.OpenId.Providers/commit/db41cfc771ceb726c63ba4f0115f415a0db730f2

pranavkm:
@PinpointTownes, either works. The one slight advantage of using a fixed version as opposed to the latest is that nuget.exe doesn't have to check the network if `0.2.2` is already installed. In general Sake hasn't had any changes and we don't anticipate seeing too many changes in Sake that break KoreBuild.

PinpointTownes:
Great, thanks for the clarification!

PR #50, opened 2015-09-14
Fitbit provider release

jmloeffler:
Here's a basic implementation for Fitbit.  I had conflicts updating my fork because of the other projects being added.  I apologize for the extra commit but it should merge cleanly.  Let me know if there are any suggestions or questions.


src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationDefaults.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.UserInformationEndpoint"/>.
+        /// </summary>
+        public const string UserInformationEndpoint = "https://api.fitbit.com/1/user/-/profile.json";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthAuthenticationOptions.Scope"/>  Modify the Scope list
+        /// to customize the permissions being requested.  Defaults to: "activity", "nutrition", "profile", "settings", "sleep", "social", "weight".
+        /// </summary>
+        public static List<string> Scope => new List<string> {"activity", "nutrition", "profile", "settings", "sleep", "social", "weight"};
```
PinpointTownes: @jerriep can you please help reviewing whether we're not too intrusive here?

PinpointTownes: @jmloeffler have you considered exposing an `IReadOnlyList<string>` here? (using `.AsReadOnly()`)

jerriep: Whoa, yes looks excessive. @jmloeffler it looks like you can get by with just the "profile" scope.

PinpointTownes:
@jmloeffler thanks for your contribution! :+1: 

Except our minor remarks, it looks fine, good job :smile:

jmloeffler:
I like the `.AsReadOnly()` suggestion.  I will incorporate that and add a commit for that.  As for the scope limiting, my intent was to document as best I could what scopes are available and let app-writers decide for themselves what scopes to limit their app to.  While profile scope is all that's required for the authentication process, it would be pretty limiting out of the box without customizing the scope options.  I thought it best to be wide open by default and only require customization for those who are further along in their app development and know what permissions they'll need.  But that's just a philosophical choice and I'm happy to concede.

Any suggestions on how to inform app-writers what scope options are possible?  The xml comment I put on the FitbitAuthenticationDefaults.cs is not very discoverable.  Maybe I just include the sample list of scopes in the Startup.cs of the sample app.  I could be over-thinking this.  Thoughts?

jmloeffler:
It isn't immediately clear to me what broke the build but I will look into it.

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationDefaults.cs
PinpointTownes: Now that it's just "profile", a list is probably necessary. Consider removing this property and hardcoding "profile" directly in the options. In the future, we may consider adding a `FitbitAuthenticationConstants.Scope` class to expose more scopes if you thing it's worth it.

PinpointTownes:
> Any suggestions on how to inform app-writers what scope options are possible? The xml comment I put on the FitbitAuthenticationDefaults.cs is not very discoverable.

@jerriep thoughts?

> It isn't immediately clear to me what broke the build but I will look into it.

Recent naming changes in aspnet/Security ruined our existing providers and our Yeoman generator. I sent a commit to fix the issues you're seeing, so make sure to rebase your branch :+1:

jerriep:
@jmloeffler I understand your reasoning about including all of them, but I can tell you from experience that people are going to complain about being so aggressive with requesting permissions. Best is to request the absolute minimum by default and let people be specific about picking extra scopes.

As far as documenting the available scopes, this is why I created http://www.oauthforaspnet.com/, so that I can easily link people to the relevant documentation. I will need to rework this for ASP.NET 5 at some stage.

/cc @PinpointTownes

jmloeffler:
I hear you @jerriep.  I'm one of those people who would probably complain about aggressive permissions if I were the one using an app written with this library.  I appreciate the feedback.

If I can be of any help down the road with the documentation work you're referring to, please let me know.

jerriep:
Sure, thanks for the offer. I still need to think about how I am going to structure things so I can give people the correct examples for both ASP.NET 5 and ASP.NET 4.6

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+
+            var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            response.EnsureSuccessStatusCode();
+
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+            
+            identity.AddOptionalClaim(ClaimTypes.NameIdentifier, FitbitAuthenticationHelper.GetIdentifier(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim(ClaimTypes.Name, FitbitAuthenticationHelper.GetLogin(payload), Options.ClaimsIssuer);
```
PinpointTownes: Note: if you think other claims are interesting - like settings or social claims - feel free to add them.

Since you use the `AddOptionalClaim`, it's safe to return null from your helper methods. Just make sure your helper doesn't throw an exception if the JSON payload you expect is not there (for instance, if less info were returned because the settings/social scope was not present in the authorization request).

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace AspNet.Security.OAuth.Fitbit {
+    /// <summary>
+    /// Contains static methods that allow to extract user's information from a <see cref="JObject"/>
+    /// instance retrieved from Fitbit after a successful authentication process.
+    /// </summary>
+    public static class FitbitAuthenticationHelper {
+        /// <summary>
+        /// Gets the identifier corresponding to the authenticated user.
+        /// </summary>
+        public static string GetIdentifier([NotNull] JObject user) => user.SelectToken("user.encodedId").ToString();
```
PinpointTownes: Have you considered using `user.Value<JObject>()?.Value<string>("encodedId")`? I think it would be a bit more efficient.

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// Gets the login corresponding to the authenticated user.
+        /// </summary>
+        public static string GetLogin([NotNull] JObject user) => user.SelectToken("user.displayName").ToString();
```
PinpointTownes: Same remark.

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHandler.cs
jmloeffler: Here's what is available from the user profile: https://dev.fitbit.com/docs/user/#example-response-format

I will probably add the avatar links as claims but if there are other things you think would be good let me know.  Nothing else looked interesting to me.

Thanks for the input!

src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHelper.cs
jmloeffler: Ok, I'll try that out.

PinpointTownes:
Rebased and merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/1836930e7d5812b0e4321014241f6488115d05d8

Thanks for your contribution! (and sorry for the delay... :cry)

PR #63, opened 2015-10-11
Add HealthGraph provider.

### LINKED ISSUE #9: Add HealthGraph provider

Res42:
Closes #9.

### DIFF
--- a/AspNet.Security.OAuth.Providers.sln
+++ b/AspNet.Security.OAuth.Providers.sln
@@ -39,6 +39,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AspNet.Security.OAuth.Fours
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AspNet.Security.OAuth.Asana", "src\AspNet.Security.OAuth.Asana\AspNet.Security.OAuth.Asana.xproj", "{DC58F830-EEE1-45BE-B5F4-AB2222E744A5}"
 EndProject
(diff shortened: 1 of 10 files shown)


PinpointTownes:
Woot, thanks for this new contribution! :clap: 

One question: what's the relationship between HealthGraph and runkeeper (used in the endpoint URLs)?

Res42:
Runkeeper owns HealthGraph.

PinpointTownes:
So `HealthGraph` is like a product name actually?

Is it worth using the `Runkeeper` name to match the conventions used by the other providers, that only use the company name?

Res42:
If I understand it right, `Runkeeper` is the company, and they own:
* the `Runkeeper` application (which uses `HealthGraph`) 
* `HealthGrapth`, which is the underlying API.

In their documentation they call the API we connect to `HealthGraph` so I called the provider that.

As for the general naming convention, I thought the convention was to call them by the product name.
What would we call the `Battle.Net` provider? `BattleNet` or `BlizzardEntertainment` or `ActivisionBlizzard`?
Instagram is owned by Facebook, should it be called `Facebook2` (`Facebook` is used for `Facebook`)?

PinpointTownes:
As you figured out, I meant the product name (which is often the same as the company name), by opposition to the API name :smile: 

(that's why the Facebook provider is named "Facebook" and not "SocialGraph" and the Google provider, "Google" and not "GooglePlus")

But if you think the API name (`HealthGraph`) is more appropriate, I'm fine :+1:

Res42:
I do not use this service (or knew about it before writing this) so I don't really care about what it is called here. I can easily change it if you want.
I only used the `HealthGraph` name because in  issue #9 and in the previous project RockstarLabs/OwinOAuthProviders@188bbe43aeed4158d7b2701d75ae055a0b86bc1d it was called `HealthGraph`.

PinpointTownes:
@jerriep any opinion? :smile:

jerriep:
I'd say keep it HealthGraph, because that is the name of the product.

PinpointTownes:
Merged, thanks for your contribution! https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/060ef4ae403606292b71dba1b12f046c4b55f320

PR #65, opened 2015-10-23
changing the user info endpoint

tatx:
changing the user info endpoint to one that does not require user permission for access to user personal information


PinpointTownes:
Merged, thanks! https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/2ac252fae215c7c4d62ac4ccbd3cfb9da1eafa1c

PR #67, opened 2015-10-29
Added Paypal provider

yannicsmeets:


yannicsmeets:
One question:
Am I using OAuth or openid in this case? I looked hard to find the OAuth endpoints but could only find the openid ones which seem to work..

PinpointTownes:
Thanks for your contribution!

> One question:
Am I using OAuth or openid in this case? I looked hard to find the OAuth endpoints but could only find the openid ones which seem to work..

You're using OpenID Connect, which - despite the name - has nothing to do with OpenID 1 or 2. OpenID Connect is derived from OAuth2 and is basically a superset, which explains why your provider works with the generic OAuth2 middleware :smile: 

In theory, one could use the generic OpenID Connect middleware developed by MSFT with Paypal, but I guess it's still useful to have a specific Paypal package.

@jerriep thoughts?

src/AspNet.Security.OAuth.Paypal/PaypalAuthenticationHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Contains static methods that allow to extract user's information from a <see cref="JObject"/>
+    /// instance retrieved from Paypal after a successful authentication process.
+    /// </summary>
+    public static class PaypalAuthenticationHelper {
+        /// <summary>
+        /// Gets the identifier corresponding to the authenticated user.
+        /// </summary>
+        public static string GetIdentifier([NotNull] JObject user)
+        {
+            return user.Value<string>("user_id").Split('/').Last();
```
PinpointTownes: Consider using a null-friendly approach:

`user.Value<string>("user_id")?.Split('/')?.LastOrDefault()`

yannicsmeets:
@PinpointTownes Ah, thanks for the clarification! I improved the code with your suggestion.

jerriep:
Yeah, I just looked at the source code for the PayPal provider in the OWIN OAuth project and I see it also uses the OpenIDConnect endpoints (https://github.com/RockstarLabs/OwinOAuthProviders/blob/master/Owin.Security.Providers/PayPal/PayPalAuthenticationOptions.cs).

Had a quick scan if the documentation and it seems to be the correct (and only) way of doing it.

yannicsmeets:
Moved this to a new branch and put in in a new PR

PR #68, opened 2015-11-10
Added Beam provider.

Schlurcher:
For https://beam.pro.


Schlurcher:
Oh yeah, and the yo generator is out of date for rc2 it seems. Had to change some usings for the Middleware.

PR #70, opened 2015-11-16
Added Instagram Provider

yannicsmeets:
Includes the option to enforce signed requests according to the documentation of instagram: https://instagram.com/developer/secure-api-requests/


PinpointTownes:
Thanks for this contribution and sorry for the delay! I'll give it a look ASAP :clap:

AspNet.Security.OAuth.Providers.sln.DotSettings
```diff
@@ -21,4 +21,7 @@
     <s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_BEFORE_TYPEOF_PARENTHESES/@EntryValue">False</s:Boolean>
     <s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/TYPE_DECLARATION_BRACES/@EntryValue">END_OF_LINE</s:String>
     <s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateInstanceFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String>
-    <s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String></wpf:ResourceDictionary>
\ No newline at end of file
+    <s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String>
```
PinpointTownes: Unrelated changes, please revert them.

samples/Mvc.Client/project.json
```diff
@@ -50,7 +51,7 @@
 
     "commands": {
         "kestrel": "Microsoft.AspNet.Hosting --server=Microsoft.AspNet.Server.Kestrel --server.urls=http://localhost:54540",
-        "web": "Microsoft.AspNet.Hosting server=Microsoft.AspNet.Server.WebListener server.urls=http://localhost:54540"
+        "web": "Microsoft.AspNet.Server.Kestrel"
```
PinpointTownes: Revert that too, please :smile:

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationDefaults.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+        /// </summary>
+        public const string TokenEndpoint = "https://api.instagram.com/oauth/access_token";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+        /// </summary>
+        public const string UserInformationEndpoint = "/users/self";
```
PinpointTownes: Please use an absolute URL here, to match the other endpoints.

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationDefaults.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Default location of the api
+        /// </summary>
+        public const string ApiLocation = "https://api.instagram.com/v1";
```
PinpointTownes: This const won't be necessary when `UserInformationEndpoint` is updated to use an absolute URL :clap:

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+namespace AspNet.Security.OAuth.Instagram {
+    public class InstagramAuthenticationHandler : OAuthHandler<InstagramAuthenticationOptions> {
+        public InstagramAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {}
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+            var userInformationEndpoint = $"{Options.UserInformationEndpoint}?access_token={tokens.AccessToken}";
+
+            if (Options.SignedRequestsEnforced) {
```
PinpointTownes: I'm not opposed to implement this (non-standard) feature, but TBH, it's almost unnecessary in our case, since access tokens will often (well, always :smile:) be stored in encrypted cookies (and not as-is in a database).

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+                var parameters = new Dictionary<string, string> {
+                    ["access_token"] = tokens.AccessToken
+                };
+
+                var sig = SignRequest(InstagramAuthenticationDefaults.UserInformationEndpoint,
```
PinpointTownes: `signature`

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+        public InstagramAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {}
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+            var userInformationEndpoint = $"{Options.UserInformationEndpoint}?access_token={tokens.AccessToken}";
+
+            if (Options.SignedRequestsEnforced) {
+                var parameters = new Dictionary<string, string> {
+                    ["access_token"] = tokens.AccessToken
```
PinpointTownes: If `access_token` is always the only parameter, is it worth using a dictionary? It may help simplifying your signature generation path.

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+        public InstagramAuthenticationOptions() {
+            AuthenticationScheme = InstagramAuthenticationDefaults.AuthenticationScheme;
+            DisplayName = InstagramAuthenticationDefaults.DisplayName;
+            ClaimsIssuer = InstagramAuthenticationDefaults.Issuer;
+
+            CallbackPath = new PathString(InstagramAuthenticationDefaults.CallbackPath);
+
+            AuthorizationEndpoint = InstagramAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = InstagramAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = InstagramAuthenticationDefaults.ApiLocation +
```
PinpointTownes: Don't forget to remove `InstagramAuthenticationDefaults.ApiLocation`.

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+            return new AuthenticationTicket(context.Principal, context.Properties, context.Options.AuthenticationScheme);
+        }
+
+        private static byte[] CreateToken([NotNull] string endpoint, [NotNull] Dictionary<string, string> parameters) {
+            var token = parameters.OrderBy(par => par.Key)
+                .Aggregate(endpoint, (current, par) => current + $"|{par.Key}={par.Value}");
+            return Encoding.UTF8.GetBytes(token);
+        }
+
+        private static string ConvertToHex(byte[] hash) {
```
PinpointTownes: Consider merging this method (and the previous one) with `SignRequest`.

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
PinpointTownes: Not to mention that this option will only impact the userinfo request, which is made before the access token is persisted some (it's still in memory at this point), so the security gain is really marginal.

PinpointTownes:
Minor my few remarks, it looks fine :+1:

yannicsmeets:
Thanks for the feedback!

I added the SignedRequestEnforced as that is an option that can be set for an application within Instagram. If an application uses the instagram api for more than just oauth, it could be that this option if set. If so, this signature is required for all requests, also the ones required for oauth.

AspNet.Security.OAuth.Providers.sln.DotSettings
```diff
@@ -21,4 +21,6 @@
     <s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_BEFORE_TYPEOF_PARENTHESES/@EntryValue">False</s:Boolean>
     <s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/TYPE_DECLARATION_BRACES/@EntryValue">END_OF_LINE</s:String>
     <s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateInstanceFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String>
-    <s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String></wpf:ResourceDictionary>
\ No newline at end of file
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
```
PinpointTownes: Note for me: remove that before merging :smile:

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using Microsoft.AspNet.Authentication.OAuth;
+using Microsoft.AspNet.Http;
+
+namespace AspNet.Security.OAuth.Instagram {
+    /// <summary>
+    /// Defines a set of options used by <see cref="InstagramAuthenticationHandler"/>.
+    /// </summary>
+    public class InstagramAuthenticationOptions : OAuthOptions {
+        public bool SignedRequestsEnforced { get; set; } = false;
```
PinpointTownes: Note for me: remove the explicit `false` before merging (since it's the default value) and add some documentation.

PinpointTownes:
I'm sold :smile: 

(I'll probably change the name before merging. How about `UseSignedRequests`?)

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationOptions.cs
yannicsmeets: Should I not give it default value?

yannicsmeets:
That sounds good! Shall I squash it in one commit?

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationOptions.cs
PinpointTownes: `false` is the default value for `bool` variables/fields/properties you don't explicitly set, so it's not really necessary.

PinpointTownes:
> That sounds good! Shall I squash it in one commit?

Yep, thanks! (if you want to fix the remarks I added before squashing, don't hesitate) :smile:

yannicsmeets:
Obviously ;) Do these look good?

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace AspNet.Security.OAuth.Instagram {
+    public class InstagramAuthenticationHandler : OAuthHandler<InstagramAuthenticationOptions> {
+        public InstagramAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {}
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+            var userInformationEndpoint = $"{Options.UserInformationEndpoint}?access_token={tokens.AccessToken}";
```
PinpointTownes: Wooops, I missed this one: please use `QueryHelpers` instead of manually concatenating the parameters :smile:

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            if (Options.UseSignedRequests) {
+                var signature = SignRequest("/users/self",
+                    tokens.AccessToken, Options.ClientSecret);
+
+                userInformationEndpoint += $"&sig={signature}";
```
PinpointTownes: Same remark :+1:

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+    public class InstagramAuthenticationHandler : OAuthHandler<InstagramAuthenticationOptions> {
+        public InstagramAuthenticationHandler([NotNull] HttpClient client)
+            : base(client) {}
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
+            var userInformationEndpoint = $"{Options.UserInformationEndpoint}?access_token={tokens.AccessToken}";
+
+            if (Options.UseSignedRequests) {
+                var signature = SignRequest("/users/self",
```
PinpointTownes: I'm not a huge fan of this hardcoded path (specially since it's essentially `userInformationEndpoint` without the scheme, domain and base path) but I can't find a better alternative, so let's keep it as-is :smile:

yannicsmeets: Yeah, thats why I initially put it in the Defaults class. Even though that isn't pretty either...

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            CallbackPath = new PathString(InstagramAuthenticationDefaults.CallbackPath);
+
+            AuthorizationEndpoint = InstagramAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = InstagramAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = InstagramAuthenticationDefaults.UserInformationEndpoint;
+
+            Scope.Add("basic");
+
+            SaveTokensAsClaims = false;
```
PinpointTownes: Note for me: remove before merging (https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/5d5b6ec7fdb338b7c72842f0f83ea0cca051c353)

src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (context.Principal?.Identity == null) {
+                return null;
+            }
+
+            return new AuthenticationTicket(context.Principal, context.Properties, context.Options.AuthenticationScheme);
+        }
+
+        private static string SignRequest([NotNull] string endpoint, [NotNull] string accessToken,
+            string secret) {
+            var token = Encoding.UTF8.GetBytes($"{endpoint}|access_token={accessToken}");
```
PinpointTownes: Hmmmm, the way the signature is formatted is not really foolproof and won't work if the user adds query parameters in the userinfo URL.

I hacked a different version. Could you please give it a try? (I don't have an Instagram account :smile:)

```csharp
protected virtual string ComputeSignature(string address) {
    using (var algorithm = new HMACSHA256(Encoding.UTF8.GetBytes(Options.ClientSecret))) {
        // Extract the parameters from the query string.
        var parameters = from parameter in QueryHelpers.ParseQuery(address)
                         orderby parameter.Key
                         from value in parameter.Value
                         orderby value
                         select $"{parameter.Key}={value}";

        // See https://www.instagram.com/developer/secure-api-requests/
        // for more information about the signature format.
        var bytes = Encoding.UTF8.GetBytes($"/users/self|{string.Join("|", parameters)}");

        // Compute the HMAC256 signature.
        var hash = algorithm.ComputeHash(bytes);

        // Convert the hash to its lowercased hexadecimal representation.
        return BitConverter.ToString(hash).Replace("-", "").ToLower();
    }
}
```

yannicsmeets: I checked this, and there are some issues with it. 
* The QueryHelpers.ParseQuery doesn't work as I expected. It seems to parse the url up until the first parameter as the first parameter. So http://localhost:54540/signin?key=value will get parsed to {http://localhost:54540/signin?key, value}. Seems like a bug to me.
* The values of the parameters are all strings. Doing `from value in parameter.Value orderby value`will therefore result in adding a pair for each character in value. I think those two lines can be omitted. 

Finally a question for my curiosity :smile:, why make this method protected? Are other classes in this namespace going to use this method?

PinpointTownes: > The QueryHelpers.ParseQuery doesn't work as I expected. It seems to parse the url up until the first parameter as the first parameter. So http://localhost:54540/signin?key=value will get parsed to {http://localhost:54540/signin?key, value}. Seems like a bug to me.

Ah yeah, my fault. Sounds like it doesn't accept a full URL, but just the query string. I guess `IndexOf` + `Substring` would do the trick. Do you wanna try?

> The values of the parameters are all strings. Doing from value in parameter.Value orderby valuewill therefore result in adding a pair for each character in value. I think those two lines can be omitted.

Hahaha, the downsides of implicit conversions :smile: 
I guess a `StringValues.ToArray()` call would work, but there's probably no point supporting duplicate values.

> Finally a question for my curiosity :smile:, why make this method protected? Are other classes in this namespace going to use this method?

Nope, but developers are free to subclass the handler class to override the default logic. Probably just a matter of personal preference.

yannicsmeets: Thanks for the info! :smile: I will try to fix the parameter fetching.

PinpointTownes:
Merged, thanks for your contribution! :tada: 

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/5e7740c5237245ef14d7e014787110c3b945c143

PR #95, opened 2016-05-27
Extended the Linkedin provider

ericgreenmix:
Added the same updates as the OwinOAuthProviders version

### DIFF
--- a/src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
@@ -24,7 +24,13 @@ namespace AspNet.Security.OAuth.LinkedIn {
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
             [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
-            var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+            var endpoint = Options.UserInformationEndpoint;
+            if (Options.Fields.Count > 0) {
+                var insertIndex = Options.UserInformationEndpoint.LastIndexOf("~") + 1;
+                endpoint = endpoint.Insert(insertIndex, $":({ string.Join(",", Options.Fields)})");
+            }
+
+            var request = new HttpRequestMessage(HttpMethod.Get, endpoint);
             request.Headers.Add("x-li-format", "json");
             request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
 
@@ -44,8 +50,26 @@ namespace AspNet.Security.OAuth.LinkedIn {
             identity.AddOptionalClaim(ClaimTypes.NameIdentifier, LinkedInAuthenticationHelper.GetIdentifier(payload), Options.ClaimsIssuer)
                     .AddOptionalClaim(ClaimTypes.Name, LinkedInAuthenticationHelper.GetName(payload), Options.ClaimsIssuer)
                     .AddOptionalClaim(ClaimTypes.Email, LinkedInAuthenticationHelper.GetEmail(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim(ClaimTypes.GivenName, LinkedInAuthenticationHelper.GetGivenName(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim(ClaimTypes.Surname, LinkedInAuthenticationHelper.GetFamilyName(payload), Options.ClaimsIssuer)
                     .AddOptionalClaim("urn:linkedin:profile", LinkedInAuthenticationHelper.GetPublicProfileUrl(payload), Options.ClaimsIssuer)
-                    .AddOptionalClaim("urn:linkedin:profilepicture", LinkedInAuthenticationHelper.GetProfilePictureUrl(payload), Options.ClaimsIssuer);
+                    .AddOptionalClaim("urn:linkedin:profilepicture", LinkedInAuthenticationHelper.GetProfilePictureUrl(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:industry", LinkedInAuthenticationHelper.GetIndustry(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:summary", LinkedInAuthenticationHelper.GetSummary(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:headline", LinkedInAuthenticationHelper.GetHeadline(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:positions", LinkedInAuthenticationHelper.GetPositions(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:maidenname", LinkedInAuthenticationHelper.GetMaidenName(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:phoneticfirstname", LinkedInAuthenticationHelper.GetPhoneticFirstName(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:phoneticlastname", LinkedInAuthenticationHelper.GetPhoneticLastName(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:phoneticname", LinkedInAuthenticationHelper.GetPhoneticName(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:location", LinkedInAuthenticationHelper.GetLocation(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:specialties", LinkedInAuthenticationHelper.GetSpecialties(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:numconnections", LinkedInAuthenticationHelper.GetNumConnections(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:numconnectionscapped", LinkedInAuthenticationHelper.GetNumConnectionsCapped(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:currentshare", LinkedInAuthenticationHelper.GetCurrentShare(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:sitestandardprofilerequest", LinkedInAuthenticationHelper.GetSiteStandardProfileRequest(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:apistandardprofilerequest", LinkedInAuthenticationHelper.GetApiStandardProfileRequest(payload), Options.ClaimsIssuer)
+                    .AddOptionalClaim("urn:linkedin:pictureurls", LinkedInAuthenticationHelper.GetPictureUrls(payload), Options.ClaimsIssuer);
 
             var principal = new ClaimsPrincipal(identity);
             var ticket = new AuthenticationTicket(principal, properties, Options.AuthenticationScheme);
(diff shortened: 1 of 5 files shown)


PinpointTownes:
Hey,

Thanks for your PR. It looks fine, but I'm concerned about the number of claims that will be eventually stored in the authentication cookies, specially since things like `summary` or `positions` might be particularly heavy (and exceed the browser cookies limits).

Since the fields are listed in the URL, it might be interesting to introduce a `Fields` option like the one that was introduced recently in the Facebook middleware: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookOptions.cs#L65

/cc @jerriep

jerriep:
Yeah I agree. I think it is highly unlikely for people to care about those particular fields so it is best to make them opt-in

One question though @PinpointTownes; I looked at the FB implementation you linked, but where do the save the values of the fields on the Auth ticket?

ericgreenmix:
I went ahead and refactored to add the `Fields` option. That was a great suggestion.

PinpointTownes:
> I looked at the FB implementation you linked, but where do the save the values of the fields on the Auth ticket?

Fields are not stored as-is in the authentication ticket. Only the corresponding claims returned by the provider are.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -24,7 +24,12 @@ public LinkedInAuthenticationHandler([NotNull] HttpClient client)
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
             [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
-            var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+            var endpoint = string.Empty;
```
PinpointTownes: Hum, if the `fields` collection is empty, the endpoint will always be empty, which will likely make `HttpClient` crash. Consider defaulting to `Options.UserInformationEndpoint`.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -24,7 +24,12 @@ public LinkedInAuthenticationHandler([NotNull] HttpClient client)
...
+            if (Options.Fields.Count > 0) {
+                endpoint = $"{Options.UserInformationEndpoint}/~:({string.Join(",", Options.Fields)})";
```
PinpointTownes: Manual concatenation is still a bit fragile when manipulating user-defined URLs. Please consider using `QueryHelpers` instead (e.g https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/03599ad33a0bd4bfd3b0ef42f4b0670ae6f21680/src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationHandler.cs#L33)

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
```diff
@@ -37,5 +47,25 @@ public class LinkedInAuthenticationHelper {
...
+
+        /// <summary>
+        /// Gets the headline corresponding to the authenticated user.
+        /// </summary>
+        public static string GetHeadline([NotNull] JObject user) => user.Value<string>("headline");
+
+        /// <summary>
+        /// Gets the positions corresponding to the authenticated user.
+        /// </summary>
+        public static string GetPositions([NotNull] JObject user) => user["positions"].ToString();
```
PinpointTownes: Curious: what's the type of `positions`?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -22,6 +23,16 @@ public class LinkedInAuthenticationOptions : OAuthOptions {
...
+            Fields.Add("id");
+            Fields.Add("first-name");
+            Fields.Add("last-name");
+            Fields.Add("formatted-name");
+            Fields.Add("email-address");
+            Fields.Add("public-profile-url");
+            Fields.Add("picture-url");
         }
+
+        public ICollection<string> Fields { get; } = new HashSet<string>();
```
PinpointTownes: Please add some XML docs love :smile:

PinpointTownes:
Sorry for the late reply. Minor a thing or two, it's ready to be merged. Please ping me when you're done with the changes :clap:

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
ericgreenmix: Good point, the user could clear the pre-set fields out. Changing this to throw an exception if there are no fields present.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
ericgreenmix: Since the url doesn't use a query string, the `QueryHelpers` won't help. With the exception being thrown if there are no fields present, as noted in the comment above, do you think this will alleviate the fragility you are talking about?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
ericgreenmix: It is a JSON blob.

ericgreenmix:
@PinpointTownes I just pushed up an update. See my comments above on your questions.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -24,7 +25,15 @@ public LinkedInAuthenticationHandler([NotNull] HttpClient client)
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
             [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens) {
-            var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+            if (Options.Fields.Count == 0) {
```
PinpointTownes: @jerriep do you happen to know if specifying at least one explicit field is mandatory? I'd expect LI to simply return basic info in this case.

@ericgreenmix if adding at least one field is mandatory, I'd recommend moving this check to `LinkedInAuthenticationMiddleware`'s constructor to crash at startup time instead of during callback requests.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
PinpointTownes: > Since the url doesn't use a query string, the QueryHelpers won't help.

People are free to specify custom query string parameters in `Options.UserInformationEndpoint`. The OAuth2 providers developed by the ASP.NET Core team have been fixed to make this scenario possible and I'm a bit concerned about making the LI provider inconsistent: https://github.com/aspnet/Security/issues/365

That said, given the weird syntax used by LI, we'd have to parse the `Options.UserInformationEndpoint` provided by the user to insert the fields "component" at the right place, which doesn't seem rock-solid.

@jerriep thoughts?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
PinpointTownes: In this case, maybe it's better to avoid storing it in claims?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
ericgreenmix: Just dug through their docs more. Looks like it will return just basic profile info if nothing is specified. So we can removed that check. I will go ahead and do that.

https://developer.linkedin.com/docs/signin-with-linkedin

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
ericgreenmix: How else would you recommend getting it back from the LI sign-in process?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
ericgreenmix: I see what you are saying. Yeah, LI strange url syntax throws a wrench into it. I will wait for @jerriep to weigh in before trying to do anything with this.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
PinpointTownes: You can override the `OnCreatingTicket` event to add more claims (e.g http://stackoverflow.com/a/35634697/542757)

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
PinpointTownes: /cc @jerriep

jerriep: Yeah, not sure what other alternatives there are. LinkedIn requires that URL to be in a specific format and from past experience with them, any slight deviation from what they require will cause errors as their API does not seem to be very resilient.

Only alternative is to specify a placeholder for the fields in the `UserInfoEndpoint`, e.g.

```
public const string UserInformationEndpoint = "https://api.linkedin.com/v1/people/~:({0})"
```

This way you will know exactly where to insert the fields, and it allows people to add other query string parameters if they want to. This will have to be documented clearly though...

ericgreenmix: Instead of adding `~:({0})` to the UserInformationEndpoint, I am having it do a check for a `~` and insert the field after that if they are supplied. This eliminates the issue of a user clearing out all the `Fields` and the url ending up being invalid with a `~:()` at the end of it. A url of `https://api.linkedin.com/v1/people/~` is valid with LinkedIn.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHelper.cs
ericgreenmix: Makes sense. I went ahead and removed positions.

ericgreenmix:
@PinpointTownes Pushed changes for the couple issues we have been discussing.

PinpointTownes:
@ericgreenmix looks fine, thanks. Can you please squash your PR into a single commit?

@jerriep if you're happy with the changes, please give us the squirrel :smile:

ericgreenmix:
Squashed.

jerriep:
I still have the concern I raised in the beginning, i.e. where are the values of the fields stored? 

I can see you save them as claims, but that is a pre-defined list. So what if the user adds a field which is not in that list? The user can request any of 20+ fields (https://developer.linkedin.com/docs/fields/basic-profile), but you cater for only +- 10 of those.

I am afraid that this will be confusing as someone way very well think they can request `maiden-name` for example, but then find that it is not stored in a claim.

My suggestion is that all fields that returned should somehow be stored

jerriep:
Also, since adding extra claim will increase the cookie size, I would strongly suggest to cut the default list of fields down only to the ones which are required for the `ClaimTypes.NameIdentifier`, `ClaimTypes.Name` and `ClaimTypes.Email` claims - so in other words:

```
Fields.Add("id");
Fields.Add("formatted-name");
Fields.Add("email-address");
```

These are the 3 which are required in most instances for the ASP.NET Identity integration to function properly.

ericgreenmix:
I can get onboard with all that. I will work on that change.

ericgreenmix:
@PinpointTownes I have scaled out to support the minimal fields by default with support of all LinkedIn basic profile fields if the user optional adds them during configuration. As per Jerrie's suggestion

PinpointTownes:
I'm personally not convinced total fields/claims fidelity was really needed, but why not (as long as it doesn't make the final cookies bigger).

@jerriep pong me when it's okay for you, I'd like to merge it tonight :smile:

jerriep:
@PinpointTownes Looks good to me :)

PinpointTownes:
Merged, thanks! :clap: 

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/d5f4d4004721bff79ce6df097286f57446efa581

PinpointTownes:
@ericgreenmix thanks for your contribution :+1:

PR #113, opened 2016-08-28
Stack exchange provider

### LINKED ISSUE #16: Add StackExchange provider

andrewlock:
Fixes #16 

Some bits of this are a little scrappy and may require some additional tidying up, but it works and is tested 😄 

There were basically 2 pain points - the StackExchange TokenEndpoint ignores the `accept` header, and always sends back a string of the form `access_token=...&expires=1234`, so that needs to be handled in `ExchangeCodeAsync`.

Also, the response from the UserInformationEndpoint is both gzipped and utf-8 encoded, which is a bit of a pain, and requires a reference to System.IO.Compression. The handler at the moment assumes that the response will always be utf-8 and gzipped, but we may want to improve it to handle future changes if the api stops doing that?

Finally, there's not actually that many useful claims you can get back - you get an id and a display name, but you can't get FirstName, LastName or Email. 

Still, it works at least!

### DIFF
--- a/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationOptions.cs
@@ -23,5 +23,15 @@ namespace AspNet.Security.OAuth.StackExchange {
             TokenEndpoint = StackExchangeAuthenticationDefaults.TokenEndpoint;
             UserInformationEndpoint = StackExchangeAuthenticationDefaults.UserInformationEndpoint;
         }
+
+        /// <summary>
+        /// The application request key, obtained when registering your application with StackApps
+        /// </summary>
+        public string RequestKey { get; set; }
+
+        /// <summary>
+        /// The site on which the user is registered
+        /// </summary>
+        public string Site { get; set; } = "StackOverflow";
     }
 }
--- a/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationOptions.cs
@@ -4,6 +4,8 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
+using System.Net;
+using System.Net.Http;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 
@@ -22,6 +24,7 @@ namespace AspNet.Security.OAuth.StackExchange {
             AuthorizationEndpoint = StackExchangeAuthenticationDefaults.AuthorizationEndpoint;
             TokenEndpoint = StackExchangeAuthenticationDefaults.TokenEndpoint;
             UserInformationEndpoint = StackExchangeAuthenticationDefaults.UserInformationEndpoint;
+            BackchannelHttpHandler = new HttpClientHandler {AutomaticDecompression = DecompressionMethods.GZip};
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


andrewlock:
It seems the `alpha` mono build is stalling, though the `latest` build completes successfully. There's no errors or anything, it just seems to stop running. I've seen this before with the `alpha` mono release but I never managed to get to the bottom of it.. any ideas?

PinpointTownes:
Thanks for your contribution!

Looks like StackExchange is going to win both the "most stupid API ever" and the "non-standard OAuth2 implementation" awards this year :trollface: 

> Also, the response from the UserInformationEndpoint is both gzipped and utf-8 encoded, which is a bit of a pain, and requires a reference to System.IO.Compression. The handler at the moment assumes that the response will always be utf-8 and gzipped, but we may want to improve it to handle future changes if the api stops doing that?

Indeed, that seems a bit fragile (:gun: the SO guys for not respecting the HTTP spec...).

That said, isn't `HttpClient` supposed to automatically decompress the HTTP response? I thought both GZIP and DEFLATE were enabled by default (well, at least in CoreFX: https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Net/Http/HttpHandlerDefaults.cs#L15).

> Finally, there's not actually that many useful claims you can get back - you get an id and a display name, but you can't get FirstName, LastName or Email.

Doesn't SE have additional scopes we could use for that?

PinpointTownes:
> I've seen this before with the alpha mono release but I never managed to get to the bottom of it.. any ideas?

Nope, no idea. This issue is impacting all our repos and I have no idea why...

andrewlock:
> Looks like StackExchange is going to win both the "most stupid API ever" and the "non-standard OAuth2 implementation" awards this year

Haha, yep it turned into bit more of a trial than I anticipated! 

> That said, isn't `HttpClient` supposed to automatically decompress the HTTP response

Hmmm, yeah that's a good point - I'll dig more into that one, I think I can probably do something with it based on the comments here: https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpContent.cs#L179

> Doesn't SE have additional scopes we could use for that?

Unfortunately not, apparently they use to expose them but they removed them due to privacy concerns. 

According to the docs http://api.stackexchange.com/docs/authentication, the additional scopes allow access to things like their inbox and write access etc. 

> Nope, no idea. This issue is impacting all our repos and I have no idea why...

Gah, how frustrating:s

src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+            AuthenticationScheme = StackExchangeAuthenticationDefaults.AuthenticationScheme;
+            DisplayName = StackExchangeAuthenticationDefaults.DisplayName;
+            ClaimsIssuer = StackExchangeAuthenticationDefaults.Issuer;
+
+            CallbackPath = new PathString(StackExchangeAuthenticationDefaults.CallbackPath);
+
+            AuthorizationEndpoint = StackExchangeAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = StackExchangeAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = StackExchangeAuthenticationDefaults.UserInformationEndpoint;
+            BackchannelHttpHandler = new HttpClientHandler {AutomaticDecompression = DecompressionMethods.GZip};
```
andrewlock: This line lets me get around having to do the gzip decompression and encoding manually.

andrewlock:
@PinpointTownes with the exception of the build failure, I think this should be good to merge now?

PinpointTownes:
Merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/a890142fb068e0129d628f4a8d31891a238c6e4b

Thanks for your contribution :clap:

PR #127, opened 2016-11-22
Take all tokens from the payload into the ClaimsPrincipal.

tvandijck:
This change allows access to all the tokens from the payload, instead of just name and url.


PinpointTownes:
Thanks for this PR.

I agree such a change would be convenient, but unfortunately, it would result in way too many claims being added to the identity (and to the authentication cookie, by extension). Don't forget most browsers enforce strict cookie max size rules: http://browsercookielimits.squawky.net/

PR #132, opened 2017-01-20
Add UserAgent option and pass it as a header when it is used

CrustyJew:
If you don't pass in a user agent to Reddit, they tend to severely rate limit you and it can make a login request take forever.

### DIFF
--- a/src/AspNet.Security.OAuth.Reddit/RedditAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Reddit/RedditAuthenticationHandler.cs
@@ -32,6 +32,10 @@ namespace AspNet.Security.OAuth.Reddit {
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
             request.Headers.Authorization = new AuthenticationHeaderValue("bearer", tokens.AccessToken);
 
+            if (!string.IsNullOrWhiteSpace(Options.UserAgent)) {
+                request.Headers.Add("User-Agent", Options.UserAgent);
+            }
+
             var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
             if (!response.IsSuccessStatusCode) {
                 Logger.LogError("An error occurred while retrieving the user profile: the remote server " +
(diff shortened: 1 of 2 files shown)


PinpointTownes:
Thanks for bringing this to our attention.

It's worth noting that the OAuth2 generic middleware already registers a default user agent: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthMiddleware.cs#L113

Do you think using an app-specific user agent should be preferred?

CrustyJew:
In this case, I think so due to how Reddit specifies their requirements for a UserAgent.

https://github.com/reddit/reddit/wiki/API

> Change your client's User-Agent string to something unique and descriptive, including the target platform, a unique application identifier, a version string, and your username as contact information, in the following format:
>
>\<platform\>:\<app ID\>:\<version string\> (by /u/\<reddit username\>)
>
>-Example:
>
>User-Agent: android:com.example.myredditapp:v1.2.3 (by /u/kemitche)
>
>-Many default User-Agents (like "Python/urllib" or "Java") are drastically limited to encourage unique and descriptive user-agent strings.
>-Including the version number and updating it as you build your application allows us to safely block old buggy/broken versions of your app.
>-NEVER lie about your user-agent. This includes spoofing popular browsers and spoofing other bots. We will ban liars with extreme prejudice.

src/AspNet.Security.OAuth.Reddit/RedditAuthenticationHandler.cs
```diff
@@ -32,6 +32,10 @@ public RedditAuthenticationHandler([NotNull] HttpClient client)
             request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
             request.Headers.Authorization = new AuthenticationHeaderValue("bearer", tokens.AccessToken);
 
+            if (!string.IsNullOrWhiteSpace(Options.UserAgent)) {
+                request.Headers.Add("User-Agent", Options.UserAgent);
```
PinpointTownes: Okay, I'm sold :smile: 

Can you please triple check the added user agent **replaces** the default user agent used by the OAuth2 generic middleware?

CrustyJew: Because the other header is set with `.DefaultHeaders` it is overridden correctly when it's expressly supplied with an actual value in the request.

Unfortunately I can't get httpclient tracing to behave properly for the request, but manually stepping through the request and response indicate that it is working properly.

PinpointTownes:
Rebased and merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/9445803036ef258dc2f82b7fbc0740344c0b4502

@CrustyJew thanks!

PinpointTownes:
FYI, it will be included in the next release (that should ship this week)

CrustyJew:
Awesome thanks!

Issue #139, opened 2017-02-26
Migrate solution and projects to new format


StummeJ:
Seems like all of the cool kids (aka aspnet) are moving to the new project format `csproj`.  Once you have the new sdk and tools installed its next to impossible to open the project.

PinpointTownes:
Done: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/1d97a85f9e95ea8de011287983af5ef0a07c6098

PR #140, opened 2017-02-26
Migrate to new csproj format 

StummeJ:
dotnet core preview 4 sdk switched to a new `csproj` format.  It is next to impossible if not impossible to get the old format to build when you install the preview 4 tools.

It also seems like all of the main line @aspnet projects are on the new format which makes it hard to contribute to both those projects and these.

In reference to issue #139


PinpointTownes:
Thanks for taking the time to do that. It looks like the .sln file was not correctly updated, which seems to be the root cause of the build errors.

(please note that I won't merge it before VS 2017 and the new MSBuild-based tooling are officially released)

samples/Mvc.Client/Mvc.Client.csproj
```diff
@@ -0,0 +1,123 @@
...
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Yahoo/AspNet.Security.OAuth.Yahoo.csproj" />
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Yammer/AspNet.Security.OAuth.Yammer.csproj" />
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Extensions/AspNet.Security.OAuth.Extensions.csproj">
+      <FromP2P>true</FromP2P>
+    </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies">
+      <Version>1.0.0</Version>
```
PinpointTownes: Hum, I thought there was a shorter syntax for that?

```xml
<PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="1.0.0" />
```

samples/Mvc.Client/Mvc.Client.csproj
```diff
@@ -0,0 +1,123 @@
+<Project Sdk="Microsoft.NET.Sdk.Web" ToolsVersion="15.0">
+
+  <PropertyGroup>
+    <TargetFrameworks>net451;netcoreapp1.0</TargetFrameworks>
+    <RuntimeIdentifiers>win7-x64;win7-x86;osx.10.10-x64;osx.10.11-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64</RuntimeIdentifiers>
```
PinpointTownes: Hum, the samples are portable apps (at least, currently). Why do we need the RIDs here?

StummeJ:
Sounds good. I'll fix the sln.

src/AspNet.Security.OAuth.Beam/AspNet.Security.OAuth.Beam.csproj
```diff
@@ -27,28 +18,13 @@
...
-      <Version>1.6.0</Version>
-    </PackageReference>
-    <PackageReference Include="JetBrains.Annotations">
-      <Version>10.1.4</Version>
-      <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth">
-      <Version>1.0.0</Version>
-    </PackageReference>
+    <PackageReference Include="JetBrains.Annotations" Version="10.1.4" />
```
PinpointTownes: Please note that we want this dependency to be build-time only.

StummeJ:
I've migrated everything over to csproj now and fixed the sln.  The migration tool in preview4 had a bug there. We're still getting build errors that look to me more bugs in the toolchain.  I'll look further into this later...

For now: **anyone that is thinking about upgrading to preview4, DONT**

PinpointTownes:
> I've migrated everything over to csproj now and fixed the sln.

Thanks! :clap: 

I took a look at the CI logs, and most/all the restore errors seem to be caused by the fact the NuGet `imports` that are needed to make the JetBrains annotations package work with .NET Standard are no longer here.

The good news is that a newer package that supports .NET Standard 1.0 was released in October: https://www.nuget.org/packages/JetBrains.Annotations/10.2.1, so I bet migrating to the latest bits should help get rid of most the errors we're seeing.

StummeJ:
Which build log are these?  It looks like even after updating it is still failing.

PinpointTownes:
Yeah, but it fails at build time, now: https://ci.appveyor.com/project/aspnet-contrib/aspnet-security-oauth-providers/build/1.0.0-beta3-374

StummeJ:
:joy:  Windows is building.  Linux is having issues still... So it  OSX... but this is a good starting point.

StummeJ:
Okay so we'll see how this goes.  There was a new project type UUID for preview4.  Once I updated to that and fixed a few things in the props files it started to build most things.  Only thing not building on my machine is the samples.

build/dependencies.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <PropertyGroup>
+    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
```
PinpointTownes: 1.6.1 shipped with .NET Core 1.1.0. Since we target LTS, we probably want 1.6.0 if we decide to explicitly reference `NETStandard.Library`.

build/dependencies.props
```diff
@@ -0,0 +1,6 @@
...
+    <CoreFxVersion>4.3.0</CoreFxVersion>
```
PinpointTownes: Same remark, we need to target the version that shipped with LTS.

samples/Mvc.Client/Mvc.Client.csproj
```diff
@@ -0,0 +1,95 @@
+<Project Sdk="Microsoft.NET.Sdk.Web" ToolsVersion="15.0">
+
+  <PropertyGroup>
+    <RuntimeIdentifiers>win7-x64;win7-x86;osx.10.10-x64;osx.10.11-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;centos.7-x64;rhel.7.2-x64;debian.8-x64;fedora.23-x64;opensuse.13.2-x64</RuntimeIdentifiers>
```
PinpointTownes: Hum, looks like the RIDs reappeared. Are they automagically added by VS?

samples/Mvc.Client/Mvc.Client.csproj
```diff
@@ -0,0 +1,95 @@
...
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.WordPress/AspNet.Security.OAuth.WordPress.csproj" />
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Yahoo/AspNet.Security.OAuth.Yahoo.csproj" />
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Yammer/AspNet.Security.OAuth.Yammer.csproj" />
+    <ProjectReference Include="../../src/AspNet.Security.OAuth.Extensions/AspNet.Security.OAuth.Extensions.csproj">
+      <FromP2P>true</FromP2P>
+    </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore" Version="1.0.3" />
```
PinpointTownes: Make sure we use `1.0.0` everywhere.

samples/Mvc.Client/Mvc.Client.csproj
StummeJ: Yeah VS adds an error when you try to open the sample project without RID.

build/dependencies.props
StummeJ: Looks like this will go away as preview4 assumes a version automatically.

src/AspNet.Security.OAuth.GitHub/project.json
```diff
@@ -1,54 +0,0 @@
...
-    "JetBrains.Annotations": { "type": "build", "version": "10.1.4" },
-    "Microsoft.AspNetCore.Authentication.OAuth": "1.0.0"
-  },
-
-  "frameworks": {
-    "net451": { },
-
-    "netstandard1.3": {
-      "dependencies": {
-        "System.Dynamic.Runtime": "4.0.11"
```
PinpointTownes: This dependency was necessary to make JSON.NET work with `dynamic`. It looks like it's no longer here in the .csproj. We'll have to make sure it doesn't cause a regression.

samples/Mvc.Client/Mvc.Client.csproj
GuardRex: @StummeJ Try with ...
```
<RuntimeFrameworkVersion>1.1.0</RuntimeFrameworkVersion>
```
I just found out that sample apps I'm working with will restore and compile for both `netcoreapp1.1` and `net451` if the props are set up this way ...
```
<PropertyGroup>
  <TargetFrameworks>net451;netcoreapp1.1</TargetFrameworks>
  <OutputType>Exe</OutputType>
  <RuntimeFrameworkVersion>1.1.0</RuntimeFrameworkVersion>
</PropertyGroup>
```
```
dotnet restore
dotnet publish -r Release -f netcoreapp1.1
dotnet publish -r Release -f net451
```

PinpointTownes:
@StummeJ wooops, looks like your PR was impacted by the version changes in project.json (a new wave was released last week), I'm sorry for that.

FYI, I spent a few hours migrating most the aspnet-contrib repos to .csproj and it looks like everything is now working fine on both AppVeyor and Travis (see https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/pull/63/files for an example).

That would be great if you could rebase your PR on top of the `dev` branch. If you can't, please ping me, I'll cherry-pick your commits and continue locally.

Thanks for your contribution! :clap:

build.ps1
```diff
@@ -3,7 +3,7 @@ cd $PSScriptRoot
 $repoFolder = $PSScriptRoot
 $env:REPO_FOLDER = $repoFolder
 
-$koreBuildZip="https://github.com/aspnet/KoreBuild/archive/1.0.0.zip"
+$koreBuildZip="https://github.com/aspnet/KoreBuild/archive/feature/msbuild.zip"
```
PinpointTownes: FYI, this version is obsolete and will be removed in the future. You need the `dev` version: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/pull/63/files#diff-f0e02e42e53ec40f5d3b8bed031ad09e

build.sh
```diff
@@ -2,7 +2,7 @@
 repoFolder="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $repoFolder
 
-koreBuildZip="https://github.com/aspnet/KoreBuild/archive/1.0.0.zip"
+koreBuildZip="https://github.com/aspnet/KoreBuild/archive/feature/msbuild.zip"
```
PinpointTownes: This file must also be updated: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/pull/63/files#diff-0b83f9dedf40d7356e5ca147a077acb4

build/common-oauth.props
```diff
@@ -0,0 +1,22 @@
+<Project>
+    <Import Project=".\common.props" />
+
+    <PropertyGroup>
+        <OutputType>Library</OutputType>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <EmbeddedResource Include="compiler\resources\**\*" />
```
PinpointTownes: This doesn't seem to be necessary.

build/common-oauth.props
```diff
@@ -0,0 +1,22 @@
...
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="../AspNet.Security.OAuth.Extensions/AspNet.Security.OAuth.Extensions.csproj" />
+        <PackageReference Include="JetBrains.Annotations" Version="10.2.1" PrivateAssets="All" />
+        <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth" Version="1.0.0" />
+    </ItemGroup>
+
+    <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
```
PinpointTownes: Please remove.

build/common-oauth.props
```diff
@@ -0,0 +1,22 @@
...
+        <OutputType>Library</OutputType>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <EmbeddedResource Include="compiler\resources\**\*" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="../AspNet.Security.OAuth.Extensions/AspNet.Security.OAuth.Extensions.csproj" />
+        <PackageReference Include="JetBrains.Annotations" Version="10.2.1" PrivateAssets="All" />
```
PinpointTownes: `10.3.0`.

That would be cool if you could use the same pattern as the one I used here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/pull/63/files#diff-30dcc9789f8ea3c7632e4f054e60c072

build/common.props
```diff
@@ -0,0 +1,20 @@
+<Project>
+  <Import Project="dependencies.props" />
+  <Import Project="..\version.props" />
+
+  <PropertyGroup>
+    <Product>Microsoft .NET Extensions</Product>
```
PinpointTownes: `aspnet-contrib`

version.props
```diff
@@ -0,0 +1,7 @@
+<!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
+<Project>
+    <PropertyGroup>
+        <VersionPrefix>1.0.0</VersionPrefix>
+        <VersionSuffix>beta3</VersionSuffix>
```
PinpointTownes: `rc1`

Please also move this file to the `build` folder, so we keep everything in the same place.

build/common.props
```diff
@@ -0,0 +1,20 @@
...
+    <WarningsAsErrors>true</WarningsAsErrors>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <PackageProjectUrl>https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers</PackageProjectUrl>
+    <PackageLicenseUrl>http://www.apache.org/licenses/LICENSE-2.0.html</PackageLicenseUrl>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>git://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers</RepositoryUrl>
+    <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard1.3' AND '$(OutputType)'=='library'">
```
PinpointTownes: You can kill that, it's what causing all these warnings in the logs (because well, the `NETStandard.Library` is automatically referenced for us, so no need to have that here)

PinpointTownes:
Replaced by https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/148.

@StummeJ thanks for your contribution.

PR #150, opened 2017-03-23
Trello Provider for AspNet.Security.OAuth

gijswijs:
Just add the following to startup.cs and you should be golden:

```
app.UseTrelloAuthentication(new TrelloAuthenticationOptions
            {
                ConsumerKey = "[consumerkey]",
                ConsumerSecret = "[consumersecret]",
                AppName = "Trello oAuth Example"
            });
```

You can find your key and secret here: https://trello.com/app-key

Trello uses oAuth 1.0, so I based my provider on the Microsoft.AspNetCore.Authentication.Twitter provider.

### DIFF
--- a/AspNet.Security.OAuth.Providers.sln
+++ b/AspNet.Security.OAuth.Providers.sln
@@ -87,6 +87,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AspNet.Security.OAuth.MailC
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AspNet.Security.OAuth.Autodesk", "src\AspNet.Security.OAuth.Autodesk\AspNet.Security.OAuth.Autodesk.xproj", "{92406BD8-0270-4BF9-B860-D1AE6508DF8C}"
 EndProject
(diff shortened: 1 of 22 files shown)


PinpointTownes:
Hum, I had not realized this provider would use OAuth1.

Unfortunately, this repository usually only accepts OAuth2 providers as only them can use the new generic middleware that is part of ASP.NET Core (as you figured out, there's no equivalent for the legacy OAuth1 protocol).

Sorry for the inconvenience.

gijswijs:
@PinpointTownes But Asp.Net Core supports Twitter as well (which is also oAuth1)

It's true that it doesn't use the generic middleware, but the Twitter oAuth implementation of Microsoft itself doesn't do that as well, and they do ship that with ASP.Net Core.

So why would this repository be more strict about it than Microsoft itself is?

(I'm ok with it either way, I'm just curious)

PinpointTownes:
@gijswijs the reason is quite simple, actually:

Unlike OAuth2, OAuth1 received absolutely no attention in ASP.NET Core (the code was just ported as-is) and no effort was made to simplify or rationalize it to support a "generic" model. Unfortunately, this means that we'd have to copy the same code again and again for each OAuth1 provider we'd like to support.

Since I'm the guy maintaining this repository, keeping this code up-to-date - and porting it to ASP.NET Core 2.0, whose authentication stack will be completely different - would be my responsibility. Yet, I already dedicate **much time** to OSS projects, and I'm not really willing to make my pain worse with duplicated legacy code :smile: 

Hope you'll understand.

gijswijs:
@PinpointTownes I understand it totally!

People can check out my fork if they want to use Trello, or as an example of an oAuth1.0 implementation.

PR #155, opened 2017-04-12
Removing automatic scope addition

arkasha:
Automatically adding vso.profile as a scope will cause any VSTS applications to return error=InvalidScope if the vso.profile scope was implicit instead of explicit.

### DIFF
--- a/src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationOptions.cs
@@ -25,8 +25,6 @@ namespace AspNet.Security.OAuth.VisualStudio
             AuthorizationEndpoint = VisualStudioAuthenticationDefaults.AuthorizationEndpoint;
             TokenEndpoint = VisualStudioAuthenticationDefaults.TokenEndpoint;
             UserInformationEndpoint = VisualStudioAuthenticationDefaults.UserInformationEndpoint;
-
-            Scope.Add("vso.profile");
         }
     }
 }


PinpointTownes:
Interesting. Do you have any link where this issue is discussed? Also, could you please clarify what you mean by "implicit scope"? If we don't add this scope, will the VS middleware be able to retrieve at least the user identifier and the name associated with the profile?

arkasha:
> If we don't add this scope, will the VS middleware be able to retrieve at least the user identifier and the name associated with the profile?

Yes, you get the name and email as claims.

As far as I know there isn't any place where this is discussed. I discovered this issue while trying to use the vso provider and having it fail no matter how I configured it. What I mean by implicit scope is when you select the scopes for your application when provisioning some scopes implicitly include other scopes. You can see the details here: https://www.visualstudio.com/en-us/docs/integrate/get-started/auth/oauth

Specifically take a look at `vso.profile`, the "Included by" list is fairly long. Here's an example, create an application that requests `vso.release_manage`. `vso.release_manage` implies `vso.release` which in turn implies `vso.profile`. When requesting an access code via OAuth, if you specify anything other than  `vso.release_manage` the call will fail with invalid_scope. So basically having `vso.profile` added in the constructor will break any application which is not explicitly configured with `vso.profile`.

PinpointTownes:
@arkasha thanks for the explanation.

I'm fine accepting your fix as long as the 4 claims we automatically add in the handler are there even if we remove the profile scope: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs#L55-L58

Can you please try that to ensure they are there when removing the profile scope?

PinpointTownes:
Note: the 1.0.0-rc1 packages are about to ship. If you want your fix to be included in this release, please confirm it's working as intended.

arkasha:
It is working as intended.

arkasha:
And yes, those claims are still present.

PinpointTownes:
Merged, thanks! :+1:

PR #158, opened 2017-05-21
Update the .csproj files to ensure AspNet.Security.OAuth.Extensions is not referenced by the NuGet packages

kevinchalet:
All the OAuth2 rc1-final packages can't be installed due to the unwanted `AspNet.Security.OAuth.Extensions` dependency added to the .nuspec definitions. Since this package is a sources project, we don't publish it and NuGet displays an errors stating that `AspNet.Security.OAuth.Extensions` `rc1-final` cannot be found.


kevinchalet:
Note: new packages will be published tomorrow to fix this critical issue.

PR #163, opened 2017-06-18
Add an Odnoklassniki provider

Vertigo093i:
* https://ok.ru/
* [Server OAuth authorization](https://apiok.ru/en/ext/oauth/server)


src/AspNet.Security.OAuth.Odnoklassniki/OdnoklassnikiAuthenticationExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using AspNet.Security.OAuth.Odnoklassniki;
+using JetBrains.Annotations;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class OdnoklassnikiAuthenticationExtensions
+    {
+        public static IApplicationBuilder UseOdnoklassnikiAuthentication(
```
kevinchalet: Can you please add the XML docs? You can copy/paste them from another provider.

src/AspNet.Security.OAuth.Odnoklassniki/OdnoklassnikiAuthenticationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            var address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, "access_token", tokens.AccessToken);
+
+            // Signing.
+            // Call API methods using access_token instead of session_key parameter
+            // Calculate every request signature parameter sig using a little bit different way described in
+            // https://apiok.ru/dev/methods/
+            // session_secret_key = MD5(access_token + application_secret_key)
+            // sig = MD5(request_params_composed_string + session_secret_key)
+            // Don't include access_token into request_params_composed_string
+            var args = new Dictionary<string, string>
```
kevinchalet: Please consider moving that to a separate private/protected method like what we do in the other providers that require this stupid signing thing.

src/AspNet.Security.OAuth.Odnoklassniki/OdnoklassnikiAuthenticationOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            AuthorizationEndpoint = OdnoklassnikiAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = OdnoklassnikiAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = OdnoklassnikiAuthenticationDefaults.UserInformationEndpoint;
+
+            Scope.Add("VALUABLE_ACCESS");
+            Scope.Add("GET_EMAIL");
+        }
+
+        public string ClientPublic { get; set; }
```
kevinchalet: XML docs? Also, not sure the name is descriptive enough. What does this property do?

kevinchalet (changes requested):

kevinchalet:
@Vertigo093i thanks for your PR (and sorry for the late review :sweat:)

I added a few remarks, but overall it looks good :+1:

kevinchalet:
@Vertigo093i I don't see the ugly signing part mentioned in the https://apiok.ru/en/ext/oauth/server document. Is it really a mandatory thing or just something optional?

Vertigo093i:
Yes, the signing is mandatory for API calls. Have a look at https://apiok.ru/en/dev/methods/.

src/AspNet.Security.OAuth.Odnoklassniki/OdnoklassnikiAuthenticationHandler.cs
Vertigo093i: OK. Found such thing in the Instagram provider.

src/AspNet.Security.OAuth.Odnoklassniki/OdnoklassnikiAuthenticationOptions.cs
Vertigo093i: Well, this one is another (besides `ClientId` and `ClientSecret`) parameter of an application. It is referred to as "application public key". I've named the property this way to be consistent with the other properties. I'm leaving it unchanged for now (only adding docs). Let me know if you require it to be named differently. May be `PublicKey` or `AppPublic`. Or even `AppPublicKey`.

kevinchalet: `ClientPublic` is confusing as it makes me think that it's a flag used to determine whether the client app is a confidential or public app (which is absolutely not what it does).

Consider renaming it to `ApplicationKey` (it's the name I've used for the OpenID2 Steam provider: https://github.com/aspnet-contrib/AspNet.Security.OpenId.Providers/blob/dev/src/AspNet.Security.OpenId.Steam/SteamAuthenticationOptions.cs#L25)

Vertigo093i: /// <summary>
        /// Gets or sets the application key used to retrieve user details from Odnoklassniki's API.
        /// </summary>
        public string ApplicationKey { get; set; }

Good enough?

Vertigo093i:
@PinpointTownes Would you like to merge this PR?

kevinchalet:
> @PinpointTownes Would you like to merge this PR?

Yep. Could you please rebase and squash your PR into a single commit? Thanks.

Vertigo093i:
No, I can't. Since you requested some changes and they were made as part of
this PR. That's how they work.

kevinchalet:
Not sure I understand what prevents you from rebasing your PR on top of the `dev` branch and squashing all your commits into a single one. All the contributors before you have been requested to do something similar. That said, if you prefer, we can simply close the PR.

PR #166, opened 2017-07-30
Fix an issue with the LinkedIn provider not parsing the standard profile request fields not being parsed correctly

AlvinStanescu:
Fix the `InvalidCastException: Cannot cast Newtonsoft.Json.Linq.JObject to Newtonsoft.Json.Linq.JToken.` thrown when adding `api-standard-profile-request` and `site-standard-profile-request` to the option fields to retrieve.

I used a similar pattern to the one from the above fields.


kevinchalet:
Thanks for your contribution. Per discussion, we concluded it was better to simply remove these helpers, which was done as part of https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/91e709c40904e2b09976fa64fec4a958085fa41c.

Issue #170, opened 2017-08-14
Port the social providers to the new ASP.NET Core 2.0 authentication stack


kevinchalet:

martincostello:
Will a final `1.0.0` be published to NuGet before this?

kevinchalet:
@martincostello not sure. The OpenID generic provider - that shares the same versioning as this repository - relies on AngleSharp, that doesn't yet offer a 1.0.0 version :cry:

kinosang:
I wanna to know where should I start the jobs? Is it okay to start working on the `dev` branch? or should we start a new branch for it?

runxc1:
Also wondering about this as I need a Strava login for asp.net core 2.0

ntemirov:
I need vkontakte authentication for asp.net core 2.0. Can I help you?

kinosang:
I'm starting working on it, would you like to create a new branch such as `2.0` because I can't port all those providers at once.

BTW, I'll pay attention to the problems you pointed at aspnet-contrib/AspNet.Security.OpenId.Providers#35.

sergeushenecz:
Hello, everyone.Can I know when you to porting to asp.net core 2.0 VK provider?

0xNF:
Hey all, I ported the Spotify provider service to the ASP.NET Core 2.0 stack. It's modeled after [Microsoft's own implementations](https://github.com/aspnet/Security/tree/dev/src/Microsoft.AspNetCore.Authentication.MicrosoftAccount) of their Facebook and MicrosoftAccount authorizations.

The problem is that it's a homeless amalgam, and I don't know where to put it. It uses Microsoft's namespaces and dependencies, and thus requires nothing from the other parts of this project, although that could obviously be changed in a future patch. 

Is there a 2.0 branch in the works for this project, or otherwise some place that I can push this? Would be a shame to keep it to myself.

kinosang:
@0xNF that's why I ask for a new branch.

vishalmgiri:
Please update us on Asp.net Core 2 Support.
I am also willing to support it.

kevinchalet:
[...]

PR #179, opened 2017-09-21
Migrate the Amazon provider to ASP.NET Core 2.0

martincostello:
Migrate the Amazon provider to ASP.NET Core 2.0.

Relates to #170.


src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationHandler.cs
```diff
@@ -4,31 +4,37 @@
...
+            [NotNull] ISystemClock clock)
+            : base(options, logger, encoder, clock)
         {
         }
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity,
             [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens)
         {
-            var endpoint = Options.UserInformationEndpoint;
+            string endpoint = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, "access_token", tokens.AccessToken);
```
kevinchalet: Why this change? We don't want functional changes as part of the 2.0 port.

src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationOptions.cs
```diff
@@ -20,28 +21,30 @@ public class AmazonAuthenticationOptions : OAuthOptions
...
         /// <summary>
         /// Gets the list of fields to retrieve from the user information endpoint.
         /// </summary>
-        public ISet<string> Fields { get; } = new HashSet<string>
-        {
-            "email",
-            "name",
-            "user_id"
-        };
+        public ICollection<string> Fields { get; } = new HashSet<string>();
```
kevinchalet: Revert.

kevinchalet (changes requested):

src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationHandler.cs
martincostello: This comes from how I originally implemented Amazon authentication in my own project before I did the initial PR to add it here: [link](https://github.com/martincostello/alexa-london-travel-site/blob/28b5adea435fdaed519add79e4f4dff1fab30889/src/LondonTravel.Site/Identity/Amazon/AmazonHandler.cs#L33)

src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationOptions.cs
martincostello: Why? It's functionally equivalent because the fields are populated by the constructor now instead of inline and `ISet<T>` wasn't found when I tried to initially compile it for 2.0.

src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationHandler.cs
kevinchalet: It's always preferable to flow the access token via the `Authorization` header, if possible (which is the approach recommended by the OAuth2 bearer spec).

If Amazon deprecated or removed that, please open a new ticket so we can change the 1.x provider too.

src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationOptions.cs
kevinchalet: Because it's not a necessary change and it makes the 2.0 signature inconsistent with what we have in the 1.x version. Only changes that are strictly necessary to make the providers work with 2.0 are accepted as part of the "2.0 port" PRs.

martincostello:
@PinpointTownes Commit updated to revert those changes.

kevinchalet:
Thanks. Looks like you forgot to remove the helper class, but I'll go ahead and remove it before merging your PR.

kevinchalet (approved):

kevinchalet:
Merged, thanks! https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/03cf0f4f69617429bbc0b39b714e235b5fc71bee

PR #124, opened 2016-11-08
update Vkontakte API

MrMeison:
* update API to actual version (https://vk.com/dev/widget_auth)
* correct method names

PR #181, opened 2017-09-21
Migrate the Odnoklassniki and Vkontakte providers to ASP.NET Core 2.0

Vertigo093i:
For Vkontakte I've reverted the #124 since it is about a client-side (browser) auth rather than a server-side one. And I've double-checked that the API is still actual.


kevinchalet:
Thanks for your PR. I took a brief look and I see many inconsistencies with the other providers (e.g we don't want to have the legacy `IApplicationBuilder` extensions).

Please fix the inconsistencies, including the style differences.

kevinchalet:
> For Vkontakte I've reverted the #124 since it is about a client-side (browser) auth rather than a server-side one.

No behavior change as part of the port PR, please. If you want to revert that, open a new ticket and we'll change it in the 1.x version too.

PR #183, opened 2017-10-02
Ported the LinkedIn authentication provider to ASP .NET Core 2.0

AlvinStanescu:
Ported the LinkedIn authentication provider to ASP .NET Core 2.0.

I followed the GitHub port from @PinpointTownes and tested it in a local app - all seems 👍.


src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationExtensions.cs
```diff
@@ -1,70 +1,77 @@
 ﻿/*
  * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
- * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * See https://LinkedIn.com/aspnet-contrib/AspNet.Security.OAuth.Providers
```
martincostello: This looks like it was changed by an over-aggressive Find-And-Replace.

AlvinStanescu: Good catch, I'll push a fix for this :)

kevinchalet:
Merged, thanks! https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/e1271ea4a029e851a19fc7ba41f43ce47acef27f

PR #185, opened 2017-10-04
Update Slack provider

### LINKED ISSUE #184: Potential Issue with Slack provider
A couple of things related to the Slack provider which I think needs to be addressed:

1. When getting the user profile, the `https://slack.com/api/auth.test` method is called, and not the `https://slack.com/api/users.identity` method which is suggested by the **Sign in with Slack** document (see https://api.slack.com/docs/sign-in-with-slack#request_info)

2. The second, and more serious issue is that the `NameIdentifier` claim is set to the User's ID. See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/c8c923a5b6149fa7693bd3a54bbcbc988fc1a56e/src/AspNet.Security.OAuth.Slack/SlackAuthenticationHandler.cs#L51 and https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/c8c923a5b6149fa7693bd3a54bbcbc988fc1a56e/src/AspNet.Security.OAuth.Slack/SlackAuthenticationHelper.cs#L22-L30

    This can potentially cause issues, as the Slack documentation states that the User ID is not guaranteed to be globally unique across all teams. See the note under the **Response** heading in this document: https://api.slack.com/methods/users.identity

@mjknowles You developed this initially. Any thoughts on these?

@PinpointTownes Not sure how we should handle this as it is a breaking change. But also, not fixing it can cause serious issues down the line as well.

jerriep:
Updates the Slack provider as follows:
* Retrieves user profile information from the `users.identity` endpoint
* Updates `NameIdentifier` claim to be a combination of Team ID and User ID
* Sets default scope of `identity.basic`
* Removes a bunch of claims for information which was not present in `users.identity` endpoint

Closes  #184

### DIFF
--- a/src/AspNet.Security.OAuth.Slack/SlackAuthenticationHelper.cs
+++ b/src/AspNet.Security.OAuth.Slack/SlackAuthenticationHelper.cs
@@ -17,6 +17,40 @@ namespace AspNet.Security.OAuth.Slack
     public static class SlackAuthenticationHelper
     {
         /// <summary>
+        /// Returns a unique identifer for the authenticated user.
+        /// </summary>
+        /// <remarks>
+        /// According to the Slack API documentation at https://api.slack.com/methods/users.identity, user IDs are
+        /// not guaranteed to be globally unique across all Slack users.  The combination of user ID and team ID, 
+        /// on the other hand, is guaranteed to be globally unique.
+        /// </remarks>
+        public static string GetUniqueIdentifier([NotNull] JObject user)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            string userId = user.Value<JObject>("user")?.Value<string>("id");
+            string teamId = user.Value<JObject>("team")?.Value<string>("id");
+
+            return $"{teamId}|{userId}";
+        }
+
+        /// <summary>
+        /// Gets the email address corresponding to the authenticated user.
+        /// </summary>
+        public static string GetUserEmail([NotNull] JObject user)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            return user.Value<JObject>("user")?.Value<string>("email");
+        }
+
+        /// <summary>
         /// Gets the identifier corresponding to the authenticated user.
         /// </summary>
         public static string GetUserIdentifier([NotNull] JObject user)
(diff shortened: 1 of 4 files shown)


src/AspNet.Security.OAuth.Slack/SlackAuthenticationHelper.cs
```diff
@@ -17,133 +17,89 @@ namespace AspNet.Security.OAuth.Slack
...
-        {
-            if (user == null)
-            {
-                throw new ArgumentNullException(nameof(user));
-            }
+            string userId = user.Value<JObject>("user")?.Value<string>("id");
+            string teamId = user.Value<JObject>("team")?.Value<string>("id");
 
-            return user.Value<string>("team_id");
+            return $"{teamId}|{userId}";
```
kevinchalet: Could we simply do `return $"{GetTeamIdentifier(user)}|{GetUserIdentifier(user)}";` and avoid the two intermediate variables?

kevinchalet:
> Removes a bunch of claims for information which was not present in users.identity endpoint

@mjknowles are you fine with that?

jerriep:
> are you fine with that?

Just a few notes regarding this:
1. Testing the provider against the old endpoint (`auth.test`) did not even work at all. I got an unauthorized error. 
2. According to the docs (https://api.slack.com/methods/auth.test) there is no mention of the attributes which were used in those claims.
3. The new `users.identity` also does not return any attributes which can be used to populate those claims.

So I do not think there is much choice but to delete those :(

kevinchalet:
Okay, I'm sold. I'll merge it tonight and leave the ticket open to ensure we adopt the same fix in the 2.0 port.

mjknowles:
@PinpointTownes everything looks good to me. Thanks for updating this code :+1:.

kevinchalet:
@mjknowles do you happen to know if Slack changed something in their `auth.test` endpoint implementation? Looks like the provider didn't work properly for @jerriep.

mjknowles:
@PinpointTownes I unfortunately do not - I haven't been keeping up with Slack's authentication API since my implementation a couple of years ago. If I remember correctly, the users.identity endpoint didn't even exist back then (which is why the original implementation was using the auth.test endpoint to get the username). Not sure what changes have been made to auth.test since.

kevinchalet:
Merged: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/504cc8a008946b9bce44a9960b68aad7a25d0b20.

I ported it to the 2.0 version: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/ec82da32baa9a10de49eb5e2060d7167a64174d4.

Thanks for your PR @jerriep.

PR #204, opened 2017-12-13
Add Baidu provider

kinosang:
Added a new provider of Baidu


kinosang:
As @khellang suggested in #216 and #217, class `BaiduClaimTypes` and newline are added.

PR #216, opened 2018-02-15
Add ClaimTypes constants

### LINKED ISSUE #97: Add constants classes to expose custom claims/parameters

khellang:
I think this should cover them all 😄 

Closes #97


src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationOptions.cs
```diff
@@ -27,10 +27,10 @@ public YahooAuthenticationOptions()
 
             ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "guid");
             ClaimActions.MapJsonKey(ClaimTypes.Name, "nickname");
-            ClaimActions.MapJsonKey("urn:yahoo:familyname", "familyName");
-            ClaimActions.MapJsonKey("urn:yahoo:givenname", "givenName");
-            ClaimActions.MapJsonKey("urn:yahoo:profile", "profileUrl");
-            ClaimActions.MapJsonKey("urn:yahoo:profileimage", "imageUrl");
+            ClaimActions.MapJsonKey(YahooClaimTypes.FamilyName, "familyName");
+            ClaimActions.MapJsonKey(YahooClaimTypes.GivenName, "givenName");
```
khellang: Is there a reason a lot of these are using custom claim types instead of the built-in ones, like `ClaimTypes.GivenName`, `ClaimTypes.Surname`, etc.? Same in a lot of other places.

kevinchalet: Most likely an oversight (most of the aspnet-contrib providers have been written by different contributors, who may not be super familiar with all the built-in claim names).

FWIW, I try to avoid using the "legacy" WS-Fed claims as much as possible (because they're super long and make cookies/tokens bigger) but the MSFT OAuth2 providers use them when it's possible so we use them too for consistency.

khellang: Right, so as long as there's no transformations for them anywhere, it doesn't really matter? Or, it might even be better to use custom claim types because of the smaller size?

kevinchalet: Let's go with the built-in `ClaimTypes` when we have a good constant for these claims. Consistency always wins, I guess.

khellang: I didn't want to change any claim types in this PR, as it would be a breaking change. Should we open an issue for later?

kevinchalet:
@khellang FYI, we use a little different pattern in the other aspnet-contrib projects. E.g:

```csharp
public static class GitHubAuthenticationConstants
{
    public static class Claims
    {
        public const string Name = "name";
    }
}
```

Do you strongly prefer your pattern, or would adopting the one I suggest be acceptable for ya?

src/AspNet.Security.OAuth.Yahoo/YahooAuthenticationOptions.cs
kevinchalet: Breaking changes are kinda acceptable as we're still in RC. But yeah, separate PR if you prefer :+1:

khellang:
> Do you strongly prefer your pattern, or would adopting the one I suggest be acceptable for ya?

It's your project. I'm just following orders 😉It'll have to be another day, cause it's bedtime here 😴

kevinchalet:
> It's your project. I'm just following orders

Let's go with the nested classes pattern, then (I like when it's consistent).

> It'll have to be another day, cause it's bedtime here sleeping

Sure, no hurry.

Thanks for this PR! :clap:

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationConstants.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace AspNet.Security.OAuth.Autodesk
+{
+    /// <summary>
+    /// Contains constants specific to the <see cref="AutodeskAuthenticationHandler"/>.
+    /// </summary>
+    public static class AutodeskAuthenticationConstants
+    {
+        public static class Claims
+        {
+            public const string Emailverified = "urn:autodesk:emailverified";
```
poke: `EmailVerified`?

khellang: ah, thanks!

kevinchalet:
Merged, thanks! https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/commit/4f46357e4677a5a53955cacf7edd2d0387c60e05

PR #221, opened 2018-02-24
fix AppVeyor Building

kinosang:
relax SSL client settings to fix the CI failed


kevinchalet:
The AppVeyor folks are working on a global fix to force .NET apps to use TLS 1.2 in all their images. Let's wait a bit to see if that fixes this issue before updating all the aspnet-contrib repos... :trollface:

kinosang:
@PinpointTownes gotcha

kinosang:
replaced by #224

PR #226, opened 2018-03-11
Fix issue with Strava scopes

### LINKED ISSUE #225: Strava provider doesn't work with extra scopes
Using the Strava provider and adding the `write` scope

``` C#
.AddStrava(options =>
{
    options.ClientId = "xyz";
    options.ClientSecret = "xyzxyz";
    options.Scope.Add("write");
});
```

gives this auth url:

`https://www.strava.com/oauth/authorize?client_id=xyz&scope=public%20write&response_type=code&redirect_uri=xyz&state=xyz`

Which returns an `invalid scope` response due to two issues:
- Strava expects a comma-separated list of scopes
- Strava doesn't tolerate the `public` scope set if any other scope is also used

According to [the documentation](https://developers.strava.com/docs/authentication/), these are the only valid scope values:

`public`
`write`
`view_private`
`view_private,write`

anderaus:
Fixes #225

### DIFF
--- a/src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
@@ -4,6 +4,7 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
+using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
@@ -60,5 +61,7 @@ namespace AspNet.Security.OAuth.Strava
             await Options.Events.CreatingTicket(context);
             return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
         }
+
+        protected override string FormatScope() => !Options.Scope.Any() ? "public" : string.Join(",", Options.Scope);
     }
 }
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
```diff
@@ -60,5 +61,7 @@ public class StravaAuthenticationHandler : OAuthHandler<StravaAuthenticationOpti
             await Options.Events.CreatingTicket(context);
             return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
         }
+
+        protected override string FormatScope() => !Options.Scope.Any() ? "public" : string.Join(",", Options.Scope);
```
kevinchalet, line 65: The `string.Join(",", Options.Scope)` part is the right thing to do, but I'm not sure I like the first part (i.e only adding "public" is no explicit scope as manually added). I'll think more about it and come back to you.

kevinchalet:
Note: we'll also need to fix that in the 1.0 provider.

src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
anderaus: Was unsure about that one myself. The reason I added it was to maintain backwards compatibility since "public" is no longer set by default. I'll be happy to remove it, though.

kevinchalet: I opted for the approach I suggested.

kevinchalet:
Merged. Thanks for your contribution :+1:

PR #230, opened 2018-03-23
Added required version parameter to VK Api call

sla1k:
VK removed default api version from their calls and "v" parameter is required for all API calls.

See more: https://vk.com/dev/version_update

### DIFF
--- a/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
+++ b/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
@@ -16,5 +16,5 @@
     <PackageReference Include="JetBrains.Annotations" Version="$(JetBrainsVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
-  
+
 </Project>
--- a/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
+++ b/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
@@ -1,6 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="..\..\build\packages.props" />
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
@@ -13,8 +12,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="JetBrains.Annotations" Version="$(JetBrainsVersion)" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="JetBrains.Annotations" Version="11.1.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth" Version="2.0.3" />
   </ItemGroup>
 
 </Project>
--- a/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
+++ b/src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
@@ -1,5 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="..\..\build\packages.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


kevinchalet:
Thanks for your PR. It will be included in the 2.0 RTM bits and backported to 1.0.

src/AspNet.Security.OAuth.Vkontakte/AspNet.Security.OAuth.Vkontakte.csproj
```diff
@@ -16,5 +16,4 @@
     <PackageReference Include="JetBrains.Annotations" Version="$(JetBrainsVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.OAuth" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
-
```
kevinchalet: Doesn't look like a wanted change?

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
```diff
@@ -27,6 +28,11 @@ public class VkontakteAuthenticationHandler : OAuthHandler<VkontakteAuthenticati
             [NotNull] ISystemClock clock)
             : base(options, logger, encoder, clock)
         {
+            if (string.IsNullOrEmpty(Options.ApiVersion))
```
kevinchalet: Consider removing that.

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
```diff
@@ -38,6 +44,7 @@ public class VkontakteAuthenticationHandler : OAuthHandler<VkontakteAuthenticati
             {
                 address = QueryHelpers.AddQueryString(address, "fields", string.Join(",", Options.Fields));
             }
+            address = QueryHelpers.AddQueryString(address, "v", Options.ApiVersion);
```
kevinchalet: Perf nit: consider removing this line and doing that instead:

```csharp
var address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, new Dictionary<string, string>
{
    ["access_token"] = tokens.AccessToken,
    ["v"] = Options.ApiVersion ?? VkontakteAuthenticationDefaults.ApiVersion
});
```

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
```diff
@@ -61,9 +68,9 @@ public class VkontakteAuthenticationHandler : OAuthHandler<VkontakteAuthenticati
 
             var principal = new ClaimsPrincipal(identity);
             var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload);
-            context.RunClaimActions(payload);
+            context.RunClaimActions();
```
kevinchalet: Revert, for consistency with the other providers.

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
```diff
@@ -61,9 +68,9 @@ public class VkontakteAuthenticationHandler : OAuthHandler<VkontakteAuthenticati
...
 
-            await Options.Events.CreatingTicket(context);
+            await Events.CreatingTicket(context);
```
kevinchalet: :+1: but as part of a different PR that fixes all the providers at once.

Thanks.

kevinchalet (changes requested):

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
Vertigo093i: Consider moving that to a OAuthOptions.Validate() override. And do not remove this check entirely since the parameter is required by the VK API.

kevinchalet: That's why I suggested doing `["v"] = Options.ApiVersion ?? VkontakteAuthenticationDefaults.ApiVersion` to safely get rid of the check :smile:

Vertigo093i: This one doesn't guard against an empty string.

kevinchalet: `["v"] = !string.IsNullOrEmpty(Options.ApiVersion) ? Options.ApiVersion : VkontakteAuthenticationDefaults.ApiVersion` then.

sla1k:
@PinpointTownes  implement all comments

kevinchalet:
Merged.

Thanks for your contribution!

PR #217, opened 2018-02-15
Make sure all files end in a newline

khellang:
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

Just annoyed me a bit 😛

PR #235, opened 2018-04-17
Make sure all files end in a newline

khellang:
Same as #217, but for the release branch 😄


kevinchalet:
Merged, thanks for your contribution! :+1:

PR #239, opened 2018-05-19
Remove explicit reference to ASPNET Core 1.0 in headline

baronfel:
The deps are for 2.0 so let's just remove this so that future visitors aren't as confused as I was :)


kevinchalet:
Haha, thanks! :+1:

Issue #242, opened 2018-06-08
Twitch Provider not working 


zAfLu:
Since the 3789cfb36f1da6726c10aaec42538da2b986088b commit the Twitch Provider is broken. 

- Userendpoint should be "https://api.twitch.tv/helix/users/" 

- Claims are not working

- Scope should be Scope.Add("user:read:email");

kevinchalet:
Thanks for the PR. Please also send a PR against the release branch (for 1.0) and I’ll make sure it’s part of RTM.

/cc @kinosang

zAfLu:
Will try, need to merge my test setup to 1.0 :) Need some time. Will update his issue as fast as i can fix it

kevinchalet:
Thanks! FYI, we ship next week, so you still have some time left :smile:

kinosang:
Sorry for the mistakes. I didn't fully test the provider when I commit these changes.

zAfLu:
btw the twitch provider is tested in Core 2.1 ^^

zAfLu:
#244 tested with .Net Framework 4.5.1 

Our .Net Core 1.0 installation is broken so we couldn't test it :/

kevinchalet:
Fixed, thanks!

PR #243, opened 2018-06-08
Fix Twitch provider for the Helix API

zAfLu:
See #242

### DIFF
--- a/src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationConstants.cs
@@ -15,6 +15,8 @@ namespace AspNet.Security.OAuth.Twitch
             public const string Type = "urn:twitch:type";
             public const string BroadcasterType = "urn:twitch:broadcastertype";
             public const string Description = "urn:twitch:description";
+            public const string ProfileImageURL = "urn:twitch:profileimageurl";
+            public const string OfflineImageURL = "urn:twitch:offlineimageurl";
         }
     }
 }
--- a/src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationConstants.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AspNet.Security.OAuth.Twitch
+{
+    /// <summary>
+    /// Contains constants specific to the <see cref="TwitchAuthenticationHandler"/>.
+    /// </summary>
+    public static class TwitchAuthenticationConstants
+    {
+        public static class Claims
+        {
+            public const string DisplayName = "urn:twitch:displayname";
+            public const string Type = "urn:twitch:type";
+            public const string BroadcasterType = "urn:twitch:broadcastertype";
+            public const string Description = "urn:twitch:description";
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


kinosang:
LGTM

src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationOptions.cs
```diff
@@ -25,10 +25,28 @@ public TwitchAuthenticationOptions()
...
+            });
+
+            ClaimActions.MapCustomJson(ClaimTypes.Name, user =>
+            {
+                return user["data"]?[0]?["login"]?.ToObject<string>();
+            });
+
+            ClaimActions.MapCustomJson(ClaimTypes.GivenName, user =>
+            {
+                return user["data"]?[0]?["display_name"]?.ToObject<string>();
```
kevinchalet: Is the display name guaranteed to be the first/given name? If no, a different claim name should be used.

zAfLu: We discussed this too but we can not find a claim which really fits. 
```json
{
    "data": [
        {
            "id": "1111",
            "login": "zaflu",
            "display_name": "zAfLu",
            "type": "",
            "broadcaster_type": "partner",
            "description": "",
            "profile_image_url": "LINKTOIMAGE",
            "offline_image_url": "LINKTOIMAE",
            "view_count": 000000,
            "email": "xxxx@yyyy.yy"
        }
    ]
}
```
So the display_name shows the real write shape. 

What claim would you choose?

kevinchalet: Feel free to use a custom claim name. E.g https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.Bitbucket/BitbucketAuthenticationConstants.cs#L16

src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationConstants.cs
```diff
@@ -0,0 +1,22 @@
...
+    public static class TwitchAuthenticationConstants
+    {
+        public static class Claims
+        {
+            public const string DisplayName = "urn:twitch:displayname";
+            public const string Type = "urn:twitch:type";
+            public const string BroadcasterType = "urn:twitch:broadcastertype";
+            public const string Description = "urn:twitch:description";
+            public const string ProfileImageURL = "urn:twitch:profileimageurl";
+            public const string OfflineImageURL = "urn:twitch:offlineimageurl";
```
kevinchalet: `URL` -> `Url` please.

src/AspNet.Security.OAuth.Twitch/TwitchAuthenticationOptions.cs
```diff
@@ -25,10 +26,52 @@ public TwitchAuthenticationOptions()
...
             UserInformationEndpoint = TwitchAuthenticationDefaults.UserInformationEndpoint;
 
-            Scope.Add("user_read");
+            Scope.Add("user:read:email");
 
-            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "_id");
-            ClaimActions.MapJsonKey(ClaimTypes.Name, "name");
+            ClaimActions.MapCustomJson(ClaimTypes.NameIdentifier, user =>
+            {
+                return user["data"]?[0]?["id"]?.ToObject<string>();
```
kevinchalet: Consider using `.Value<string>("id")` instead of `ToObject()`.

kevinchalet:
@zAfLu when you're done, please also update the 1.0 PR so all the claims are consistent between the two versions.

zAfLu:
both updated :)

kevinchalet:
Merged, thanks for your contribution :+1:

PR #244, opened 2018-06-09
FIX Twitch Provider in 1.0

zAfLu:
See #242

### DIFF
--- a/src/AspNet.Security.OAuth.Twitch/AspNet.Security.OAuth.Twitch.csproj
+++ b/src/AspNet.Security.OAuth.Twitch/AspNet.Security.OAuth.Twitch.csproj
@@ -17,10 +17,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Remove="TwitchAuthenticationConstants.cs~RF3a3f7f.TMP" />
-  </ItemGroup>
-
-  <ItemGroup>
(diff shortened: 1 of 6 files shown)


src/AspNet.Security.OAuth.Twitch/AspNet.Security.OAuth.Twitch.csproj
```diff
@@ -16,6 +16,10 @@
     <Compile Include="..\..\shared\AspNet.Security.OAuth.Extensions\*.cs" />
   </ItemGroup>
 
+  <ItemGroup>
```
kinosang: remove this, keep the csproj file clear

kinosang: remove this, keep the csproj file clear

kinosang (changes requested):

src/AspNet.Security.OAuth.Twitch/AspNet.Security.OAuth.Twitch.csproj
zAfLu: Should be removed now

kevinchalet:
Merged, thanks for your contribution :+1:

PR #252, opened 2018-07-11
Dropbox user information endpoint

cristiz1992:
See [#251](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/251)


kevinchalet:
Thanks for your contribution! (and sorry for the long delay!)

PR #255, opened 2018-07-18
Add mapper for email to claims

NekitCorp:
https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/af1e3373cf00e82b186ea2a4bba0696d476b0dff/src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs#L69

At this line email field from payload has not been added to identity claims


kevinchalet:
Thanks for your contribution! (and sorry for the late reply, I've been busy with tons of other things :sweat_smile:)

PR #256, opened 2018-07-28
Add force_verify option to Twitch Provider

zAfLu:
Adding an option for the twitch force_verify parameter. 

You can enable or disable this option. I choose disabled as default value, so it works like the previous version. 
If you enable it, you always get the following window on every authorization request.

image

Test and build with Microsoft.AspNetCore.Authentication.OAuth 2.0.0


kevinchalet:
Thanks for your PR! :clap:

PR #172, opened 2017-08-21
Add weixin webpage provider

huww98:

PR #262, opened 2019-01-09
Increase support for authorization of Wechat Webpage

saber-wang:
增加对微信网页授权的支持
https://mp.weixin.qq.com/wiki?t=resource/res_main&id=mp1421140842

PR #266, opened 2019-01-12
Add WeChat Media Platform (aka WeChat Public Platform or Weixin Webpage) provider

kinosang:
Special thanks to @huww98 and his #172.

It's for Asp.Net Core 2.0 and support using UnionID as NameIdentifier.

### DIFF
--- a/src/AspNet.Security.OAuth.WeChat/StoreInCacheFormat.cs
+++ b/src/AspNet.Security.OAuth.WeChat/StoreInCacheFormat.cs
@@ -1,0 +1,59 @@
+﻿using Microsoft.Extensions.Caching.Memory;
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Authentication;
+
+namespace AspNet.Security.OAuth.WeChat
+{
+    internal class StoreInCacheFormat : ISecureDataFormat<AuthenticationProperties>
+    {
+        private readonly IMemoryCache _cache;
+        private readonly TimeSpan _timeout;
+
+        public StoreInCacheFormat(IMemoryCache cache, TimeSpan timeout)
+        {
+            _cache = cache;
+            _timeout = timeout;
+        }
+
+        public string Protect(AuthenticationProperties data)
+        {
+            return Protect(data, purpose: null);
+        }
+
+        public string Protect(AuthenticationProperties data, string purpose)
+        {
+            Guid guid = Guid.NewGuid();
+            var guidStr = guid.ToString();
+            string key = guidStr;
+            if (!string.IsNullOrEmpty(purpose))
+            {
+                key = purpose + key;
+            }
+
+            _cache.Set(key, data, _timeout);
+
+            return guidStr;
+        }
+
+        public AuthenticationProperties Unprotect(string protectedText)
+        {
+            return Unprotect(protectedText, purpose: null);
+        }
+
+        public AuthenticationProperties Unprotect(string protectedText, string purpose)
+        {
+            var key = protectedText;
+            if (!string.IsNullOrEmpty(purpose))
+            {
+                key = purpose + key;
+            }
+
+            var props = _cache.Get<AuthenticationProperties>(key);
+
+            //Weixin may return code multiple times
+            //copy props to prevent failure coursed by the changes to props.
+            return new AuthenticationProperties(new Dictionary<string, string>(props.Items));
+        }
+    }
+}
--- a/src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationExtensions.cs
+++ b/src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationExtensions.cs
@@ -8,6 +8,8 @@ using System;
 using AspNet.Security.OAuth.WeChat;
 using JetBrains.Annotations;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -70,6 +72,10 @@ namespace Microsoft.Extensions.DependencyInjection
             [NotNull] string scheme, [CanBeNull] string caption,
             [NotNull] Action<WeChatAuthenticationOptions> configuration)
         {
+            builder.Services.TryAddEnumerable(
+                ServiceDescriptor.Singleton<IPostConfigureOptions<WeChatAuthenticationOptions>,
+                    WeChatAuthenticationInitializer>());
+
             return builder.AddOAuth<WeChatAuthenticationOptions, WeChatAuthenticationHandler>(scheme, caption, configuration);
         }
     }
--- a/src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationExtensions.cs
+++ b/src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationExtensions.cs
@@ -1,0 +1,76 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using System;
+using AspNet.Security.OAuth.WeChat;
+using JetBrains.Annotations;
+using Microsoft.AspNetCore.Authentication;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods to add WeChat authentication capabilities to an HTTP application pipeline.
+    /// </summary>
+    public static class WeChatAuthenticationExtensions
+    {
+        /// <summary>
+        /// Adds <see cref="WeChatAuthenticationHandler"/> to the specified
+        /// <see cref="AuthenticationBuilder"/>, which enables WeChat authentication capabilities.
+        /// </summary>
+        /// <param name="builder">The authentication builder.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static AuthenticationBuilder AddWeChat([NotNull] this AuthenticationBuilder builder)
+        {
+            return builder.AddWeChat(WeChatAuthenticationDefaults.AuthenticationScheme, options => { });
+        }
+
+        /// <summary>
+        /// Adds <see cref="WeChatAuthenticationHandler"/> to the specified
+        /// <see cref="AuthenticationBuilder"/>, which enables WeChat authentication capabilities.
+        /// </summary>
+        /// <param name="builder">The authentication builder.</param>
+        /// <param name="configuration">The delegate used to configure the OpenID 2.0 options.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static AuthenticationBuilder AddWeChat(
+            [NotNull] this AuthenticationBuilder builder,
+            [NotNull] Action<WeChatAuthenticationOptions> configuration)
+        {
+            return builder.AddWeChat(WeChatAuthenticationDefaults.AuthenticationScheme, configuration);
+        }
+
+        /// <summary>
+        /// Adds <see cref="WeChatAuthenticationHandler"/> to the specified
+        /// <see cref="AuthenticationBuilder"/>, which enables WeChat authentication capabilities.
+        /// </summary>
+        /// <param name="builder">The authentication builder.</param>
+        /// <param name="scheme">The authentication scheme associated with this instance.</param>
+        /// <param name="configuration">The delegate used to configure the WeChat options.</param>
+        /// <returns>The <see cref="AuthenticationBuilder"/>.</returns>
+        public static AuthenticationBuilder AddWeChat(
+            [NotNull] this AuthenticationBuilder builder, [NotNull] string scheme,
+            [NotNull] Action<WeChatAuthenticationOptions> configuration)
+        {
+            return builder.AddWeChat(scheme, WeChatAuthenticationDefaults.DisplayName, configuration);
+        }
+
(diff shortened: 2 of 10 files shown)


src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+            ClaimActions.MapCustomJson(Claims.Privilege, user =>
+            {
+                var value = user.Value<JArray>("privilege");
+                if (value == null)
+                {
+                    return null;
+                }
+
+                return string.Join(",", value.ToObject<string[]>());
+            });
```
John0King, line 52: Should here use `MapJsonKey` ?  
`MapJsonKey` will add multiple claims if the value is a `JArray` , the problem here is if the priviliege contains ","  then you can not map it back to array again

src/AspNet.Security.OAuth.WeChat/StoreInCacheFormat.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Microsoft.Extensions.Caching.Memory;
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Authentication;
+
+namespace AspNet.Security.OAuth.WeChat
+{
+    internal class StoreInCacheFormat : ISecureDataFormat<AuthenticationProperties>
```
martincostello, line 8: What's the primary motivation for the need to add caching into this provider? The only reason I ask is because none of the existing providers do this.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <param name="scheme">The authentication scheme associated with this instance.</param>
+        /// <param name="caption">The optional display name associated with this instance.</param>
+        /// <param name="configuration">The delegate used to configure the WeChat options.</param>
+        /// <returns>The <see cref="AuthenticationBuilder"/>.</returns>
+        public static AuthenticationBuilder AddWeChat(
+            [NotNull] this AuthenticationBuilder builder,
+            [NotNull] string scheme, [CanBeNull] string caption,
+            [NotNull] Action<WeChatAuthenticationOptions> configuration)
+        {
+            builder.Services.TryAddEnumerable(
```
martincostello, line 75: Why not use `TryAddSingleton()`?

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            [NotNull] UrlEncoder encoder,
+            [NotNull] ISystemClock clock)
+            : base(options, logger, encoder, clock)
+        {
+        }
+
+        protected override async Task<AuthenticationTicket> CreateTicketAsync([NotNull] ClaimsIdentity identity, [NotNull] AuthenticationProperties properties, [NotNull] OAuthTokenResponse tokens)
+        {
+            var payload = new JObject();
+            switch (Options.Scope.SingleOrDefault())
```
martincostello, line 38: Is there a specific reason to use `Single...()` here? If more than one were returned the integrator would just get a vague error message with an `InvalidOperationException` telling them a sequence contained more than one element, which would be hard to diagnose.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                case "snsapi_base":
+                    payload["openid"] = tokens.Response.Value<string>("openid");
+                    break;
+                case "snsapi_userinfo":
+                    var address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint,
```
martincostello, line 44: There's a lot of code here for a switch statement case. It would be better factored out into its own method.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+                    }
+
+                    payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+                    if (!string.IsNullOrEmpty(payload.Value<string>("errcode")))
+                    {
+                        Logger.LogError("An error occurred while retrieving the user profile: the remote server " +
+                                        "returned a {Status} response with the following payload: {Headers} {Body}.",
+                            /* Status: */ response.StatusCode,
+                            /* Headers: */ response.Headers.ToString(),
+                            /* Body: */ await response.Content.ReadAsStringAsync());
```
martincostello, line 70: Content is being re-read here. If the stream was read only it could fail.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                        throw new HttpRequestException("An error occurred while retrieving user information.");
+                    }
+
+                    payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+                    if (!string.IsNullOrEmpty(payload.Value<string>("errcode")))
+                    {
+                        Logger.LogError("An error occurred while retrieving the user profile: the remote server " +
```
martincostello, line 66: This code block is a duplicate of the one above.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+                            /* Status: */ response.StatusCode,
+                            /* Headers: */ response.Headers.ToString(),
+                            /* Body: */ await response.Content.ReadAsStringAsync());
+
+                        throw new HttpRequestException("An error occurred while retrieving user information.");
+                    }
+
+                    break;
+                default:
+                    throw new InvalidOperationException("Invalid scope");
```
martincostello, line 77: It would be useful to include what the invalid scope was in the exception message.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+            }
+
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+            if (!string.IsNullOrEmpty(payload.Value<string>("errcode")))
+            {
+                Logger.LogError("An error occurred while retrieving an access token: the remote server " +
```
martincostello, line 114: Same comments as above.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            var response = await Backchannel.GetAsync(address);
+            if (!response.IsSuccessStatusCode)
+            {
+                Logger.LogError("An error occurred while retrieving an access token: the remote server " +
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                /* Status: */ response.StatusCode,
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
```
martincostello, line 108: `Exception` shouldn't really be used to throw exceptions as it's so non-specific. Is there a better way you can achieve this?

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+            if (!string.IsNullOrEmpty(payload.Value<string>("errcode")))
+            {
+                Logger.LogError("An error occurred while retrieving an access token: the remote server " +
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                /* Status: */ response.StatusCode,
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
```
martincostello, line 120: As-per above comment.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+        protected override string BuildChallengeUrl(AuthenticationProperties properties, string redirectUri)
+        {
+            return QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, new Dictionary<string, string>
+            {
+                { "appid", Options.ClientId },
+                { "redirect_uri", redirectUri },
+                { "response_type", "code" },
+                { "scope", FormatScope() },
+                { "state", Options.StateDataFormat.Protect(properties) }
+            }) + "#wechat_redirect" /*required*/;
```
martincostello, line 134: Link to documentation or something here to explain what it's for?

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        protected override string FormatScope()
+        {
+            if (Options.Scope.Count > 1)
+            {
+                throw new InvalidOperationException("WeChat MP OAuth does not accept multiple scopes");
+            }
+
+            return Options.Scope.SingleOrDefault();
```
martincostello, line 144: `FirstOrDefault()`?

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationInitializer.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Creates a new instance of the <see cref="WeChatAuthenticationInitializer"/> class.
+        /// </summary>
+        public WeChatAuthenticationInitializer([NotNull] IMemoryCache cache)
+        {
+            _cache = cache;
+        }
+
+        /// <summary>
+        /// A ugly hack to deal with WeChat MP OAuth state length limitation.
```
martincostello, line 21: This is an internal implementation detail, so shouldn't be in public-facing XML Intellisense comments.

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+            ClaimActions.MapJsonKey(Claims.HeadImgUrl, "headimgurl");
+            ClaimActions.MapCustomJson(Claims.Privilege, user =>
+            {
+                var value = user.Value<JArray>("privilege");
+                if (value == null)
+                {
+                    return null;
+                }
+
+                return string.Join(",", value.ToObject<string[]>());
```
martincostello, line 51: `return string.Join(",", value.Values<string>());`?

src/AspNet.Security.OAuth.WeChat/WeChatAuthenticationOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Use UnionID as NameIdentifier.
+        /// </summary>
+        /// <remarks>
+        /// Enable it only when you have your WeChat MP account linked to a WeChat Open Platform account.
+        /// </remarks>
+        public void UseUnionId()
+        {
+            if (Scope.All(s => s != "snsapi_userinfo"))
+            {
+                throw new InvalidOperationException("Invalid scope");
```
martincostello, line 65: Again, exception message is vague and doesn't help lead the integrator into the pit of success.

martincostello (commented):
I'm not familiar with WeChat Media Platform - in what way does it differ and need its own provider compared to our QQ and Weixin providers?

src/AspNet.Security.OAuth.WeChat/StoreInCacheFormat.cs
kinosang: @martincostello WeChat has a limitation on `state`, the max length is 128 bytes, so we can't use the default DataProtection state.

martincostello: In that case we might need to use `IDistributedCache`, otherwise this wouldn't cache property in a web farm deployment?

kinosang:
@saber-wang maybe we can merge the two provider (WeChat WebPage and WeChat QrConnect) and provide methods such as `.UseQrConnect()` and `.UseWebPage()`, or provide a property `Type` just like `Region` in the BattleNet provider.

Also we can keep `.UseUnionId()` to use the UnionId for NameIdentifier.

saber-wang:
Yes, but I don't think a new provider is needed because the biggest difference between / authorize and / qrconnect is the state length limit.

kinosang:
@saber-wang I mean I'll close this pr and we can just modify the `AspNet.Security.OAuth.Weixin` provider.

saber-wang:
@kinosang That's good.

martincostello:
#262 has been merged to `dev` now. If you all could work out the best approach to move forward this PR that would be great.

The preferred option would be one provider that supports the different use cases if it's just a few parameter tweaks, as then there's less code to maintain overall.

kinosang:
@martincostello I'll make the changes upon #262 and the Wexin provider asap.

PR #268, opened 2019-01-13
Refactored handlers to explicitly dispose resources

JuliaPileva:
Also added CodeRush to ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,6 @@ $RECYCLE.BIN/
 
 # Mac desktop service store files
 .DS_Store
+
+# CodeRush
+.cr/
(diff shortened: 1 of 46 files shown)


martincostello:
Thank you for your PR @JuliaPileva!

I've raised aspnet/AspNetCore#10229 based on this to ask for some guidance whether this is strictly necessary or not, as the built-in ASP.NET Core OAuth handlers do not do this themselves either, which is I guess why the ones here don't too.

Once there's an answer there, we can look at where to go further with this PR.

kevinchalet:
@martincostello if we do that, I suggest we opt for the new `using var ...` syntax, that would make the code so much nicer :smile:

martincostello:
@JuliaPileva The ASP.NET Core team don't think this is necessary https://github.com/aspnet/AspNetCore/issues/10229#issuecomment-492300332, so I'd err on the side of not accepting this PR at the moment due to the amount of code churn it introduces.

However, it looks like it would be a lot easier to do after #280 when the `using var ...` syntax could be used instead to make the change less of a code change.

martincostello:
Hi @JuliaPileva, thanks again for your contribution. I've followed the ASP.NET Core team's suggestion, and resolved this by using the new C# 8.0 syntax in #280.

This achieves the same objective as this PR, but with a lot less code churn, so I'm going to close this one.

Issue #275, opened 2019-02-06
Newtonsoft.Json types replaced in Microsoft.AspNetCore.Authentication APIs


kinosang:
aspnet/Announcements#341

> For derived OAuth implementations the most common change is to replace JObject.Parse with JsonDocument.Parse in the CreateTicketAsync override as shown [here](https://github.com/aspnet/AspNetCore/pull/7105/files?utf8=%E2%9C%93&diff=unified&w=1#diff-e1c9f9740a6fe8021020a6f249c589b0L40). Note JsonDocument is IDisposable.

martincostello:
Updating to 3.0 preview 3, the auth providers now fail with an exception such as:

```
System.MissingMethodException: Method not found: 'Void Microsoft.AspNetCore.Authentication.OAuth.OAuthCreatingTicketContext..ctor(System.Security.Claims.ClaimsPrincipal, Microsoft.AspNetCore.Authentication.AuthenticationProperties, Microsoft.AspNetCore.Http.HttpContext, Microsoft.AspNetCore.Authentication.AuthenticationScheme, Microsoft.AspNetCore.Authentication.OAuth.OAuthOptions, System.Net.Http.HttpClient, Microsoft.AspNetCore.Authentication.OAuth.OAuthTokenResponse, Newtonsoft.Json.Linq.JObject)'.
   at AspNet.Security.OAuth.Amazon.AmazonAuthenticationHandler.CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at AspNet.Security.OAuth.Amazon.AmazonAuthenticationHandler.CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
   at Microsoft.AspNetCore.Authentication.OAuth.OAuthHandler`1.HandleRemoteAuthenticateAsync()
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
```

martincostello:
Have started a migration for some of the providers in #280.

martincostello:
Will be resolved by #280.

[...]

PR #280, opened 2019-03-06
Support ASP.NET Core 3.0

martincostello:
Adds support for ASP.NET Core 3.0 to ~various simple~ all providers.

~So far, this PR has only updated providers that had one explicit usage of Newtonsoft.Json. All others have been left targeting .NET Core 2.0 for now.~

~There's lots of `#if` and copied code here - I'm sure there's a better way it could be factored.~

Relates to #275.


kevinchalet:
Thanks for your PR!

It looks good but let's just get rid of the `netstandard20` TFM: ASP.NET Core 3.0 will be .NET Core-only so there's no point trying to support .NET Standard (and having 2 TFM that target different versions is a terrible recipe for a disaster 😅)

martincostello:
Great, that’s much easier. I assumed you’d want to keep both so fixed could be back-ported to 2.0 users, but if that’s not a problem than that makes this much easier 😃

martincostello:
This should be mostly functional for ASP.NET Core 3.0 now, based on preview 3. I can't test all the providers, but most are simple migrations based on https://github.com/aspnet/AspNetCore/pull/7105.

However, there's four bits that need further review:
  1. Adding the email to the payload for VKontakte: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/24f548328ccd81a15a5881bec4b8ced59ed60d88/src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs#L67-L68
  1. Determine the best way to copy the tokens for Yammer: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/24f548328ccd81a15a5881bec4b8ced59ed60d88/src/AspNet.Security.OAuth.Yammer/YammerAuthenticationHandler.cs#L102-L109
  1. Determine the best way to copy the tokens for StackExchange: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/24f548328ccd81a15a5881bec4b8ced59ed60d88/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationHandler.cs#L102-L109
  1. Determine the best way to copy the tokens for QQ: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/24f548328ccd81a15a5881bec4b8ced59ed60d88/src/AspNet.Security.OAuth.QQ/QQAuthenticationHandler.cs#L118-L125

martincostello:
I've tested the updated Amazon authentication handler with the 3.0 branch of my site (https://github.com/martincostello/alexa-london-travel-site/pull/255), and things are working as expected.

niltor:
When it can be updated to nuget?

kevinchalet:
> However, there's four bits that need further review.

/cc @Tratcher @ahsonkhan could you please shed some light on this? 

> When it can be updated to nuget?

We'll likely make them available via the MyGet.org feed but I wouldn't expect a public release (i.e on NuGet.org) before 3.0 general availability.

martincostello:
@PinpointTownes FYI I’m working on a separate PR that I’ll submit for separate review at the weekend to add some integration tests for a couple of the providers as a proof-of-concept. If that’s acceptable when I submit it I’ll add it to this PR so we’re confident the changes are all working as I can only easily test the Amazon one myself.

martincostello:
Draft tests PR for review opened in #282.

martincostello:
Excluding any code improvements that could be made as referenced by my earlier https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/280#issuecomment-471171850, I believe this PR is now feature complete.

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
```diff
@@ -56,20 +57,27 @@ public class VkontakteAuthenticationHandler : OAuthHandler<VkontakteAuthenticati
...
-            context.RunClaimActions(payload);
+                var principal = new ClaimsPrincipal(identity);
+                var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload);
+                context.RunClaimActions();
 
-            await Options.Events.CreatingTicket(context);
-            return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
+                if (tokens.Response.RootElement.TryGetProperty("email", out var email))
+                {
+                    identity.AddClaim(new Claim(ClaimTypes.Email, email.GetString(), ClaimValueTypes.String, ClaimsIssuer));
```
Tratcher: How about calling RunClaimActions again and pass in `tokens.Response.RootElement`?

src/AspNet.Security.OAuth.Yammer/YammerAuthenticationHandler.cs
```diff
@@ -89,13 +93,47 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] st
...
+
+            stream.Seek(0, SeekOrigin.Begin);
+        }
+
+        private void WriteAccessToken(string token, IBufferWriter<byte> bufferWriter)
+        {
+            var writer = new Utf8JsonWriter(bufferWriter);
+
+            writer.WriteStartObject();
+            writer.WriteString("access_token", token);
```
Tratcher: @bartonjs is there an easier way to construct a JsonDocument from a fixed set of keys & values other than fully serializing and re-parsing it?

bartonjs: > is there an easier way to construct a JsonDocument from a fixed set of keys & values other than fully serializing and re-parsing it?

Not currently, no.  It doesn't have the ability to create JSON, it just understands how to cursor over the text (semi-)intelligently.

src/AspNet.Security.OAuth.QQ/QQAuthenticationHandler.cs
```diff
@@ -110,10 +114,16 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] st
...
-                .ToDictionary(pair => pair.Key, k => k.Value.ToString()));
+            var content = QueryHelpers.ParseQuery(await response.Content.ReadAsStringAsync());
 
-            return OAuthTokenResponse.Success(payload);
+            // HACK Work out the best way to do this with System.Text.Json
+            using (var stream = new MemoryStream())
+            {
+                await CopyPayloadAsync(content, stream);
+
+                var payload = JsonDocument.Parse(stream);
```
bartonjs: Is this ever disposed of?

src/AspNet.Security.OAuth.QQ/QQAuthenticationHandler.cs
```diff
@@ -141,11 +151,38 @@ private async Task<string> GetUserIdentifierAsync(OAuthTokenResponse tokens)
                 body = body.Substring(index, body.LastIndexOf("}") - index + 1);
             }
 
-            var payload = JObject.Parse(body);
+            var payload = JsonDocument.Parse(body);
```
bartonjs: This should be in a using, since it doesn't need to outlive the GetString call

src/AspNet.Security.OAuth.QQ/QQAuthenticationHandler.cs
Tratcher: Yes, the caller will dispose it via the OAuthTokenResponse.

martincostello:
Thanks for the review @Tratcher @bartonjs, I’ll push up some changes tomorrow.

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationHandler.cs
martincostello: That was much simpler 😃

src/AspNet.Security.OAuth.QQ/QQAuthenticationHandler.cs
martincostello: Fixed.

martincostello:
@Tratcher I've updated the hacky "copy" code to use some of the new preview 5 APIs in 9a80d10c494ff5d505bf427e72dd2bc6f0c508fb. If you get some time, could you take a quick look to check that the approach is a valid usage of the new bits?

Tratcher:
Looks cleaner. @brennanconroy?

kevinchalet:
@Tratcher are there plans to eventually add less hackish APIs to support creating JSON constructs without a backing array (a-la `JObject`)?

I hate to be that guy, but I'm sure that `System.Xml.Linq` (and even `System.Xml` 😅) currently offer a better experience for these scenarios.

BrennanConroy:
Yeah looks good.

martincostello:
@Tratcher @BrennanConroy Thanks both.

Tratcher:
@PinpointTownes I completely agree and have given the Json folks that feedback.

The other possibility here is that this isn't the right datatype to expose in the API, we should have used something more abstract that you could implement using a JsonDocument or your own underlying structure. I hesitate to invent a new abstraction though, and I wouldn't want to use an existing one like nested `IDictionary<string, object>`. Open to suggestions.

Tratcher:
Filed at https://github.com/dotnet/corefx/issues/37565

kevinchalet:
@Tratcher @rynowak excellent, thanks!

martincostello:
Need to fix this either now or when preview 6 arrives: https://github.com/aspnet/AspNetCore/issues/10615#issuecomment-497014985

martincostello:
@PinpointTownes This PR is getting quite large and unwieldy now, and there's still at least 2 more previews to go until RTM. I was thinking about creating a `dev/3.0.0` branch off master, re-targeting this PR to it and then merging.

Then I can do PRs for each preview into that branch until 3.0 ships, and then do a final squash PR back to dev when the time comes. The added benefit of that approach is that 3.0 preview supporting packages get published to MyGet.

Thoughts?

kevinchalet:
Do you still expect major changes in the 2.x packages? If not, we should probably simply merge this PR into dev and forget about 2.x stuff. If we need to release patches, we can simply use the `2.1.0` tag and create a new patch branch from there.

martincostello:
I'm bit wary about making the main dev branch be against a preview version, rather than a stable one.

Otherwise it might discourage the community from submitting new providers as they'd have to work against preview API surfaces they might not know as much about (e.g. the new JSON stuff) for the next 4 months until 3.0 RTM lands.

PR #281, opened 2019-03-10
Fix Stack Exchange authentication handler

poke:
The payload from the SE API contains an `items` array which contains a single user object that contains all the claims. As such, the authentication handler should extract that user object already and run the claims actions on that single user object.

Previously, the array was attempted to be parsed as an object (which failed) and the built-in claims mapping did expect the result to contain the raw payload (with the `items` array).

Also, although specifying a `key` is recommended for Stack Exchange to receive higher request quotas, it is not a mandatory argument to pass in the API.


kevinchalet:
Thanks for the PR!
Did you take a look at the 1.x provider? If so, is it also impacted? (just to know if we should backport this fix).

poke:
No, I didn't check that, but I believe this was broken in the update for 2.0: 55172529350ce37a587e5507312f5e38b1ea38df

martincostello:
I've also ported this fix to the PR for .NET Core support: #280

martincostello (approved):

Issue #277, opened 2019-02-17
Update Linkedin to support api/v2


kchowanski:
Hello,
Linkedin enforce users to migrate to API version 2.0 by March 1.

"...all developers need to migrate to Version 2.0 of our APIs and OAuth 2.0 by March 1."

https://engineering.linkedin.com/blog/2018/12/developer-program-updates.

Mattzard:
Whilst this work is being completed, I don't know if it would be recommended to also make the call to return the email address within the claims? 

For version 2 of the API email address has been separated into a different call (_https://api.linkedin.com/v2/emailAddress?q=members&projection=(elements*(handle~))_) and isn't returned with the new _lite_profile_, yet previously it would have been returned in the claims for version 1 of the API so may aid people who are using that already.

martincostello:
This sounds similar to how the GitHub and Bitbucket providers work, where emails need to be requested separately with a different HTTP request.

jandieg:
Hi, is this related? Just using a new LinkedIn api (v2)

> An unhandled exception occurred while processing the request.
> HttpRequestException: An error occurred while retrieving the user profile.
> AspNet.Security.OAuth.LinkedIn.LinkedInAuthenticationHandler.CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
> 
> Exception: An error was encountered while handling the remote login.
> Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler<TOptions>.HandleRequestAsync()

odahcam:
I'm having some trouble to use the LinkeIn authentication as well. In my case, when the user comes back from the LinkedIn's authorization page to my route callback, the first thing I'm doing is:

```csharp
var loginInfo = await AuthenticationManager.GetExternalLoginInfoAsync();
```

But this method call, returns `null` for LinkedIn while works perfectly with Twitter, Google SignIn, Facebook and Instagram.

Are you here with the same problem?

jandieg:
I think it is part of the same. Current implementations assume the **email** is coming as a claim, but on Linkedin API v2 it is no longer.

@devs is there a workaround for this? At least to ignore the email claim?

odahcam:
As I could see `loginInfo` is completely empty and `nameidentifier` is `0`. 😕

Mattzard:
[...]

PR #298, opened 2019-05-21
Fix LinkedIn provider with v2 endpoints to get name, email and profile pictures

asiffermann:
As mentioned in #277, Linkedin has enforced users to migrate to API version 2.0 by May 1.

The [migration FAQ](https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq?context=linkedin/consumer/context) describes the main differences:

- Usage of the Lite Profile consisting of a member's id, first name, last name, and profile picture, instead of the previous Basic Profile which contained much more informations
- A dedicated API endpoint to retrieve a member's email
- OAuth scopes to include to call those endpoints: r_liteprofile and r_emailaddress
- Changes for the response JSON payload

This pull request implements all those changes. Feel free to ask me any changes you'd like!


src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,147 +13,38 @@ public static class LinkedInAuthenticationConstants
...
+        public const string EmailAddressField = "emailAddress";
 
-        // https://developer.linkedin.com/docs/fields/basic-profile
+        // https://docs.microsoft.com/en-us/linkedin/shared/references/v2/profile/lite-profile?context=linkedin/consumer/context
         public static class ProfileFields
         {
             /// <summary>
-            /// A unique identifying value for the member.
-            ///This value is linked to your specific application.Any attempts to use it with a different application will result in a "404 - Invalid member id" error.
+            /// The unique identifier for the given member. May also be referenced as the personId within a Person URN (urn:li:person:{personId}).
```
martincostello: Can you wrap the _code-y_ indentifiers in `<c>` tags please?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,147 +13,38 @@ public static class LinkedInAuthenticationConstants
...
+            /// The id is unique to your specific developer application. Any attempts to use the id with other developer applications will not succeed.
             /// </summary>
             public const string Id = "id";
 
             /// <summary>
-            /// The member's first name.
+            /// First name of the member. Represented as a MultiLocaleString object type.
+            /// See https://docs.microsoft.com/en-us/linkedin/shared/references/v2/object-types#multilocalestring
```
martincostello: Make the URL a link as either an `<a>` or a `<see>`.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,147 +13,38 @@ public static class LinkedInAuthenticationConstants
...
             /// </summary>
-            public const string FirstName = "first-name";
+            public const string FirstName = "firstName";
 
             /// <summary>
-            /// The member's last name.
+            /// Last name of the member. Represented as a MultiLocaleString object type.
+            /// See https://docs.microsoft.com/en-us/linkedin/shared/references/v2/object-types#multilocalestring
```
martincostello: As above.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,147 +13,38 @@ public static class LinkedInAuthenticationConstants
...
             /// </summary>
-            public const string LastName = "last-name";
+            public const string LastName = "lastName";
 
             /// <summary>
-            /// The member's name, formatted based on language.
+            /// Metadata about the member's picture in the profile. See Profile Picture Fields for more information.
+            /// See https://docs.microsoft.com/en-us/linkedin/shared/references/v2/profile/profile-picture
```
martincostello: As above.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationDefaults.cs
```diff
@@ -46,9 +46,13 @@ public class LinkedInAuthenticationDefaults
...
         /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
-        /// Note: the endpoint must follow the LinkedIn convention and contain a '~' to append fields to, if they are specified.
-        /// See https://developer.linkedin.com/docs/signin-with-linkedin for more information.
         /// </summary>
-        public const string UserInformationEndpoint = "https://api.linkedin.com/v1/people/~";
+        public const string UserInformationEndpoint = "https://api.linkedin.com/v2/me";
+
+        /// <summary>
+        /// Specific endpoint to retrieve the LinkedIn member's email address.
+        /// See https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin for more information.
```
martincostello: As above. (I realise they were this way already, but new eyes and diff views highlight things 😃)

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -33,12 +34,13 @@ public class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthentication
...
             var address = Options.UserInformationEndpoint;
+            var fields = Options.Fields.Where(f => f != LinkedInAuthenticationConstants.EmailAddressField).ToList();
 
             // If at least one field is specified,
             // append the fields to the endpoint URL.
-            if (Options.Fields.Count != 0)
+            if (fields.Count != 0)
             {
-                address = address.Insert(address.LastIndexOf("~") + 1, $":({ string.Join(",", Options.Fields)})");
+                address = $"{address}?projection=({string.Join(",", fields)})";
```
martincostello: If for some reason `UserInformationEndpoint` already has a query string in it, this would break because it would try and add another. You should use a `UriBuilder` or the `AddQueryString(...)` helper used elsewhere to append the extra query string parameter.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -58,6 +60,32 @@ public class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthentication
             }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+
+            if (Options.Fields.Contains(LinkedInAuthenticationConstants.EmailAddressField))
```
martincostello, line 72: Should this `Contains()` check maybe be case-insensitive?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -58,6 +60,32 @@ public class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthentication
...
+            {
+                var emailAddressRequest = new HttpRequestMessage(HttpMethod.Get, Options.EmailAddressEndpoint);
```
martincostello: Can you factor out this additional HTTP request into a private method please? I believe the GitHub provider does a similar thing as it also does an extra call to fetch the email address(es).

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -58,6 +60,32 @@ public class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthentication
...
+                                    "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                    /* Status: */ emailAddressResponse.StatusCode,
+                                    /* Headers: */ emailAddressResponse.Headers.ToString(),
+                                    /* Body: */ await emailAddressResponse.Content.ReadAsStringAsync());
+
+                    throw new HttpRequestException("An error occurred while retrieving the email address.");
+                }
+
+                var emailAddressPayload = JObject.Parse(await emailAddressResponse.Content.ReadAsStringAsync());
+                var emailAddress = emailAddressPayload.SelectToken("elements[0].handle~.emailAddress").ToString();
```
martincostello: Could you do this without using `SelectToken(...)`? This won't exist in .NET Core 3.0 with the Newtonsoft.Json dependency removed (see #280), so will make it easier to port over to the new JSON APIs.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,77 @@ public LinkedInAuthenticationOptions()
...
-            ClaimActions.MapJsonKey(Claims.PhoneticFirstName, "phoneticFirstName");
-            ClaimActions.MapJsonKey(Claims.PhoneticLastName, "phoneticLastName");
-            ClaimActions.MapJsonKey(Claims.FormattedPhoneticName, "formattedPhoneticName");
-            ClaimActions.MapCustomJson(Claims.Location, user => user["location"]?.ToString());
-            ClaimActions.MapJsonKey(Claims.Specialties, "specialties");
-            ClaimActions.MapJsonKey(Claims.NumConnections, "numConnections");
-            ClaimActions.MapJsonKey(Claims.NumConnectionsCapped, "numConnectionsCapped");
-            ClaimActions.MapJsonKey(Claims.CurrentShare, "currentShare");
-            ClaimActions.MapCustomJson(Claims.PictureUrls, user => user["pictureUrls"]?.ToString());
+            ClaimActions.MapCustomJson(ClaimTypes.Name, user => $"{GetMultiLocaleString(user, "firstName")} {GetMultiLocaleString(user, "lastName")}");
```
martincostello: This would create trailing spaces if a person only had one name.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,77 @@ public LinkedInAuthenticationOptions()
...
+            ClaimActions.MapCustomJson(ClaimTypes.GivenName, user => GetMultiLocaleString(user, "firstName"));
+            ClaimActions.MapCustomJson(ClaimTypes.Surname, user => GetMultiLocaleString(user, "lastName"));
+            ClaimActions.MapCustomJson(Claims.PictureUrl, user
+                => user.SelectTokens("$.profilePicture..elements[*].identifiers[0].identifier").Select(u => u.Value<string>()).LastOrDefault());
```
martincostello: Again regarding use of `SelectTokens(...)` here and on the line below.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,77 @@ public LinkedInAuthenticationOptions()
...
 
+        /// <summary>
+        /// Gets or sets the email address endpoint.
+        /// </summary>
+        public string EmailAddressEndpoint { get; set; }
+
         /// <summary>
         /// Gets the list of fields to retrieve from the user information endpoint.
-        /// See https://developer.linkedin.com/docs/fields/basic-profile for more information.
+        /// See https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin for more information.
```
martincostello: `<a>` or `<see>` for the link.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,77 @@ public LinkedInAuthenticationOptions()
...
         {
             ProfileFields.Id,
             ProfileFields.FirstName,
             ProfileFields.LastName,
-            ProfileFields.FormattedName,
             LinkedInAuthenticationConstants.EmailAddressField
         };
+
+        /// <summary>
+        /// Gets the MultiLocaleString value.
```
martincostello: Same comments about tags for code and URLs.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,77 @@ public LinkedInAuthenticationOptions()
...
+            {
+                var preferredKey = $"{preferredLocale["language"]}_{preferredLocale["country"]}";
+                var preferredLocalizedValue = user[propertyName][localizedKey][preferredKey];
+                if (preferredLocalizedValue != null)
+                {
+                    return preferredLocalizedValue.Value<string>();
+                }
+            }
+
+            var currentUiKey = Thread.CurrentThread.CurrentUICulture.ToString().Replace('-', '_');
```
martincostello: I wonder whether we should put all of them into claims and let the integrator decide, rather than guessing at the behaviour they might want (maybe they don't use the current server thread to make UI decisions)?

test/AspNet.Security.OAuth.Providers.Tests/LinkedIn/bundle.json
```diff
@@ -12,20 +12,48 @@
...
+          "localized": {
+            "en_US": "Baggins"
+          },
+          "preferredLocale": {
+            "country": "US",
+            "language": "en"
+          }
+        },
+        "profilePicture": {
+          "displayImage": "urn:li:digitalmediaAsset:B54328XZFfe2134zTyq"
```
martincostello: If this is being mapped, can you add it as a claims test case?

martincostello:
Thanks for the contribution @asiffermann - as this includes code breaking-changes, this is probably not going to be released as part of any 2.x release (if we do one) as it would need to be 3.0 for SemVer, and not shipping ASP.NET Core 3.0 support in the 3.0 release would be confusing.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
asiffermann: Done

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
asiffermann: Good point, I'm using `AddQueryString(...)` now

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
asiffermann: Good point too! I am now using the `HashSet(System.Collections.Generic.IEqualityComparer<T> comparer)` constructor to initialize `Fields` property

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
asiffermann: I did it like the GitHub provider, with a protected virtual method

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
asiffermann: I changed it to use Linq syntax, as other providers

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: It is now resilient and I have extracted the code in a private method

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: Same as above, I extracted this in a private static method using Linq syntax

test/AspNet.Security.OAuth.Providers.Tests/LinkedIn/bundle.json
asiffermann: I added a unit test to check this claim, using the sample JSON provided in LinkedIn documentation

martincostello:
Other than my comment regarding localisation, this looks good! 😎

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: From my tests, LinkedIn always provides the `preferredLocale` property, but as [it is documented](https://docs.microsoft.com/en-us/linkedin/shared/references/v2/object-types#multilocalestring) as non-required I wanted to implement a fallback, but it will be used as first choice.

The current UI culture seems a pretty good candidate as it is set by AspNetCore localization mechanism to load resource files...
And while I'm thinking about it, I am even wondering if it shouldn't be the first option. The names' localization only makes sense for non-latin characters. Maybe a Chinese guy would have set his preferred locale for his LinkedIn profile in Chinese, but an English ASP.Net application would rather get his name in English?

BTW, I agree that we should put all the names into claims, but I'm not sure how to do it. How put the names and the associated cultures in a claim? And maybe we are lacking another extension method to put several values for a single claim (which I could use for profile pictures, too)?

martincostello: Two immediate thoughts:

  1) Maybe there's a way we could add a `Func<>` property of some kind to the options class that would allow the configuring application to chose the culture and/or the multi-locale string?
  2) I'm not sure about how to add them all. Maybe there's a way the claims actions in the options can be configured to add multiple values? Failing that, they could be enumerated in the handler and added manually for each locale found in the response.

asiffermann:
I was answering to your comment regarding localization, I'd be glad to read your thoughts about it 😉

About the release, I understand it will be difficult for you to release it independently, but it is now a critical issue for this library, as LinkedIn apps are now being forced to use the new API (the v1 endpoint now returns HTTP error code 410). So it should be considered as a bug fix rather than a breaking change enhancement 😉

martincostello:
I understand, although in the short-term you could use the version with your fixes (once this is merged) from the MyGet feed from a pre-release version of the package.

Maybe @PinpointTownes has some opinions on how to manage this?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: I'm working on it, and push it as soon as possible!

martincostello: No rush 😃 - I'll probably not be able to re-review until later this evening now anyway.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -63,46 +63,39 @@ public LinkedInAuthenticationOptions()
...
         /// </summary>
         /// <param name="user">The payload returned by the user info endpoint.</param>
         /// <param name="propertyName">The name of the <c>MultiLocaleString</c> property.</param>
         /// <returns>The property value.</returns>
-        private static string GetMultiLocaleString(JObject user, string propertyName)
+        private string GetMultiLocaleString(JObject user, string propertyName)
         {
-            if (user[propertyName] == null)
+            const string localizedKey = "localized";
+            if (user[propertyName] == null || user[propertyName][localizedKey] == null)
```
martincostello: `user[propertyName]` is accessed twice. If you put the results into a local variable after the first access then you don't need to look it up a second time.

martincostello: Three times actually, as it's accessed again below.

martincostello: Similarly `user[propertyName][localizedKey]` is accessed once here and then again below.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -63,46 +63,39 @@ public LinkedInAuthenticationOptions()
...
-                }
-            }
-
-            var currentUiKey = Thread.CurrentThread.CurrentUICulture.ToString().Replace('-', '_');
-            var currentUiLocalizedValue = user[propertyName][localizedKey][currentUiKey];
-            if (currentUiLocalizedValue != null)
-            {
-                return currentUiLocalizedValue.Value<string>();
-            }
+            var preferredLocaleKey = preferredLocale == null ? null : $"{preferredLocale["language"]}_{preferredLocale["country"]}";
```
martincostello: Use `string` rather than `var`.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: It's done (with additional tests to cover it) 🙂

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
martincostello: Explicitly get the values as strings, rather than rely on the format operation to call `ToString()`, as again, this won't work in 3.0.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -63,46 +63,39 @@ public LinkedInAuthenticationOptions()
...
+            var values = user[propertyName][localizedKey].ToObject<Dictionary<string, string>>();
```
martincostello: `ToObject<>()` won't be available in 3.0.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -128,5 +121,31 @@ private static IEnumerable<string> GetPictureUrls(JObject user)
...
+        /// 1. Returns the <c>preferredLocale</c> value if it is set and has a value.
+        /// 2. Returns the value corresponding to the <see cref="Thread.CurrentUICulture"/> if it exists.
+        /// 3. Returns the first value.
+        /// </summary>
+        /// <param name="localizedValues">The localized values with culture keys.</param>
+        /// <param name="preferredLocale">The preferred locale, if provided by LinkedIn.</param>
+        /// <returns>The localized value.</returns>
+        private static string DefaultMultiLocaleStringResolver(IReadOnlyDictionary<string, string> localizedValues, string preferredLocale)
+        {
+            if (!string.IsNullOrEmpty(preferredLocale) && localizedValues.ContainsKey(preferredLocale))
```
martincostello: Use `TryGetValue()` rather than `ContainsKey()` + `[]`.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -128,5 +121,31 @@ private static IEnumerable<string> GetPictureUrls(JObject user)
...
+            {
+                return localizedValues[preferredLocale];
+            }
+
+            var currentUiKey = Thread.CurrentThread.CurrentUICulture.ToString().Replace('-', '_');
```
martincostello: Use `string` not `var`.

martincostello: `currentUIKey` (`UI` not `Ui`).

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -128,5 +121,31 @@ private static IEnumerable<string> GetPictureUrls(JObject user)
...
+            if (localizedValues.ContainsKey(currentUiKey))
```
martincostello: Use `TryGetValue()`.

test/AspNet.Security.OAuth.Providers.Tests/LinkedIn/LinkedInTests.cs
```diff
@@ -50,5 +55,33 @@ public async Task Can_Sign_In_Using_LinkedIn(string claimType, string claimValue
...
+
+        [Theory]
+        [InlineData(ClaimTypes.NameIdentifier, "1R2RtA")]
+        [InlineData(ClaimTypes.Name, "Frodon Sacquet")]
+        [InlineData(ClaimTypes.GivenName, "Frodon")]
+        [InlineData(ClaimTypes.Surname, "Sacquet")]
+        public async Task Can_Sign_In_Using_LinkedIn_Localized(string claimType, string claimValue)
+        {
+            // Arrange
+            additionnalConfiguration = options => options.MultiLocaleStringResolver = (values, preferredLocale) =>
```
martincostello: Could also add a second test method case here where a `try-finally` block is used to change the current thread culture and test that the default using the thread's culture also correctly returns the French version?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
asiffermann: Where is your .editorconfig to enforce contributors to do that? 😉

martincostello: _Should_ have gone in as part of #299 😄

asiffermann: I updated my fork with your changes, but I don't get it. In your editorconfig, you have:
```
csharp_style_var_for_built_in_types = true:silent 
csharp_style_var_when_type_is_apparent = true:silent 
csharp_style_var_elsewhere = true:silent 
```
And you asked me to change two `var` into `string`, so you just want to disallow var for built-in types?

martincostello: Sorry, yes, that's what it should be based on how the rest of the code appears to be styled already.

I just used the new VS 2019 16.1 feature to generate an EditorConfig based on existing styles, and I assumed it'd done the right thing because I ran a code cleanup on it after I'd added it.

I didn't fully audit the file though as there as so many settings!

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -27,41 +31,121 @@ public LinkedInAuthenticationOptions()
...
+            Scope.Add("r_liteprofile");
+            Scope.Add("r_emailaddress");
+
+            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, ProfileFields.Id);
+            ClaimActions.MapJsonKey(ClaimTypes.Email, LinkedInAuthenticationConstants.EmailAddressField);
+            ClaimActions.MapCustomJson(ClaimTypes.Name, user => GetFullName(user));
+            ClaimActions.MapCustomJson(ClaimTypes.GivenName, user => GetMultiLocaleString(user, ProfileFields.FirstName));
+            ClaimActions.MapCustomJson(ClaimTypes.Surname, user => GetMultiLocaleString(user, ProfileFields.LastName));
+            ClaimActions.MapCustomJson(Claims.PictureUrl, user => GetPictureUrls(user)?.LastOrDefault());
+            ClaimActions.MapCustomJson(Claims.PictureUrls, user => string.Join(",", GetPictureUrls(user)));
```
martincostello: As noted here https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/300#issuecomment-494873451 this needs to handle there being no pictures.

asiffermann: Yes, just noticed with my unit tests, update is on its way!

test/AspNet.Security.OAuth.Providers.Tests/LinkedIn/LinkedInTests.cs
asiffermann: In fact, it is not the test thread culture that should be changed, but the test server one.
So I added a virtual method on `OAuthTests<TOptions>` to register additional middlewares, and implemented it in `LinkedInTests` to add `UseRequestLocalization` on the application builder.

.editorconfig
```diff
@@ -85,7 +85,7 @@ dotnet_naming_symbols.constant_fields.required_modifiers          = const
 [*.cs] 
 
 # var preferences 
-csharp_style_var_for_built_in_types = true:silent 
+csharp_style_var_for_built_in_types = false:suggestion
```
martincostello, line 88: 👍

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -82,22 +86,25 @@ public LinkedInAuthenticationOptions()
         /// <returns>The property value.</returns>
         private string GetMultiLocaleString(JObject user, string propertyName)
         {
-            const string localizedKey = "localized";
-            if (user[propertyName] == null || user[propertyName][localizedKey] == null)
+            var property = user[propertyName];
+            var propertyLocalized = property["localized"];
```
martincostello, line 95: You're getting this value before you do the `null` check, so it would throw a `NullReferenceException` if it were.

asiffermann: 🤦‍♂ me...

asiffermann: Hopefully you would squash merge 😉

jandieg:
> I understand, although in the short-term you could use the version with your fixes (once this is merged) from the MyGet feed from a pre-release version of the package.
> 
> Maybe @PinpointTownes has some opinions on how to manage this?

Yeas please, any simple instructions in how to get this fix? (in the case it won't be released soon)

martincostello:
The _quickest_ way to consume this is to download the package from the AppVeyor CI Build and check into source control and configure a local from-disk NuGet feed and reference the prerelease version from there.

For instance I’m using this approach for 3.0 support for the Amazon provider here: https://github.com/martincostello/alexa-london-travel-site/pull/255

If this is merged but retained for a future 3.0 release in September (i.e. we don’t push to NuGet.org as SemVer breaking and want to reserve the 3.0 version number for simplicity), an alternative is that you could consume it from the MyGet feed in the README.

martincostello:
As an alternative, which would allow this to be a minor release, you could add back all the deleted members and instead mark them as `[Obsolete(“blah blah”)]` and then I can remove them as part of the 3.0 release.

martincostello:
In fact, I think that's the best approach.

Restore the deleted constants, and mark them as something like `[Obsolete("This constant is not used in the LinkedIn v2.0 API.")]`.

This could then be released as a 2.1 version and remain SemVer compatible. Users who've actually consumed the constants whose values have changed in their own code would just need to recompile against a later version to pick up the value changes.

martincostello (approved):

martincostello:
Thanks for your work on this @asiffermann 🎈 

I'll merge it later on today.

asiffermann:
Thanks to you for your review! 😉

Issue #321, opened 2019-06-13
KoreBuild is obsolete


martincostello:
As noted here https://github.com/aspnet/BuildTools/pull/977, KoreBuild is now obsolete.

This breaks the #280 branch from 3.0 preview 6 and onwards as there is no version available to download that supports these SDK versions.

The question is - what's the replacement?

ASP.NET Core uses "Arcade" now, but I imagine that's not designed for our needs and (I imagine) not simple to integrate.

To unblock #280 for now, I've just pulled in the simple scripts I use for my own projects until there's an agreed replacement.

/cc @PinpointTownes @poke

kevinchalet:
OpenIddict was updated to use Arcade: as you said, it was designed for Microsoft's needs, so it's super opinionated. Disabling certain features (like packages signing) is a painful process, and you have to dig into the massive documentation to learn how to do that.

That said, I'm convinced it's the right way to go: while it's painful to configure, it works like a charm once it's set up, and it's something we won't have to maintain in the long term.

It's also a cool way to find things that could be simplified or improved. For instance, it will encourage you to remove the `xunit`/`Microsoft.NET.Test.Sdk` references from the tests projects, as they are now part of the SDK (which is something VS doesn't tell you if you don't use Arcade).

martincostello:
I've got no problem with adopting Arcade - I just found the docs a bit impenetrable at first glance, so figured it was worth discussion prior to trying to adopt it and then finding it's not what we wanted to do.

Do you have a link the to PR for converting OpenIddict? I can use that as a basis to migrate the 3.0 branch over before preview 9 lands.

kevinchalet:
> I've got no problem with adopting Arcade - I just found the docs a bit impenetrable at first glance, so figured it was worth discussion prior to trying to adopt it and then finding it's not what we wanted to do.

Makes total sense 😄 

Here are the relevant parts:
  - https://github.com/openiddict/openiddict-core/tree/dev/eng
  - https://github.com/openiddict/openiddict-core/blob/dev/Directory.Build.props
  - https://github.com/openiddict/openiddict-core/blob/dev/Directory.Build.targets
  - https://github.com/openiddict/openiddict-core/blob/dev/global.json
  - https://github.com/openiddict/openiddict-core/blob/dev/Build.cmd
  - https://github.com/openiddict/openiddict-core/blob/dev/build.sh

[...]

PR #335, opened 2019-08-17
Use Arcade for build

martincostello:
Switch to Arcade to build the project.

Addresses #321.


martincostello:
_Almost_ done.

The only thing missing is that the test results [aren't reported](https://ci.appveyor.com/project/aspnet-contrib/aspnet-security-oauth-providers/builds/26768704/tests) in AppVeyor.

Any ideas @PinpointTownes? I can [see them](https://ci.appveyor.com/project/openiddict/openiddict-core/builds/26757595/tests) in openiddict-core. 🤔

kevinchalet:
> Any ideas @PinpointTownes?

Honestly no. It's weird: they are correctly executed but not reported 😕 

(note: you shouldn't merge it until we solve the versioning issue, otherwise it will be a PITA for nightly builds 😄)

martincostello:
Yeah, I'm just having a dig around in the code and docs now to see if I can work it out.

We _might_ have to add some custom MSBuild overrides somewhere to take the AppVeyor properties and turn them into AzDO ones for Arcade to pick up.

martincostello:
So from looking at this:

https://github.com/dotnet/arcade/blob/331447d81b1968bed7161c4a6a8d70e4990327e1/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L18-L25

and the [AppVeyor](https://www.appveyor.com/docs/environment-variables/) docs, I think we can "fix" it by adding this to `Directory.Build.props`:

```xml
<OfficialBuild Condition=" '$(APPVEYOR)' == 'True' AND '$(APPVEYOR_PULL_REQUEST_NUMBER)' == '' ">true</OfficialBuild>
```

kevinchalet:
We can, but it has a downside: in this case, it generates an official build number based on the current date with a fixed ".1" revision, so if we merge more than 1 commit a day, it will end up overwriting existing packages.

We could provide an `OfficialBuildId`, but the format doesn't allow revisions that have more than 2 digits. Since AppVeyor build numbers are not reset every day, it's super limiting 😢

martincostello:
Hmm. I'll have a further think about the revision part...

martincostello:
OK, I'm not proud of this, but it works*...

*as long as you don't publish to MyGet more than once per quarter of the hour.

If we have to rollover every 24 hours, and we can't have more than 2 digits, what about an auto-incrementing build number that's based on which quarter-hour of the day it is right now?

There's 24 hours in a day, so there's 96 quarter hours a day.

A build between `00:00:00` and `00:14:59` would be revision `1`.
A build between `23:45:00` and `23:59:59` would be revision `97`.

If I run this command on my laptop at 20:29, I get this:

```cmd
.\build.cmd -pack /p:APPVEYOR=True

<...lots of build output...>

Successfully created package 'AspNet.Security.OAuth.MailRu.3.0.0-dev.19417.82.symbols.nupkg'.
```

Code to add to `Directory.Build.props` below:

```xml
<!--
  Arcade only allows the revision to contain up to two characters, and AppVeyor does not roll-over
  build numbers every day like Azure DevOps does. To balance these two requirements, set the official
  build ID to be the same format as the built-in default from Arcade, except with the revision number
  being the number of the quarter hour of the current time of day (24 * 4 = 96, which is less than 100).
  So a build between 00:00 and 00:14 would have a revision of 1, and a build between 23:45 and 23:59:59
  would have a revision of 97.
-->
<PropertyGroup Condition=" '$(APPVEYOR)' == 'True' AND '$(APPVEYOR_PULL_REQUEST_NUMBER)' == '' ">
  <_Hours>$([MSBuild]::Multiply($([System.DateTime]::Now.ToString(HH)), 4))</_Hours>
  <_QuarterHours>$([MSBuild]::Divide($([System.DateTime]::Now.ToString(mm)), 15))</_QuarterHours>
  <_QuarterHours>$([System.Math]::Floor($(_QuarterHours)))</_QuarterHours>
  <_AppVeyorBuildRevision>$([MSBuild]::Add($(_Hours), $(_QuarterHours)))</_AppVeyorBuildRevision>
  <_AppVeyorBuildRevision>$([MSBuild]::Add($(_AppVeyorBuildRevision), 1))</_AppVeyorBuildRevision>
  <OfficialBuild>true</OfficialBuild>
  <OfficialBuildId Condition=" '$(OfficialBuild)' == 'true' ">$([System.DateTime]::Now.ToString(yyyyMMdd)).$(_AppVeyorBuildRevision)</OfficialBuildId>
</PropertyGroup>
```

kevinchalet:
Clever (and certainly better than nothing 😅)

That said, I'll ping the Arcade folks to see what they think about that, just to make sure we're not missing something obvious.

martincostello:
Sure, this is definitely a hack!

martincostello:
Thoughts on merging this as-is and opening a new PR if we find a better solution via https://github.com/dotnet/arcade/issues/3706?

martincostello:
I'm going to merge this to the 3.0.0 branch with the current "1/4 hours in a day" versioning approach so I can make a start on the updates for preview 9 with minimal fuss.

PR #336, opened 2019-09-03
incorrect LoginPath

mgroves:
In this startup file, it previously said "login", but in AuthenticationController the route is defined as "signin". When I was following along, I got an error message because of this discrepancy.


martincostello (approved):

martincostello:
Thanks for the fix @mgroves!

PR #343, opened 2019-09-20
Add Sign in with Apple provider

martincostello:
This PR merges `dev` into `dev-3.0.0` to add the _Sign in with Apple_ provider for ASP.NET Core 3.0.0 from #318.

It also updates the .NET Core test SDK to 16.3.0.


samples/Mvc.Client/Startup.cs
```diff
@@ -58,7 +82,18 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
-            app.UseStaticFiles();
+            if (HostingEnvironment.IsDevelopment())
+            {
+                IdentityModelEventSource.ShowPII = true;
+            }
+
+            // Required to serve files with no extension in the .well-known folder
```
Tratcher, line 90: ?

martincostello: I've had issues in the past where files in the `.well-known` folder aren't served, which are needed to do the domain ownership with Apple for things like Sign in with Apple and Apple Pay JS: https://github.com/justeat/ApplePayJSSample/commit/e17ba6f8d9b96361b5c9c34fe930c8477b4fe999

If this has since become redundant or there's a different way to get it working I can update it.

Tratcher: Might make more sense as an app endpoint (or middleware) than a static file?

martincostello: Yeah, that might be better way to do it for the 3.0 version. The file names could be configurable and then people can add in extra ones in the future that this doesn't need or they might want extra.

This was just a quick way to get things up and running for the demo/dev site I stood up in App Service using publish from Visual Studio: https://signinwithapple.azurewebsites.net/ using

PR #352, opened 2019-10-13
3.1.0 preparation

martincostello:
* Bump version to `3.1.0`
  * Reset prelease label to `preview`
  * Update to the latest release of `JustEat.HttpClientInterception`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 3.0.0.{build}
+version: 3.1.0.{build}
 
 build_script:
   - eng\common\CIBuild.cmd -configuration Release -prepareMachine
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -14,7 +14,7 @@
     <JetBrainsVersion>2019.1.3</JetBrainsVersion>
     <JustEatHttpClientInterceptionVersion>3.0.0</JustEatHttpClientInterceptionVersion>
     <MartinCostelloLoggingXUnitVersion>0.1.0</MartinCostelloLoggingXUnitVersion>
-    <MicrosoftNETTestSdkVersion>16.3.0</MicrosoftNETTestSdkVersion>
+    <MicrosoftNETTestSdkVersion>16.4.0</MicrosoftNETTestSdkVersion>
     <ShouldlyVersion>3.0.2</ShouldlyVersion>
     <SystemIdentityModelTokensJwtVersion>5.3.0</SystemIdentityModelTokensJwtVersion>
     <TwitterProviderVersion>3.0.0</TwitterProviderVersion>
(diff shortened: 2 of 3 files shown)


kevinchalet:
> Will merge once 3.0.0 ships.

What do we want to do here? Shipping 3.0.0 with an incorrect strong name - i.e backed with a MSFT signing key we don't even own - is not an option. So we can either skip the 3.0.0 release and only push 3.0.1 to NuGet.org or call that release 3.0.0 (my favorite option).

kevinchalet:
Oh, scratch that, I had not realized it was sent against the `dev` branch. Should we use `3.1.0` instead of `3.0.1` for the dev branch?

Also, it would be preferable to simply merge `rel/3.0.0` into `dev` instead of cherry-picking.

martincostello:
I was in the middle of typing that I was confused before the second comment came in 😄 

Can change to `3.1.0` if you want to - I assumed "next" would likely to be a patch, and that a minor would come later and we'd bump the version again if we needed to.

I'm generally in the habit of just bumping patch by one immediately after doing a release just to make sure it's higher before putting any actual thought into what the next release will actually be.

martincostello:
For the avoidance of doubt, [this build](https://ci.appveyor.com/project/aspnet-contrib/aspnet-security-oauth-providers/builds/28061548/artifacts) with your fix in for the strong name in should be the `3.0.0` release I reckon.

kevinchalet:
> I'm generally in the habit of just bumping patch by one immediately after doing a release just to make sure it's higher before putting any actual thought into what the next release will actually be.

That's certainly not an uncommon or a bad approach 😄 

So far, I've used pretty much the same approach as the ASP.NET team for non-major versions: once a new version is released, a git tag is created and the minor version of the development branch is incremented by 1. Once we need to patch this release, a `rel/` branch is simply derived from this tag so we can push new commits or cherry-pick existing fixes from the main branch.

This has one pro: we can add quite massive changes in the `dev` branch (e.g new providers), without being annoyed the day we need to release a patch without all these changes 😄 

But if you prefer your option, it's totally fine too.

martincostello:
I did start as a merge like I did with OpenId, but I figured it was more that it had diverged than needed merging back as there was a duplicated commit as I’d already just done one fix straight to dev after doing it to rel/3.0.0.

I can redo this as a proper merge though if needed.

martincostello:
I’ve got the tag for 3.0.0 prepped in the releases for once you _push the button_ on the release.

Happy to re-do this PR as 3.1 and a merge instead and push it up later on.

martincostello:
Changed version to 3.1.0 and reworked as a merge.

martincostello:
I might have messed up the merge on this, as it's only giving me the option to squash...

PR #362, opened 2019-12-12
Add alipay provider

vicenteyu:
Signed-off-by: vicente.yu 
add alipay provider


vicenteyu:
Usage:

.AddAlipay(options =>
{
    options.ClientId = Configuration["Authentication:Alipay:AppId"];
    options.ClientSecret = Configuration["Authentication:Alipay:MerchantPrivateKey"];
})

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static class AlipayAuthenticationConstants
+    {
+        public static class Claims
+        {
+            public const string UserId = "urn:alipay:user_id";
+            public const string Avatar = "urn:alipay:avatar";
+            public const string Province = "urn:alipay:province";
+            public const string City = "urn:alipay:city";
+            public const string NickName = "urn:alipay:nick_name";
```
martincostello: In English, this would be `Nickname` as it is one word, not two.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationDefaults.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="RemoteAuthenticationOptions.CallbackPath"/>.
+        /// </summary>
+        public const string CallbackPath = "/signin-alipay";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+        /// </summary>
+        public const string AuthorizationEndpoint = "https://openauth.alipay.com/oauth2/publicAppAuthorize.htm";
```
martincostello, line 40: Just checking, as it's unusual, this is definitely `.htm`?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationDefaults.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+        /// </summary>
+        public const string TokenEndpoint = "https://openapi.alipay.com/gateway.do";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+        /// </summary>
+        public const string UserInformationEndpoint = "https://openapi.alipay.com/gateway.do";
```
martincostello, line 50: Is it correct that the same URL is used for tokens and users? Are the responses identical, or is there some other parameter used to differentiate between getting a token and getting a users' details?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
+﻿using JetBrains.Annotations;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.OAuth;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using System;
```
martincostello, line 7: Sort `using`s with `System*` directives first please.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace AspNet.Security.OAuth.Alipay
```
martincostello: This file, and all the other `.cs` files, need the Apache 2.0 license header adding like the rest of the files in the repo:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/1a32506eedbacd83fed18b985aa11d58928192d7/src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationDefaults.cs#L1-L5

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+using System.Security.Claims;
+using System.Text;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace AspNet.Security.OAuth.Alipay
+{
+    /// <summary>
+    /// 支付宝 OAUTH2.0 网站登录
```
martincostello: Either remove this or use comments like the below. This sort of information is not useful for Intellisense comments in an IDE.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/1a32506eedbacd83fed18b985aa11d58928192d7/src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationHandler.cs#L22-L24

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+        protected override string BuildChallengeUrl(AuthenticationProperties properties, string redirectUri)
+        {
+            string stateValue = Options.StateDataFormat.Protect(properties);
+
+            var address = QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, new Dictionary<string, string>
+            {
+                ["app_id"] = Options.ClientId,
+                ["scope"] = FormatScope(),
+                ["redirect_uri"] = redirectUri,
+                [State] = stateValue
```
martincostello: This constant isn't used anywhere else. Just remove it and use a literal like with the other 3 keys.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            });
+            return address;
+        }
```
martincostello, line 55: Nit: new line between methods.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+        protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
+        {
+            var query = Request.Query.ToDictionary(c => c.Key, c => c.Value, StringComparer.OrdinalIgnoreCase);
+            if (query.TryGetValue("auth_code", out var auth_code))
```
martincostello: Rename `auth_code` to `authCode`.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                query["code"] = auth_code;
+                Request.QueryString = QueryString.Create(query);
+            }
+            return await base.HandleRemoteAuthenticateAsync();
+        }
```
martincostello, line 68: Nit: new line between methods.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+        protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
+        {
+            var sortedparams = new SortedDictionary<string, string>()
```
martincostello: `sortedparams` -> `sortedParameters` or `sortedParams`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+                ["method"] = "alipay.system.oauth.token",
+                ["format"] = "JSON",
+                ["charset"] = "UTF-8",
+                ["sign_type"] = "RSA2",
+                ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss"),
+                ["version"] = "1.0",
+                ["code"] = context.Code,
+                ["grant_type"] = "authorization_code",
+            };
+            var signparames = RSA2Sign(sortedparams);
```
martincostello: `signedParams`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            if (Options.NotValidateSign)
```
martincostello: Rename `NotValidateSign` to `SkipSignatureValidation` (I'm assuming that's what this is doing?).

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                signparames.Remove("timestamp");
```
martincostello: Would be more efficient to only add them if needed, rather than add them and then throw them away again if not.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+                Logger.LogError("An error occurred while retrieving an access token: the remote server " +
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                /* Status: */ response.StatusCode,
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+            }
+            using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
+            using var alipay_system_oauth_token_response = JsonDocument.Parse(payload.RootElement.GetProperty("alipay_system_oauth_token_response").ToString());
```
martincostello: Rename to `tokenResponse`.

martincostello: Change to `payload.RootElement.GetProperty("alipay_system_oauth_token_response").GetString()`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            return OAuthTokenResponse.Success(alipay_system_oauth_token_response);
+        }
+        protected override async Task<AuthenticationTicket> CreateTicketAsync(
```
martincostello, line 107: Nit: new line between methods.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            [NotNull] ClaimsIdentity identity,
+            [NotNull] AuthenticationProperties properties,
+            [NotNull] OAuthTokenResponse tokens)
+        {
+            var sortedparams = new SortedDictionary<string, string>()
```
martincostello: Same comments as above apply to this method too.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                /* Status: */ response.StatusCode,
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                throw new HttpRequestException("An error occurred while retrieving user information.");
+            }
+            using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
+
+            using var alipay_user_info_share_response = JsonDocument.Parse(payload.RootElement.GetProperty("alipay_user_info_share_response").ToString());
```
martincostello: Rename to `shareResponse`.

martincostello: Use `GetString()` not `ToString()`

martincostello: I think all of the code above this line is exactly the same as in the previous method. The commonalities can be factored into a private method to move the duplication.

martincostello: In fact, could just be `payload.RootElement.GetString("alipay_user_info_share_response")`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: In fact could just be `payload.RootElement.GetString("alipay_system_oauth_token_response")`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            string status = alipay_user_info_share_response.RootElement.GetString("code");
+            if (status != "10000")
```
martincostello: What does this status code mean?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            await Options.Events.CreatingTicket(context);
+            return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
+        }
+
+        protected override string FormatScope() => string.Join(",", Options.Scope);
+        private IDictionary<string, string> RSA2Sign(SortedDictionary<string, string> source)
+        {
+            var textparams = source.Where(p => !string.IsNullOrEmpty(p.Value)).Select(p => p.Key + "=" + p.Value);
+            if (!textparams.Any())
+                throw new HttpRequestException("sign fail");
```
martincostello: This isn't a useful exception message for someone consuming the library to try and debug/diagnose.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+            var rsa2 = new RSAHelper(RSAType.RSA2, Encoding.UTF8, Options.ClientSecret, Options.AlipayPublicKey);
+            source.Add("sign", rsa2.Sign(string.Join("&", textparams)));
+            return source;
```
martincostello: This is confusing - at the call site it appears to create a new object, but it's actually mutating the state and returning the original. It would be clearer for this method to be called something like `SignParameters()` and return the value. Then the caller can append the sign parameter there.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+using static AspNet.Security.OAuth.Alipay.AlipayAuthenticationConstants;
+
+namespace AspNet.Security.OAuth.Alipay
+{
+    /// <summary>
+    /// Defines a set of options used by <see cref="AlipayAuthenticationHandler"/>.
+    /// </summary>
+    public class AlipayAuthenticationOptions : OAuthOptions
+    {
+        public string AlipayPublicKey { get; set; }
```
martincostello: Nit: new line below.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether [not validate sign].
```
martincostello: I think it would be clearer if this was named `ValidateSignature` and had a default value of `true`. The `Not` in the name being used to invert the condition isn't very idiomatic C# in my opinion.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        /// <value><c>true</c> if [not validate sign]; otherwise, <c>false</c>.</value>
+        public bool NotValidateSign { get; set; } = false;
+        public AlipayAuthenticationOptions()
```
martincostello, line 20: Nit: new line.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+            ClaimsIssuer = AlipayAuthenticationDefaults.Issuer;
+            CallbackPath = new PathString(AlipayAuthenticationDefaults.CallbackPath);
+
+            AuthorizationEndpoint = AlipayAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = AlipayAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = AlipayAuthenticationDefaults.UserInformationEndpoint;
+
+            Scope.Add("auth_user");
+
+            ClaimActions.MapJsonKey(Claims.UserId, "user_id");
```
martincostello: Sort these assignments, it makes the code easier to maintain and read.

src/AspNet.Security.OAuth.Alipay/AspNet.Security.OAuth.Alipay.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>ASP.NET Core security middleware enabling Alipay authentication.</Description>
+    <Authors>vicente</Authors>
+    <PackageTags>aspnetcore;authentication;oauth;alipay;security</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
```
martincostello, line 13: This item group is redundant, delete it.

src/AspNet.Security.OAuth.Alipay/AspNet.Security.OAuth.Alipay.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <None Remove="bin\**" />
+    <None Remove="obj\**" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
```
martincostello: This can be put in the above item group so the two references are together (like in all the other `.csproj` files in this repo).

src/AspNet.Security.OAuth.Alipay/RSAHelper.cs
```diff
@@ -0,0 +1,317 @@
+﻿using System;
```
martincostello: The need for this entire class here concerns me. We the maintainers don't really want to look after all this cryptographic code that _may_ contain security vulnerabilities on an ongoing basis.

It's also non-trivial, and documented in Mandarin/Cantonese, which none of the maintainers speak.

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
```diff
@@ -0,0 +1,58 @@
...
...3FQ7oFNdBbu4eSMWVagbyECJuD6q8gOSjyjCxDHqY+Df4Fb/h3BOdQZK7mq6UEC+65YWsAgC/+sRshrvRQjFoKkYsjaQ0THWw2WZQrc+vS6h39MrQRCaYtRHp+JINMgKah1mE334gNv0yJwzV210GDzHO1QytN7/UlvgOGwOIw+uohEhsbwdb++MRamtpLcjD96xxSixJlEAubUhfQxh1bBTAuCPU1I5H/J+rrZAC2YUQVb2Goy8LUqjn70ximuoXV+xZCJ1Hgr9e6vzvnuLy2wUgOkrE/HS57wqjkkBSVO5bQ1j5H9NLouOUZ5WNKOfECgYEAyNh4ATFsWpsx9lBFoxgFWrwhM5R8OLODbqs8qAJLfC6DNVgaCxeQGmg3RZ5jxQBLdpSCWSCIT/eKPvu0CiGLdny7+t3PotwLuBfXcy+dAdI4ubg3A2vGRDUk0U9ZkuUuaM7DvXQOmLva8T98blX4vOZwEL1wmchOxgk8yMxYHCcCgYEAxFpT0UPqc8LImsReb+o5LAeBZsAJpG2s/fsgfS/gsCwduf4OJ5N5TlUVHbOSzISCWDPmVz9sWEFj9s6co0DYLGmaid6+s5RI9DwRhVaL8Sl0PYaam7frNuiTedyRfMLF7n8PhHqensJQjR/xP+CZOs80KRhdvjwXv0pFS7zWNoUCgYACIUGMibCjeVfbS9ihNSUBZFNZz65Nj6HKL6iyA2w3gerESw3jpjlR/l7vrxFRyoICXOrQ9SZc6rwdlN/A58Ap3oLD00xbZsf9CMuxHgUlOsx9M7XppF/y4zljutqxUxrd46txu+RXvE5DFrBEH0dHAY6Yrtmd1+D1+q0ZWAlrKwKBgHCpLzx8DnLbSUTb9R+bsbAcole1ShMJRt/3jj2mEfKjbW8BYVe92zVhxhrjpRAp26wGI1zeLCk7Y8MB64gUNbTN5vjCUIMzSuSv7pGmBeealHKU3/MHBTPdIHkkYGnIS887IkkuHgMTlSpUMJUpJmJC7zfBHlB/pFSCpd39/J8NAoGBAJn+/16vao78AuXk62OlmsNf9yGMDQpAH+vswjO+HwwG5845D3T4fNH1qfo5jc592/8LpEtuTDkMwHRz34MPTIwVJpkQrHhoW74YlQMU5rTM23gETVXtX91rUZS71rd9EdBB34vexLqHNyD+1/cH5MRdgW9ZpM34IsaGuvXKaS/e";
```
martincostello: Is this a real client secret? Where did it come from?

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                // Act
+                var claims = await AuthenticateUserAsync(server);
+
+                // Assert
+                AssertClaim(claims, claimType, claimValue);
+            }
+        }
+    }
+
```
martincostello, line 56: Nit: blank line here not needed.

martincostello (changes requested):
The `RSAHelper` class in this PR doesn't look like something we'd want to have to maintain. The functionality needed should come in via a trusted and maintained open source project via a NuGet package reference.

There's also some duplication of code, and various minor style changes to fit in with the other providers for consistency.

martincostello:
Thank you for contributing this PR @vicenteyu - however, this is going to need various changes before it's something we'd consider merging and maintaining.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationDefaults.cs
vicenteyu: Yes. Confirmed.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationDefaults.cs
vicenteyu: Yes.
Get token : https://openapi.alipay.com/gateway.do?method=alipay.system.oauth.token... 
Get user info : https://openapi.alipay.com/gateway.do?method=alipay.user.info.share...

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: If status code == 10000 then success else faild.

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
vicenteyu: It's test secret, not need now. :)

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -1,18 +1,63 @@
...
             public const string City = "urn:alipay:city";
-            public const string NickName = "urn:alipay:nick_name";
+
+            /// <summary>
+            /// The nickname
+            /// </summary>
+            public const string Nickname = "urn:alipay:nick_name";
+
+            /// <summary>
+            /// The is student certified. T: True; F: False;
```
martincostello: Change to _"Whether the student is certified."_

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -1,18 +1,63 @@
...
+            /// </summary>
             public const string UserType = "urn:alipay:user_type";
+
+            /// <summary>
+            /// The user status. Q: Quick; T: Typical; B: Block; W: Waiting for Activation;
+            /// </summary>
             public const string UserStatus = "urn:alipay:user_status";
+
+            /// <summary>
+            /// The is certified. T: True; F: False;
```
martincostello: Change to _"Whether the user is certified."_

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -1,18 +1,63 @@
...
+            /// </summary>
             public const string IsCertified = "urn:alipay:is_certified";
+
+            /// <summary>
+            /// The gender. F: Femal; M: Male;
```
martincostello: Change to _"The user's gender."_

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -40,46 +46,52 @@ protected override string BuildChallengeUrl(AuthenticationProperties properties,
             var address = QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, new Dictionary<string, string>
             {
                 ["app_id"] = Options.ClientId,
-                ["scope"] = FormatScope(),
+                ["scope"] = string.Join(",", Options.Scope),
```
martincostello: What was the reason for removing the `FormatScope()` override?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
             }
-            using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
 
-            using var alipay_user_info_share_response = JsonDocument.Parse(payload.RootElement.GetProperty("alipay_user_info_share_response").ToString());
+            using var payload = await getPayloadAsync(response, "alipay_user_info_share_response");
+            string statusCode = payload.RootElement.GetString("code");
 
-            string status = alipay_user_info_share_response.RootElement.GetString("code");
-            if (status != "10000")
+            if (statusCode != "10000")/*  */
```
martincostello: Forgot to add text to the comment?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
 
-        protected override string FormatScope() => string.Join(",", Options.Scope);
-        private IDictionary<string, string> RSA2Sign(SortedDictionary<string, string> source)
+        /// <summary>
+        /// get payload as an asynchronous operation.
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="nodeName">Name of the node.</param>
+        /// <returns>Task&lt;JsonDocument&gt;.</returns>
+        private async Task<JsonDocument> getPayloadAsync(HttpResponseMessage message, string nodeName)
```
martincostello: Rename to `ParseJsonAsync(HttpResponseMessage message, string elementName)`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
             context.RunClaimActions();
 
             await Options.Events.CreatingTicket(context);
             return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
         }
 
-        protected override string FormatScope() => string.Join(",", Options.Scope);
-        private IDictionary<string, string> RSA2Sign(SortedDictionary<string, string> source)
+        /// <summary>
+        /// get payload as an asynchronous operation.
```
martincostello: _"Gets an HTTP payload as as an asynchronous operation."_

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
         {
-            var textparams = source.Where(p => !string.IsNullOrEmpty(p.Value)).Select(p => p.Key + "=" + p.Value);
-            if (!textparams.Any())
-                throw new HttpRequestException("sign fail");
-            var rsa2 = new RSAHelper(RSAType.RSA2, Encoding.UTF8, Options.ClientSecret, Options.AlipayPublicKey);
-            source.Add("sign", rsa2.Sign(string.Join("&", textparams)));
-            return source;
+            var messageContent = await message.Content.ReadAsStringAsync();
+
+            //Logger.LogInformation(messageContent);
```
martincostello: Remove.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
+
+            var rootJson = JsonDocument.Parse(messageContent);
```
martincostello: `document`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
+
+        /// <summary>
+        /// Gets the RSA2 signature.
+        /// </summary>
+        /// <param name="source">The source.</param>
+        /// <param name="privateKey">The private key.</param>
+        /// <returns>System.String.</returns>
+        private string getRSA2Signature(SortedDictionary<string, string> source, [NotNull] string privateKey)
+        {
+            var textParams = source
```
martincostello: I still think it would be better to build this up using `QueryHelpers.AddQueryString()` to ensure it's correctly encoded.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
+                .Where(p => !string.Equals(p.Key, "sign", StringComparison.OrdinalIgnoreCase)
+                    && !string.IsNullOrEmpty(p.Value))
+                .Select(p => p.Key + "=" + p.Value);
+
+            var privateKeyRsaProvider = CreateRsaProviderFromPrivateKey(privateKey);
```
martincostello: Instances of `RSA` implement `IDisposable`, so this needs to be disposed of:

```csharp
using RSA rsa = CreateRsaProviderFromPrivateKey(privateKey);
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -130,40 +144,158 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
+        /// <param name="privateKey">The private key.</param>
+        /// <returns>RSA.</returns>
+        /// <exception cref="Exception">
+        /// Unexpected value read binr.ReadUInt16()
+        /// or
+        /// Unexpected version
+        /// or
+        /// Unexpected value read binr.ReadByte()
+        /// </exception>
+        private RSA CreateRsaProviderFromPrivateKey(string privateKey)
```
martincostello: Thanks - this is much clearer to its intent compared to the previous code.

As pointed out to me by @vcsjones, there's APIs in .NET Core 3.0 that allow you to do this in a much simpler way:

```csharp
private RSA CreateAlgorithm()
{
    byte[] privateKeyBytes = Convert.FromBase64String(Options.ClientSecret);
    var algorithm = RSA.Create();

    try
    {
        algorithm.ImportRSAPrivateKey(privateKeyBytes, out int _);
        return algorithm;
    }
    catch (Exception)
    {
        algorithm.Dispose();
        throw;
    }
}
```

You can see a similar usage in our Sign in with Apple provider:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/1a32506eedbacd83fed18b985aa11d58928192d7/src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs#L94-L108

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -11,12 +16,12 @@ namespace AspNet.Security.OAuth.Alipay
     /// </summary>
     public class AlipayAuthenticationOptions : OAuthOptions
     {
-        public string AlipayPublicKey { get; set; }
         /// <summary>
-        /// Gets or sets a value indicating whether [not validate sign].
+        /// Gets or sets a value indicating whether [validate signature].
```
martincostello: _"Gets or sets a value indicating whether to validate the signature."_

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -11,12 +16,12 @@ namespace AspNet.Security.OAuth.Alipay
...
         /// </summary>
-        /// <value><c>true</c> if [not validate sign]; otherwise, <c>false</c>.</value>
-        public bool NotValidateSign { get; set; } = false;
+        /// <value><c>true</c> if [validate signature]; otherwise, <c>false</c>.</value>
```
martincostello: `<c>true</c> to validate the signature; otherwise, <c>false</c>.

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
```diff
@@ -25,8 +25,7 @@ protected internal override void RegisterAuthentication(AuthenticationBuilder bu
...
-                options.NotValidateSign = true;
+                options.ValidateSignature = false;
```
martincostello: As long as it isn't a real secret that works with production Alipay, restore a client secret and re-enable the validation, otherwise those default code paths aren't being tested.

There's a test PKCS#8 certificate that I generated myself used in the Apple provider's test for this sort of purpose: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev/test/AspNet.Security.OAuth.Providers.Tests/Apple

martincostello (commented):
This is coming along nicely, we're getting there.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: Only 1 time reference, keep override?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: Yes. :)

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: As the provider base class already provides functionality to do this, it's better to override it.

vicenteyu: OK. Thanks. :)

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: Sorry, i don't know how to modify?

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
vicenteyu: Set a fixed timestamp and restore a client secret, all code paths are tested.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: url encode and signature has not related.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -170,19 +164,17 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
         }
 
         /// <summary>
-        /// get payload as an asynchronous operation.
+        /// Gets an HTTP payload as as an asynchronous operation.
```
martincostello: _"Gets an HTTP payload as an asynchronous operation."_ (_as_ was duplicated)

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -16,11 +17,7 @@ namespace AspNet.Security.OAuth.Alipay
     /// </summary>
     public class AlipayAuthenticationOptions : OAuthOptions
     {
-        /// <summary>
-        /// Gets or sets a value indicating whether [validate signature].
-        /// </summary>
-        /// <value><c>true</c> if [validate signature]; otherwise, <c>false</c>.</value>
-        public bool ValidateSignature { get; set; } = true;
+        public DateTime? Timestamp { get; set; }
```
martincostello: Revert this part of the change. The ~`IClock`~ `ISystemClock` interface is provided for testability - use that in the tests to "freeze" the clock to the time `new DateTime(2019, 12, 14, 22, 22, 22)` instead.

In fact, in doing that you can also just remove the option to validate the signature entirely and always do it. The tests can freeze the clock so they pass, and users then aren't given an option to make it easy to make the provider less secure.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: Sorry, noticed that was my mistake in the original suggestion, but it was still wrong 😄

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: This is what I was thinking:

```csharp
string dataToSign = string.Empty;

foreach (var pair in source)
{
    if (string.Equals(pair.Key, "sign", StringComparison.Ordinal))
    {
        continue;
    }

    dataToSign = QueryHelpers.AddQueryString(dataToSign, pair.Key, pair.Value);
}


byte[] dataBytes = Encoding.UTF8.GetBytes(dataToSign[1..]);
```

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
martincostello: As per comment, use ~`IClock`~ `ISystemClock` to fix the timestamp.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
martincostello: You should be able to do this by adding the following:

```csharp
protected internal override void RegisterAuthentication(AuthenticationBuilder builder)
{
    builder.Services.AddSingleton<ISystemClock, FixedClock>();
    builder.AddAlipay(options =>
    {
        ConfigureDefaults(builder, options);
        options.ClientSecret = "{the key}";
    });
}

private sealed class FixedClock : ISystemClock
{
    public DateTimeOffset UtcNow => new DateTime(2019, 12, 14, 22, 22, 22, DateTimeKind.Utc);
}
```

martincostello: Actually, I think it'll need to be registered before `AddAuthentication()` based on [this code](https://github.com/aspnet/AspNetCore/blob/62351067ff4c1401556725b401478e648b66acdc/src/Security/Authentication/Core/src/AuthenticationServiceCollectionExtensions.cs#L26) that wires up the default `ISystemClock`.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/a865eedad1ad0717fca85c42086a2cd3a34c7d8f/test/AspNet.Security.OAuth.Providers.Tests/Infrastructure/ApplicationFactory.cs#L77

vicenteyu: Thanks for your help. All Tests Pass.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: 1, sort by ascii.
2, except empty value and "sign".
3, signature (must original value).
4, url encode.
5, send.

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
```diff
@@ -54,5 +54,10 @@ public async Task Can_Sign_In_Using_Alipay(string claimType, string claimValue)
                 AssertClaim(claims, claimType, claimValue);
             }
         }
+
+        private sealed class FixedClock : ISystemClock
+        {
+            public DateTimeOffset UtcNow => new DateTime(2019, 12, 14, 22, 22, 22, DateTimeKind.Local);
```
martincostello: Sorry, my mistake here to use the wrong type in the snippet I gave you. It should be:

```csharp
public DateTimeOffset UtcNow => new DateTimeOffset(2019, 12, 14, 22, 22, 22, TimeSpan.Zero);
```

vicenteyu: It's my mistake, I set "timestamp" .ToLocalTime(),  so you tests has not pass. I will change all to utc. it will be ok.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -0,0 +1,64 @@
...
+            /// </summary>
+            public const string UserStatus = "urn:alipay:user_status";
+
+            /// <summary>
+            /// Whether the user is certified. T: True; F: False;
+            /// </summary>
+            public const string IsCertified = "urn:alipay:is_certified";
+
+            /// <summary>
+            /// The user's gender. F: Femal; M: Male;
```
martincostello: Typo: `Femal` -> `Female`

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: How about this instead? It saves allocating lots of intermediate strings, and also from what I can see the sorted dictionary never contains `"sign"` anymore, so the code to filter that isn't needed:

```csharp
private string GetRSA2Signature(SortedDictionary<string, string> source)
{
    var builder = new StringBuilder();

    foreach (var pair in source)
    {
        if (string.IsNullOrEmpty(pair.Value))
        {
            continue;
        }

        builder.Append(pair.Key)
               .Append('=')
               .Append(pair.Value)
               .Append('&');
    }

    builder.Remove(builder.Length - 1, 1);

    byte[] dataBytes = Encoding.UTF8.GetBytes(builder.ToString());

    using var privateKeyRsaProvider = CreateAlgorithm();
    byte[] signatureBytes = privateKeyRsaProvider.SignData(dataBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);

    return Convert.ToBase64String(signatureBytes);
}
```

test/AspNet.Security.OAuth.Providers.Tests/Alipay/AlipayTest.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public override string DefaultScheme => AlipayAuthenticationDefaults.AuthenticationScheme;
+
+        protected internal override void RegisterAuthentication(AuthenticationBuilder builder)
+        {
+            builder.Services.AddSingleton<ISystemClock, FixedClock>();
+            builder.AddAlipay(options =>
+            {
+                ConfigureDefaults(builder, options);
+                //test secret
```
martincostello: I think you can just remove this comment.

martincostello (approved):

martincostello:
If you'd like to add yourself to the [contributors](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers#contributing) feel free to do that, otherwise I'll merge this a bit later today 👍

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        /// <summary>
+        /// Gets an HTTP payload as an asynchronous operation.
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="elementName">Name of the element.</param>
+        /// <returns>Task&lt;JsonDocument&gt;.</returns>
+        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
+        {
+            var messageContent = await message.Content.ReadAsStringAsync();
```
kevinchalet: This method seems rather inefficient and dangerous, as `JsonDocument` instances won't be properly disposed here. Can't we just use `JsonSerialize.DeserializeAsync<JsonElement>(stream)` and return `JsonElement`?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+        public AlipayAuthenticationHandler(
+            [NotNull] IOptionsMonitor<AlipayAuthenticationOptions> options,
+            [NotNull] ILoggerFactory logger,
+            [NotNull] UrlEncoder encoder,
+            [NotNull] ISystemClock clock)
+            : base(options, logger, encoder, clock)
+        {
+        }
+
+        protected override string FormatScope() => string.Join(",", Options.Scope);
```
kevinchalet: Bloody Chinese providers... :rofl:

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+            var algorithm = RSA.Create();
+
+            try
+            {
+                algorithm.ImportRSAPrivateKey(privateKeyBytes, out int _);
+                return algorithm;
+            }
+            catch (Exception)
+            {
+                algorithm?.Dispose();
```
kevinchalet: I'd remove the `?` as there's no reason for `algorithm` to be null here.

kevinchalet (changes requested):

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: Ah, my bad. This is copy-pasta from the Apple provider that must’ve got leftover after a refactor.

martincostello: Ah, my bad. This is copy-pasta from the Apple provider that must’ve got leftover after a refactor.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: How about: 
`
        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
        {
            var messageStream = await message.Content.ReadAsStreamAsync();

            var document = await JsonSerializer.DeserializeAsync<JsonElement>(messageStream);
            var elementContent = document.GetProperty(elementName).GetRawText();

            return JsonDocument.Parse(elementContent);
        }
`

vicenteyu: How about : 
```
        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
        {
            var messageStream = await message.Content.ReadAsStreamAsync();

            var document = await JsonSerializer.DeserializeAsync<JsonElement>(messageStream);
            var elementContent = document.GetProperty(elementName).GetRawText();

            return JsonDocument.Parse(elementContent);
        }

File: src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs [hunk 54]
```diff
@@ -0,0 +1,231 @@
...
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="elementName">Name of the element.</param>
+        /// <returns>Task&lt;JsonDocument&gt;.</returns>
+        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
+        {
+            var messageStream = await message.Content.ReadAsStreamAsync();
+
+            var document = await JsonSerializer.DeserializeAsync<JsonElement>(messageStream);
+            var elementContent = document.GetProperty(elementName).GetRawText();
```
kevinchalet: I know `System.Text.Json`'s limited API surface sometimes requires doing weird things, but in this particular case, I still don't understand why you need to export the RAW JSON of your `elementName` node to re-parse it. Why not directly returning `elementContent`?

kevinchalet (changes requested):

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: public static OAuthTokenResponse Success(JsonDocument response);

JsonElement must be convert to JsonDocument.

kevinchalet: Understood, but it's only required in `ExchangeCodeAsync`, not in `CreateTicketAsync` where you use `payload.RootElement`. Let's get rid of this private method and only do the `GetRawText()` trick for `ExchangeCodeAsync`.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        /// <summary>
+        /// Gets an HTTP payload as an asynchronous operation.
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="elementName">Name of the element.</param>
+        /// <returns>Task&lt;JsonDocument&gt;.</returns>
+        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
+        {
+            var messageStream = await message.Content.ReadAsStreamAsync();
```
kevinchalet, line 173: One question: are we 100% sure this provider ALWAYS returns UTF8-encoded payloads? Since their OAuth 2.0 is almost 100% wrong, I wouldn't be surprised if they returned weird charsets.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: It's my mistake. 
how about: 
            var elementContent = document.GetString(elementName);
            return JsonDocument.Parse(elementContent);

kevinchalet: `GetString` will only work if the `elementName` node is a string. If it's a JSON object or array, it will fail.

vicenteyu: > ValueKind = Object : "{
  "alipay_system_oauth_token_response": {
    "user_id": "my-id",
    "access_token": "secret-access-token",
    "expires_in": "300",
    "refresh_token": "secret-refresh-token",
    "re_expires_in": "300"
  },
  "sign": "sign"
}"

var elementContent = document.GetString(elementName);
Return:

> {
    "user_id": "my-id",
    "access_token": "secret-access-token",
    "expires_in": "300",
    "refresh_token": "secret-refresh-token",
    "re_expires_in": "300"
  }

Seem to work.

kevinchalet: Oh yeah, that's the ASP.NET Core extension, not `JsonElement.GetString()`: https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs#L10

I'd keep `GetRawString()` to reduce potential confusion here.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
vicenteyu: Reqeust parameter ["charset"] = "UTF-8", so returns UTF-8.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -168,15 +168,15 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
         /// </summary>
         /// <param name="message">The message.</param>
         /// <param name="elementName">Name of the element.</param>
-        /// <returns>Task&lt;JsonDocument&gt;.</returns>
-        private async Task<JsonDocument> ParseJsonAsync(HttpResponseMessage message, string elementName)
+        /// <returns>Task&lt;System.String&gt;.</returns>
```
martincostello: `<returns><see cref="Task{T}"/></returns>`, though it's not adding much to this method here so maybe just remove the comments from it?

vicenteyu: Removed.

vicenteyu:
Hi @martincostello ,  why removed this from the 3.1.0 milestone?

martincostello:
Kévin hasn’t yet approved the PR and there’s had been no activity for a month at that point.

There’s no date for a 3.1 release at the moment as we’ve both been busy with other things.

But just because the milestone has been removed doesn’t mean it won’t be included with it, but it’s no longer “ready-to-go” as it was when I added it.

vicenteyu:
OK. Thank you. :)

martincostello:
Hi @vicenteyu - are you planning to do any further work on this PR? There are still unresolved comments.

If so, please rebase onto the dev branch. The next release is planned for around late March when .NET Core 3.0 reaches end-of-life.

PR #364, opened 2019-12-14
Update to .NET Core 3.1.101 SDK and support netcoreapp3.1

martincostello:
This PR updates Arcade and the .NET Core SDK to support .NET Core 3.1.

It also updates the providers and tests to dual-target `netcoreapp3.0` and `netcoreapp3.1`, the idea being that 3.1.0 of the providers will explicitly support both, then after 3.0 goes out of support in March 2020 we can just target `netcoreapp31`. The sample client is updated to only support `netcoreapp3.1`.


martincostello:
@PinpointTownes Thoughts on the dual-targeting?

I figured that the #362 change would want releasing by the author once it's merged, so we could combine that with an update for ASP.NET Core 3.1 without forcing 3.1.0 as a minimum version/TFM, so that all the providers change, instead of just having to publish the Alipay one.

Then once 3.0 goes out of support in March next year it can just go back to a single TFM against the 3.1 LTS version.

kevinchalet:
@martincostello do we rely on 3.1-only APIs? Could targeting 3.1 allow us to reduce the dependencies graph? If not, it would be likely simpler and better to keep targeting `3.0` and move to `3.1`-only in March.

martincostello:
Not explicitly for the 3.1 APIs, but I’d noticed that various 3.1.0 packages that aren’t part of the shared framework only target `netcoreapp31` now, so I wonder if there’s anything that becomes simpler for 3.1 apps using our packages if there’s an explicit TFM for it rather than referencing them via `netcoreapp30`.

That said I’m not tied to the 3.1 TFM so can remove it and just leave 3.1.100 building 3.0, but if we ship the Alipay and other packages as 3.1.0 then the `netcoreapp31` versions would be 3.2.0.

Directory.Build.props
```diff
@@ -42,7 +42,7 @@
     <Product>aspnet-contrib</Product>
     <Company>$(Authors)</Company>
     <_ProjectCopyright>© AspNet.Security.OAuth.Providers contributors. All rights reserved.</_ProjectCopyright>
-    <PackageIcon>package-icon.png</PackageIcon>
+    <PackageIconFullPath>$(MSBuildThisFileDirectory)package-icon.png</PackageIconFullPath>
```
kevinchalet, line 45: Dunno if it's related to this particular change, but our icon now appears twice in each package:

image

martincostello: Yeah, I noticed that too.

With what was there before it complained that it couldn’t add the .NET logo as we had already specified ours, and if I overrode this and left before then it complained about there still being two logos.

Maybe I’ll try reverting back and blanking out this new one and seeing if it goes away.

This might be another one of those annoying hard-to-override arcade things again like we had with the build number.

martincostello: Fixed it - Arcade adds the relevant item to include the icon itself by convention somehow, so I just needed to remove the line where I was adding ours manually.

kevinchalet: Thanks!

eng/Version.Details.xml
```diff
@@ -3,13 +3,13 @@
   <ProductDependencies>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="1.0.0-beta.19405.1">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="1.0.0-beta.19517.3">
```
kevinchalet, line 6: Is there any chance there's a non-beta version of Arcade available somewhere?

martincostello: I'll have a snoop and see what I can find. Wouldn't surprise me if there wasn't though.

I remember they only made a 1.0.0 of SourceLink because people thought it hadn't been "officially" released yet. I think stuff they don't ship just seems to be a forever increasing beta.

martincostello: Looking at dotnet/aspnetcore, they're on a beta of version 5 now:

https://github.com/dotnet/aspnetcore/blob/602f4678c1e764e4e4c4bcdf26a5f83abe39ecdb/eng/Version.Details.xml#L415

Directory.Build.props
kevinchalet: BTW, is that property 100% required? Looks like EF was updated to not set it explicitly (as you said, it's likely some sort of convention used by Arcade): https://github.com/dotnet/efcore/pull/18805/files

martincostello: If I remember correctly, this is so it doesn't use the default Microsoft icon.

Unless if there's a specific file name our icon could be instead and then it'd "just work".

martincostello: Just tried renaming it to `packageIcon.png` in the root, and we get the .NET icon instead.

kevinchalet: Perfect, thanks! 😄

eng/Version.Details.xml
kevinchalet: Roger. If `1.0.0-beta.19517` works fine with 3.1, we can wait before moving to 5.0.

martincostello:
@kevinchalet So I can get this PR finally merged in and closed off, what do you want to do on the TFMs?

I think the options are:
  1. Dual-target `netcoreapp3.0` and `netcoreapp3.1` as a 3.1.0 release, and ship it "soon", with a 3.2.0 for `netcoreapp3.1` only in March.
  1. Change to target `netcoreapp3.1` only as a 3.1.0 release, and aim to ship some time in March once 3.0 is EOL'd.

I'm happy to go either way. Option 1 lets us get new providers being added by the community shipped to NuGet.org sooner, where as Option 2 is less to maintain.

kevinchalet:
The lazy me really likes option 2. 

If you feel March is not soon enough to ship the next aspnet-contrib release, maybe we can ship a 3.1-only version later this month?

martincostello:
Nah, option 2 is fine.

Maybe if I'd not been lazy and fixed the duplicated icon thing two months ago back in December there'd have been more of an argument to not delay until March 😅

I'll fix-up the PR tomorrow to just target `netcoreapp3.1` and let any contributors who ask know that we're aiming for a release in March after the 3.0 EOL 👍

Issue #369, opened 2020-01-02
Add Deezer OAuth 2.0 Provider


tolbxela:
I am going to implement an OAuth 2.0 Provider with the Deezer API
Documentation: [Deezer API / OAuth](https://developers.deezer.com/api/oauth)

tolbxela:
I suppose, you will close the Issue. Or should I do it?

martincostello:
I'll close it down when #373 is merged 👍

martincostello:
Resolved by #373.

PR #372, opened 2020-01-02
Deezer OAuth 2.0 Provider

tolbxela:
Issue: aspnet-contrib/AspNet.Security.OAuth.Providers#369

The Deezer OAuth 2.0 Provider was implemented.

Documentation: [Deezer API - OAuth](https://developers.deezer.com/api/oauth)
APP Registration at [Deezer Developer - My Apps](https://developers.deezer.com/myapps)

The Deezer API Permission is set by default to **basic_access** ( allow to access users basic information: name, firstname, profile picture) and **does not allow to get the email**.

Here is the config with the **email** permission:

```
.AddDeezer(options =>
{
    options.App_Id = "app_id";
    options.Secret = "secret";
    options.Permissions = new List<string> {
        DeezerAuthenticationConstants.Permissions.Basic_Access,
        DeezerAuthenticationConstants.Permissions.Email
    };
});
```


tolbxela:
The implementation of Permissions was wrong. I will implement they through scopes.

PR #388, opened 2020-02-19
Map name identifier to email

denis-ivanov:
Mail.ru for some reasons doesn't return name identifier like other providers. When we use current options after first enter with mail.ru other users will enter as first user because they have the same name identifier. client_id - it's an application identifier, not a user's.

### DIFF
--- a/src/AspNet.Security.OAuth.MailRu/MailRuAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.MailRu/MailRuAuthenticationOptions.cs
@@ -25,7 +25,7 @@ namespace AspNet.Security.OAuth.MailRu
             TokenEndpoint = MailRuAuthenticationDefaults.TokenEndpoint;
             UserInformationEndpoint = MailRuAuthenticationDefaults.UserInformationEndpoint;
 
-            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "client_id");
+            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "email");
             ClaimActions.MapJsonKey(ClaimTypes.Name, "nickname");
             ClaimActions.MapJsonKey(ClaimTypes.Gender, "gender");
             ClaimActions.MapJsonKey(ClaimTypes.Email, "email");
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR! 👏 

@martincostello looks like 2.2.0 is also impacted. Should we consider backporting fixes like this one to 2.x?

martincostello (approved):

martincostello:
@kevinchalet I guess we'd want to fork of off 2.1 and fix it there for anyone who might be LTS?

As 2.2 is out-of-support now, do we need to worry about that version?

kevinchalet:
@martincostello wooops, sorry, I should have been clearer: I meant _our_ 2.2.0 release, that targets ASP.NET Core 2.0 (and so can used on 2.1 LTS) 😄

martincostello:
Ah - I thought they were targeting `netcoreapp2.x`, but I just looked at the branch and saw it's `netstandard2.0`.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/768f34c99ee65ea8b6a180a2427b134ae696c821/src/AspNet.Security.OAuth.MailRu/AspNet.Security.OAuth.MailRu.csproj#L6

Yeah, in that case I think we should backport the fix there too.

I'll look at merging this and doing the backports later this evening. Should we backport for 3.0 as well as that isn't _yet_ EOL?

kevinchalet:
> Should we backport for 3.0 as well as that isn't yet EOL?

I'd say "not worth it", but if you think it's preferable, it's fine too 😄

kevinchalet (approved):

martincostello:
Cool - I'll sort these out later this evening then 👍

martincostello:
Have pushed up a [`rel/2.2.1`](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/rel/2.2.1) branch containing the fix.

martincostello:
Two packages containing this fix are now available on NuGet.org:
  * [2.2.1](https://www.nuget.org/packages/AspNet.Security.OAuth.MailRu/2.2.1)
  * [3.1.0](https://www.nuget.org/packages/AspNet.Security.OAuth.MailRu/3.1.0)

Issue #391, opened 2020-02-28
Sign in with Apple documentation


martincostello:
The new _Sign in with Apple_ provider generates a relatively large amount of issues, which are often configuration related and not actual bugs.

To try and reduce some of these issues, a small amount of documentation should be added somewhere (a Markdown file in the repo linked to by the README?) that gives a short and high-level documentation example of using it. That can then link off to other resources to go into more detail so it doesn't turn into a sprawling document that needs to be maintained too much.

martincostello:
Once I've done this (which I'll try to do later today), I'll create a branch to cut the 3.1.0 release.

@kevinchalet Now that I've got the appropriate permissions in NuGet.org, shall I do the package publish for 3.1.0 once it's built and ready, or do you want to do the honours for that?

PR #397, opened 2020-03-26
Replace singleton-like cache with IMemoryCache for Apple

### LINKED ISSUE #396: Replace Sign in with Apple DefaultAppleClientSecretGenerator's `_clientSecret` with `IMemoryCache`
I want to use it with multiple `ClientId`. I don't use default OAuth callback handler, and I make `*Context` and call `AppleAuthentication*`'s interfaces myself.  
But `DefaultAppleClientSecretGenerator` strongly assumes that there's only one `ClientId` per app, and it is registered as singleton service.
Since I don't have access to the `Internal.Default*` classes. I've made my copy of it and modified it to use `IMemoryCache` to handle multiple `ClientId`.

Use `IMemoryCache` rather than de facto global singleton like `_clientSecret`

Disable client secret cache?

**Additional context**
I already have code so I can make a PR.

foriequal0:
Resolves: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/396

### DIFF
--- a/src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
+++ b/src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
@@ -11,6 +11,7 @@ using System.Security.Cryptography;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Tokens;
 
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
```diff
@@ -22,39 +23,49 @@ internal sealed class DefaultAppleClientSecretGenerator : AppleClientSecretGener
...
-
-        private string? _clientSecret;
-        private DateTimeOffset _expiresAt;
+        private readonly IMemoryCache _cache;
 
         public DefaultAppleClientSecretGenerator(
             [NotNull] AppleKeyStore keyStore,
             [NotNull] ISystemClock clock,
             [NotNull] JwtSecurityTokenHandler tokenHandler,
+            [NotNull] AppleMemoryCache cache,
```
martincostello: Why not just inject `IMemoryCache` directly, rather than create a whole separate cache?

foriequal0: Directly inject `IMemoryCache` would make key collision and setting collision.
I followed this document: https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-3.1#use-setsize-size-and-sizelimit-to-limit-cache-size
I didn't set `SizeLimit`, and don't specify cache entry's `Size`. If someone tries to set `SizeLimit` on `.AddMemoryCache()`, then it would throw an exception and vise versa.

martincostello: You could avoid a key collision by making the key namespace larger and more specific to the implementation (e.g. `DefaultAppleClientSecretGenerator+...`).

The limit should really be up to the owner of the integrating app via their existing cache setup, rather than creating an extra private cache that needs extra setup. If someone wants a size limit and it gets breached and throws that would be what they'd configured it to do.

foriequal0: Okay. I'll change that.

kevinchalet: IMHO, things are starting to look quite messy here. Shouldn't we remove all these DI-resolved dependencies and move them to the options class? I suspect `JwtSecurityTokenHandler` is also one of those things people generating their own JWT tokens will register themselves, which may collide with the instance we use (e.g they could customize the claims mapping).

martincostello: The DI registrations all use `TryAdd...()` so that users' own things aren't overridden if they're registered already so we honour their own registrations if they're different:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/8e4c19008f518f3730bab90a980e01347ba6f3d3/src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs#L81

kevinchalet: True, it means the Apple handler will use the `JwtSecurityTokenHandler` configured by the users - which may be configured to use things like claims mapping or have some of its methods overridden - and not the "default" one we'd register by default. My point is that users may change the behavior of the Apple handler without even realizing it, which is not good.

martincostello: Is that something we can handle via docs as part of #391? I'd rather not radically (as in add lots of new options members that don't come from DI anymore) change the provider right before publishing the 3.1 version.

martincostello:
I have ported this change to a new PR in #484 which will be part of the 5.0 release.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
martincostello: Addressed as part of #484.

PR #408, opened 2020-04-14
Add apple email claim

### LINKED ISSUE #407: The presence of a "System.Security.Claims.ClaimTypes.Email" claim is not present for the Apple provider.
**Describe the bug**
There appears to be a bug in the Apple Provider.  Specifically the AppleAuthenticationHandler.cs file.  I noticed today, when debugging an app that utilizes the Apple provider, that the "System.Security.Claims.ClaimTypes.Email" claim was not present in the claims.  However, the claim of type "email" was present in the claims.  It looks like the method named "ExtractClaimsFromUser" is trying to add the "System.Security.Claims.ClaimTypes.Email" claim, however it appears to never be called.  I debugged the code and this is precisely what is happening.  There is a check for the existence of a "user" property in the parameters list, however there is no user property, therefore the "ExtractClaimsFromUser" method never gets called and the "System.Security.Claims.ClaimTypes.Email" claim is never added to the claims.  The app that I noticed this issue with utilizes the "authorization_code" flow, if that helps.  Also, the Apple user I was using to authenticate with did NOT hide the email.  I will submit a pull-request shortly with a proposed code change to resolve this issue.

**Steps To reproduce**
Authenticate via an authorization_code flow utilizing the Apple provider and do NOT choose to "Hide" your email.

**Expected behaviour**
The presence of a "System.Security.Claims.ClaimTypes.Email" claim should be in the list of claims.

**Actual behaviour**
The claim of type "email" was present in the claims.  

**System information:**
 - OS: [e.g. Windows 10]
 - Library Version [e.g. 2.2.0]
 - .NET version 2.0

lukefulliton:
Adding a "System.Security.Claims.ClaimTypes.Email" claim to the list of claims if the presence of an identical claim type is not present.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/407


src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
```diff
@@ -137,19 +138,28 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
         /// Extracts the claims from the token received from the token endpoint.
         /// </summary>
         /// <param name="token">The token to extract the claims from.</param>
+        /// <param name="identity">Represents a claims-based identity</param>
```
martincostello:
```suggestion
        /// <param name="identity">Represents a claims-based identity.</param>
```

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
```diff
@@ -137,19 +138,28 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
         /// <returns>
         /// An <see cref="IEnumerable{Claim}"/> containing the claims extracted from the token.
         /// </returns>
-        protected virtual IEnumerable<Claim> ExtractClaimsFromToken([NotNull] string token)
+        protected virtual IEnumerable<Claim> ExtractClaimsFromToken([NotNull] string token, [NotNull] ClaimsIdentity identity)
```
martincostello: This would be a breaking change to the API surface, so can't be part of a v3 release. Is there a way to refactor this in such a way as to not require a breaking change and still fix the issue?

martincostello:
Also can you add a regression test for this scenario please?

martincostello:
Would a simpler fix be to add a claim action to the options that adds the email claim if it's not already been added from the response from Apple?

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/e3b31779701f7a38ef1da6f25a77065d6b6caf13/src/AspNet.Security.OAuth.Apple/AppleAuthenticationOptions.cs#L21-L33

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
lukefulliton: Good point.  I will work towards a different resolution.

lukefulliton:
I like your proposed solution, however in order to implement this we would need to be able to access the JwtSecurityTokenHandler inside the AppleAuthenticationOptions in order to read the security token and map to the email claim.  Naturally we have access to the "response" and the "id_token" but since that "id_token" is a JWT, I am not sure how we could access the "email" value inside the JWT inside the AppleAuthenticationOptions?  Also, it seems cleaner to leave the logic for this inside the AppleAuthenticationHandler.  I will work to submit a new pull-request and see what you think.  Thank you for your collaboration on this issue.  Let me know if you have any additional thoughts.

lukefulliton:
Also, if you are okay with it, I would like to remove the initial attempt to populate the "ClaimTypes.Email" claim from the parameters Dictionary.  This initial attempt occurs inside the "ExtractClaimsFromUser" method and it looks like the source is either the "Request.Form" or the "Request.Query" ,  however obtaining the user's email address from the "id_token" would be a more trusted source.  What are your thoughts?

lukefulliton:
I am not as familiar with some of the other OAuth grants.  Is there typically a "user" object returned in the JSON?  And the "authorization_code" grant omits the user object due to the nature of the token exchange?  If so, then we may need to leave the "ExtractClaimsFromUser" method as is.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
lukefulliton: What are your thoughts about overloading this method?

lukefulliton:
Also, to answer your question about adding a regression test.  I will need a new "id_token" that contains an "email" claim, in the bundle.json file.  I believe you will have to generate that because the "id_token" that is currently there does not have an "email" claim.  If you send it to me, then I can add a test.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
martincostello: This is still a breaking change, but now functionally. The original method is no longer called, so if a caller had overridden it their customised implementation would no longer be called. To preserve that behaviour, you'd have to call the existing method from the new overload.

lukefulliton: Do you have a preference?  I could call the original method from the overloaded method as you suggest, or use the new overloaded method that is called right alongside, but not instead of, the original method?

martincostello: Is there a way to edit the logic in the existing method and put the new code before or after it's called? Then we don't need to add any new members and it just becomes an internal change in behaviour.

lukefulliton: Yes, we could add the logic to the existing "ExtractClaimsFromToken" method, however, without the ClaimsIdentity "identity" value, there is no way to check if the "Email" claim had already been added to the "identity" by the "ExtractClaimsFromUser" method.  So the "identity" could end up with a duplicate "Email" claim.

lukefulliton: I have the new overloaded method calling the original method at present.  That code change is pushed up.  You can see the dilemma and how the "identity" value is used to prevent the duplicate "Email" claim.  I enjoy the collaboration, so let me know if you see an alternative that you prefer.

martincostello:
I'll check out this branch later today and have a play around with some different approaches with the debugger running to see if there's a way we can make this "just work" without having to add new members or double-process the ID token.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
lukefulliton: I created an alternative solution for comparison and consideration here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/409

lukefulliton:
Awesome, sounds good.

alaincroisetiere:
Hi, can you release this fix in a 2.x version ? This change will fix the aspnet identity integration.

martincostello:
> Hi, can you release this fix in a 2.x version ? This change will fix the aspnet identity integration.

Once we have a fix, I’ll look into porting it to the 2.x branch.

martincostello:
Superseded by #411.

PR #413, opened 2020-04-26
Add ExactOnline provider

FWest98:
Adds an ExactOnline provider including unit tests and README update.


src/AspNet.Security.OAuth.ExactOnline/ExactOnlineAuthenticationDefaults.cs
```diff
@@ -0,0 +1,52 @@
...
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.OAuth;
+
+namespace AspNet.Security.OAuth.ExactOnline
+{
+    /// <summary>
+    /// Defaults used for exact online authentication provider.
```
martincostello:
```suggestion
    /// Defaults used for Exact Online authentication provider.
```

src/AspNet.Security.OAuth.ExactOnline/ExactOnlineAuthenticationHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+
+            using var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            if (!response.IsSuccessStatusCode)
+            {
+                Logger.LogError("An error occurred while retrieving the user profile: the remote server " +
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                    /* Status: */ response.StatusCode,
+                    /* Headers: */ response.Headers.ToString(),
+                    /* Body: */ await response.Content.ReadAsStringAsync());
```
martincostello:
```suggestion
                                /* Status: */ response.StatusCode,
                                /* Headers: */ response.Headers.ToString(),
                                /* Body: */ await response.Content.ReadAsStringAsync());
```

test/AspNet.Security.OAuth.Providers.Tests/ExactOnline/bundle.json
```diff
@@ -0,0 +1,61 @@
...
+        "d": {
+          "results": [
+            {
+              "__metadata": {
+                "uri": "https://start.exactonline.nl/api/v1/current/Me(guid'00000000-0000-0000-0000-000000000000')",
+                "type": "Exact.Web.Api.System.Me"
+              },
+              "CurrentDivision": 12345,
+              "DivisionCustomer": "00000000-0000-0000-0000-000000000001",
+              "DivisionCustomerCode": "          12345678",
```
martincostello, line 28: Should this have all the leading spaces?

```suggestion
              "DivisionCustomerCode": "12345678",
```

src/AspNet.Security.OAuth.ExactOnline/ExactOnlineAuthenticationHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                throw new HttpRequestException("An error occurred while retrieving the user profile.");
+            }
+
+            using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
+            var root = payload.RootElement.GetProperty("d").GetProperty("results")[0];
+
+            var principal = new ClaimsPrincipal(identity);
+            var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, root);
+            context.RunClaimActions();
```
martincostello: What about something similar the below code from the Foursquare provider instead?

Then the `OAuthCreatingTicketContext` has access to the root response, rather than just the bit we've extracted from it, and we can also be a bit more defensive about checking the response, rather than assuming they array will have one item in it from the `[0]`? Then if something changed server side in the future and broke things it would be a bit easier to debug than throwing an `IndexOutOfRangeException` or similar.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/28f6c0d8114f3195c0897af60304bb12d37d9517/src/AspNet.Security.OAuth.Foursquare/FoursquareAuthenticationHandler.cs#L61-L73

martincostello:
Thanks for the PR @FWest98 - just a few comments, most minor.

src/AspNet.Security.OAuth.ExactOnline/ExactOnlineAuthenticationHandler.cs
FWest98: The entire API is an OData API, so this d/results/0 part is not specific for ExactOnline, but part of the OData standard. If they were to move away from OData, then I think many more things break. That's why I thought it was safe to keep it this way.

test/AspNet.Security.OAuth.Providers.Tests/ExactOnline/bundle.json
FWest98: Yes (unfortunately). These codes are defined as a 15-character numeric string, left-padded with spaces.

src/AspNet.Security.OAuth.ExactOnline/ExactOnlineAuthenticationHandler.cs
martincostello: In that case maybe something like this would be better. It makes it clearer that the code is expecting an array and also makes the entire JSON response availble to the `OAuthCreatingTicketContext`:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/28f6c0d8114f3195c0897af60304bb12d37d9517/src/AspNet.Security.OAuth.StackExchange/StackExchangeAuthenticationHandler.cs#L79-L80

So it would end up like:

```csharp
var user = payload.RootElement
    .GetProperty("d")
    .GetProperty("results")
    .EnumerateArray()
    .First();

var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
context.RunClaimActions(user);
```

FWest98: Ah, that is a good idea. I did not know the `RunClaimActions` accepted this as extra parameter.

martincostello (approved):

martincostello:
Thanks for the contribution @FWest98! 🎈 

We haven't planned when the next release will be yet (we only just shipped 3.1.0), but in the meantime you can consume the new provider from the MyGet feed that builds from CI: https://www.myget.org/gallery/aspnet-contrib

martincostello:
This provider is now available from NuGet.org: [AspNet.Security.OAuth.ExactOnline 3.1.1](https://www.nuget.org/packages/AspNet.Security.OAuth.ExactOnline/3.1.1)

Thanks again for your contribution @FWest98.

PR #423, opened 2020-05-08
Add GitHub Actions CI

martincostello:
Adds a workflow for using GitHub Actions for CI (build, test, package, optional publish).

The rationale for the suggestion to change the CI process are:
  1. Builds and tests for all three platforms in one provider, rather than the two used today.
  1. Centralises configuration here into GitHub into one file YAML (excluding secrets).
  1. Supports automatic publish of NuGet packages for those with push access (if secrets configured) to:
      1. MyGet for all branches (but not PRs or forks)
      1. NuGet.org for tags (i.e. when we publish a release), so no manual download from AppVeyor and scripting upload to NuGet.org required

I've based the YAML for this on what I've set up for some of my own repos having migrated them onto GitHub Actions for build test and publish.

## TODO

- [ ] Disable Travis CI builds for pushes and PRs
- [ ] Configure AppVeyor to ignore branches without an `appveyor.yml` file
- [x] ~Configure `MYGET_TOKEN` and `NUGET_TOKEN` in the repo's secrets.~
- [x] ~Uncomment the steps to upload the NuGet packages~


martincostello:
Example build here: https://github.com/martincostello/AspNet.Security.OAuth.Providers/actions/runs/99265431

kevinchalet:
Sounds like a good idea. I like the fact we only push packages built from tags to NuGet. 

> Builds and tests for all three platforms in one provider, rather than the two used today.

Just so things are 100% clear, you'd like to remove the AppVeyor and Travis integration?

.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+            os_name: linux
+          - os: windows-latest
+            os_name: windows
+
+    steps:
+
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Setup .NET Core SDK
```
kevinchalet, line 32: I guess it's only required to be able to use `dotnet nuget`, as Arcade is supposed to provide us with the .NET Core SDK for the build part, right?

martincostello:
If you think it's a good idea, yes. I left everything else around in the branch for the time being to see what you thought of the proposal first 😃

.github/workflows/build.yml
martincostello: I think arcade would use the same scripts to download this anyway, but my understanding is that the action does some caching stuff so that it might end up being quicker.

That said, this is just copy-paste from my own non-Arcade stuff, so I could also just remove it.

martincostello: Wait, I’m an idiot. I misread `dotnet NuGet` as `dotnet restore`, so actually, yes it is needed! 😅

kevinchalet:
> If you think it's a good idea, yes.

In this case, it's something we should do globally for all the aspnet-contrib repos and all the active branches. Just to be sure there's no nasty surprise, could you please try with the 2.x branch, that uses KoreBuild?

martincostello:
Sure, I'll do PR for that later today and try it out. I'll do that on a branch here rather than my fork so it'll run the checks in this repo.

Once we were happy with this and it was merged and configured, I was planning on doing it for the 2.x branch and the OpenID repo too as well.

martincostello:
Have pushed up a branch for 2.2 here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/GitHub-Actions-22x

The build doesn't seem to have started, so I'm going to guess it needs enabling or something.

kevinchalet:
> The build doesn't seem to have started, so I'm going to guess it needs enabling or something.

It's weird, actions are already enabled in this repo:

image

martincostello:
Hmm - maybe there's some sort of permissions thing because it's new? (as in, it's a new workflow in the repo)

martincostello:
Huh - it's started building this time!

kevinchalet:
Nice 😄 

That said, I'm wondering if we shouldn't keep AppVeyor and Travis enabled, for the (hopefully rare) cases where GitHub Actions would be down?

Do you need help for the `MYGET_TOKEN`/`NUGET_TOKEN` thing?

martincostello:
Keeping a backup for “manual” publish if they go down is up to you. I’m happy to keep it and have three CIs if you want it, but I don’t think (hope) there’d be no need for us to publish a package so urgently that we couldn’t wait for an outage to resolve.

Yeah, I’d need you to configure the secrets please as I don’t have access.

martincostello:
@kevinchalet Thoughts on the above?

kevinchalet:
> @kevinchalet Thoughts on the above?

I agree, it's likely not necessary. I'll take some time today to add the MyGet and NuGet API keys in the organization secrets. I'll let you know when it's done.

Directory.Build.props
```diff
@@ -22,22 +22,22 @@
...
+    Arcade only allows the revision to contain up to two characters, and GitHub Actions does not roll-over
     build numbers every day like Azure DevOps does. To balance these two requirements, set the official
     build ID to be the same format as the built-in default from Arcade, except with the revision number
     being the number of the quarter hour of the current time of day (24 * 4 = 96, which is less than 100).
     So a build between 00:00 and 00:14 would have a revision of 1, and a build between 23:45 and 23:59:59
     would have a revision of 97.
   -->
 
-  <PropertyGroup Condition=" '$(APPVEYOR)' == 'True' AND '$(APPVEYOR_PULL_REQUEST_NUMBER)' == '' ">
+  <PropertyGroup Condition=" '$(GITHUB_ACTIONS)' == 'true' AND '$(GITHUB_REF.StartsWith(`refs/pull/`))' == 'false' ">
```
kevinchalet, line 34: FYI: while it's a great workaround, it sadly proved to be a bit unreliable in the OpenIddict repo, where builds tend to last longer than in this repo: let's say you push a commit at 14:25, if `dotnet build/pack` is still running at 14:30, then the packages built and packed after 14:30 will get a different version number than the previous ones, which causes issues when the packages produced after depend on packages built before. It's very likely because these properties are always re-evaluated per project.

Now that we're moving to GitHub actions, do you think we could move this logic to a custom action?

martincostello: Ah interesting. So you mean an action that would run at the start of the build and set the number in an environment variable, and then the build would just reference that so it's constant for that specific build?

kevinchalet: Yes, absolutely! Or maybe there's a way to compute that directly in the build.yml file using inline expressions? 🤷‍♂️

martincostello: I know you can persist env vars - I'd be happy to look into it. OK for me to do that as a follow up PR?

kevinchalet: > OK for me to do that as a follow up PR?

Sure, thanks! 👏

kevinchalet:
> I'll take some time today to add the MyGet and NuGet API keys in the organization secrets. I'll let you know when it's done.

Done, the two secrets are named `MYGET_API_KEY` and `NUGET_API_KEY` 😃

martincostello:
Great, thanks! I'll look at getting this merged a bit later today 🥳

kevinchalet:
@martincostello FYI, I removed this repository from AppVeyor and disabled the TravisCI webhook.

Do you plan to send a similar PR for the OpenID repo?

martincostello:
Cheers.

> Do you plan to send a similar PR for the OpenID repo?

Yep, will push one up once I've rebased my inflight PRs here 👍

kevinchalet:
Fantastic! Do you mind if I reuse your build file for OpenIddict? 😃

martincostello:
Not at all! 😄

kevinchalet:
While we're at it, maybe we should also start pushing our nightly builds to GitHub Packages?

martincostello:
> While we're at it, maybe we should also start pushing our nightly builds to GitHub Packages?

We can look into if if you like, but I've not _yet_ had success getting it to work. I experimented with it on one of my own repos a few months ago, but couldn't get the uploads to work due to vague authorization errors I couldn't seem to get past.

martincostello:
> Do you plan to send a similar PR for the OpenID repo?

https://github.com/aspnet-contrib/AspNet.Security.OpenId.Providers/pull/89

Directory.Build.props
martincostello: #448

PR #429, opened 2020-05-31
Remove SDK workaround

martincostello:
Remove HACK for bug which the 3.1.3xx SDK fixes.


martincostello:
Appears to be working as expected:

image

kevinchalet (approved):
If it fixes the bugs we had to work around when using the previous SDK version, then :shipit:

kevinchalet (approved):
🚢 🇮🇹

PR #443, opened 2020-07-02
Switch "public" with "read" since that is the base scope for Strava.

bryantlikes:
Based on https://developers.strava.com/docs/authentication/ there is no "public" scope for Strava.

Uses "read" instead of "public" as the default scope.


martincostello (approved):

martincostello:
Thanks @bryantlikes - I assume this came out of you testing it out with a real integration and this fixed it?

bryantlikes:
> Thanks @bryantlikes - I assume this came out of you testing it out with a real integration and this fixed it?

Yes, if you pass "public" you will get an error back from Strava:

{"message":"Bad Request","errors":[{"resource":"Authorize","field":"scope","code":"invalid"}]}

martincostello:
Thanks @bryantlikes! 🥳

martincostello:
Version 3.1.2 of the Strava provider which contains this fix is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Strava/

PR #444, opened 2020-07-02
Throw exception if user unchecks requested scopes

bryantlikes:
When the user is sent to Strava to grant access to the application it includes checkboxes for all the requested scopes (configured in the options). 

If the user unchecks any of these options, the authentication flow will still succeed and the application will get the access tokens to make API calls on behalf of the user. 

However, when these calls are made, it is assumed the access token includes the requested scopes, which in this case it will not and the request will fail. There is currently no way to know that this is what is occurring. 

The proposed change will throw an exception if the user does not grant the application the scope that it was requesting since this is an error.

I'm not sure if this is the best approach, it might be better to somehow translate the scopes to roles so that the application can determine the appropriate steps to take, but this is an attempt to solve the issue.


bryantlikes:
Had to add the StravaLoopbackRedirectHandler since the default handler doesn't return the granted scope in the query string like Strava does.

src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
```diff
@@ -32,6 +34,19 @@ public class StravaAuthenticationHandler : OAuthHandler<StravaAuthenticationOpti
...
+
+            if (scopeValues.Count == 0 || scopeValues.First().Split(",").Length < Options.Scope.Count)
+            {
+                var scopeEx = new Exception("User did not grant access to requested scopes.");
+                return HandleRequestResult.Fail(scopeEx);
+            }
+
+            return await base.HandleRemoteAuthenticateAsync();
+        }
+
```
martincostello: What about the suggested approach below instead?

Here any HTTP or standard OAuth authentication/authorization errors are handled first, then we give the user the _option_ of handling access being denied (as they effectively denied permission for a requested scope) in a custom manner of their own using the existing `AccessDenied` event.

Then you can just hook up a `OAuthEvents.AccessDenied` handler and the code will follow this flow:

https://github.com/dotnet/aspnetcore/blob/ef5ab43b6f7bfeaa3cfb2c0735715a7cb69fda37/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs#L244-L285

In your case, you can throw an exception, where as another user may chose to just continue regardless as they may consider the particular scope denied to be optional, rather than block the entire authentication flow due to _something_ being missing.

It also has the benefit of redirecting the user to the `AccessDeniedPath` for custom messaging and/or logic, rather than to an HTTP 500 error due to the exception.

```suggestion
        protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
        {
            HandleRequestResult result = await base.HandleRemoteAuthenticateAsync();

            if (result.Succeeded)
            {
                var scopeValues = Request.Query[OAuthChallengeProperties.ScopeKey];

                if (scopeValues.Count == 0 || scopeValues[0].Split(",").Length < Options.Scope.Count)
                {
                    result = await HandleAccessDeniedErrorAsync(result.Properties);
                }
            }

            return result;
        }
```

src/AspNet.Security.OAuth.Strava/StravaAuthenticationHandler.cs
```diff
@@ -1,9 +1,11 @@
-/*
+﻿/*
  * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
  * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
  * for more information concerning the license and the contributors participating to this project.
  */
 
+using System;
+using System.Linq;
```
martincostello: `StringValues` implements `IList<string>`, so you can just use `[0]` rather than `First()`, so you don't need this `using`.

test/AspNet.Security.OAuth.Providers.Tests/Strava/StravaTests.cs
```diff
@@ -18,6 +19,11 @@ public class StravaTests : OAuthTests<StravaAuthenticationOptions>
         public StravaTests(ITestOutputHelper outputHelper)
         {
             OutputHelper = outputHelper;
+
+            LoopbackRedirectHandler = new StravaLoopbackRedirectHandler()
```
martincostello: I've merged some refactoring into the dev branch to remove the need to create a custom loopback handler. If you rebase onto dev, this can then be simplified to just this and you can delete `StravaLoopbackRedirectHandler`:

```csharp
LoopbackRedirectHandler.LoopbackParameters.Add("scope", "read");
```

kevinchalet:
Thanks for your PR, @bryantlikes. That said, I'm not so sure we should implement that, for two reasons:

  - It's trivial to bypass: the user can intercept the callback (e.g using Fiddler and its breakpoint mode) and change the `scope` query string parameter before the request reaches the `/signin-*` endpoint. That's why in the **standard** OAuth 2.0 code flow, this parameter is not returned as part of the authorization response but by the token endpoint... 🤣 

  - It's applicable to any OAuth 2.0-based provider. If you think it's a nice-to-have feature, it would likely be more productive to add such a feature in the base handler maintained by MSFT.

martincostello:
I'm happy to defer to Kévin's opinion on this one - he knows a lot about the intricacies of the OAuth specs that I don't 😄

bryantlikes:
Yes, I realize it is easy to get around if the user wanted to, it was more to just catch scenarios where user's uncheck a scope without realizing what it will mean. I'll take a look to make sure Strava isn't returning the scopes with the token.

bryantlikes:
Ok, just verified that Strava does not send the scopes back in the token response, it only returns the tokens plus the athlete object (user details). The even have it in the documentation that the user needs to verify the scope that was returned with the code. 

https://developers.strava.com/images/getting-started-5.png

So if you guys don't think this shouldn't be in the Strava handler then I can cancel this PR and just leave it implemented in my local code as an override since I still need to make this check. In the meantime I'll reach out to Strava to see if they can correct their implementation.

bryantlikes:
I'm raising the issue with Strava since according to the OAuth documentation the scopes property is **required** if the granted scopes are different from the requested scopes. 

https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

It seems like there should be an option to handle this scenario if the scopes that are granted are different that what was requested.

kevinchalet:
@bryantlikes I opened https://github.com/dotnet/aspnetcore/issues/23749 on your behalf.

bryantlikes:
Sorry for all the comments, but in thinking about this more it seems like the best approach to handle this would be:

1) Have the asp.net oauth handler handle the scopes issue correctly. I'm not 100% sure how it should handle it, but the developer should be able to respond to this scenario. Maybe the scopes need to be stored along with the token? 

2) Until Strava fixes thier OAuth implementation, take the scopes from the URL and add them to the token response so that the oauth handler can respond appropriately (from #1). Then when Strava fixes their implementation this extra code can be removed. 

Does this sound like a better approach?

bryantlikes:
I'm going to close the PR since I came up with a solution that works for me. I updated the options for the provider to add the scopes as a claim:

```
options.Events.OnCreatingTicket = ctx =>
{
  var scope = ctx.HttpContext.Request.Query[OAuthChallengeProperties.ScopeKey];

  if (!StringValues.IsNullOrEmpty(scope))
  {
    ctx.Identity.AddClaim(new Claim("urn:strava:scope", scope));
  }

  return Task.CompletedTask;
};
```

Then I save that claim to the database in my external auth controller. Now when I go to use the access token I can check that the token has the correct scopes required by the API I'm calling. Perhaps if the aspnetcore issues get updated I can switch to that code. Thanks for your feedback.

nCubed:
@bryantlikes Appreciate the follow through with the multiple scopes and your workaround. I'd suggest another guard in the OnCreatingTicket to verify if the claim already exists, re: covers the case of the OAuth team adding in the missing scope values as a claim and/or Strava fixing their OAuth reply to be in compliance:

```csharp
private Task OnCreatingTicket(OAuthCreatingTicketContext context)
{
    string scopeKey = OAuthChallengeProperties.ScopeKey;

    if (!context.Identity.HasClaim(x => x.Type == scopeKey))
    {
        string scope = context.Request.Query[scopeKey];

        if (!string.IsNullOrWhiteSpace(scope))
        {
            context.Identity.AddClaim(new Claim($"urn:strava:{scopeKey}", scope));
        }
    }

    return Task.CompletedTask;
}

```

PR #446, opened 2020-07-03
Update to the Patreon v2 API

### LINKED ISSUE #433: Update to Patreon v2 API
As raised in #432, Patreon now has a [v2 API in beta](https://docs.patreon.com/#what-39-s-new).

Once it is stable, we should update the Patreon provider to use the new v2 API.

martincostello:
Updates the Patreon provider to use the v2 API.

Resolves #433.

The only thing I'm not sure about is whether this would be part of 3.1.2 as a "fix", or would be a follow-up 3.2.0 as it adds new functionality.

### DIFF
--- a/src/AspNet.Security.OAuth.Patreon/PatreonAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Patreon/PatreonAuthenticationHandler.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.OAuth;
+using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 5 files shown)


martincostello:
@kevinchalet What's your opinion on whether this should be part of 3.1.2, or should wait until a 3.2.0?

src/AspNet.Security.OAuth.Patreon/PatreonAuthenticationHandler.cs
```diff
@@ -34,7 +35,19 @@ public class PatreonAuthenticationHandler : OAuthHandler<PatreonAuthenticationOp
             [NotNull] AuthenticationProperties properties,
             [NotNull] OAuthTokenResponse tokens)
         {
-            using var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+            string endpoint = Options.UserInformationEndpoint;
+
+            if (Options.Fields?.Count > 0)
```
kevinchalet: Looks like these two collections have no setter and are always non-null, so the `?` is probably not necessary.

kevinchalet (commented):
> @kevinchalet What's your opinion on whether this should be part of 3.1.2, or should wait until a 3.2.0?

The v1 API is still working, right? If you plan to release the Auth0 provider as part of 3.2.0, then maybe it can wait until then?

src/AspNet.Security.OAuth.Patreon/PatreonAuthenticationHandler.cs
martincostello: Good spot! Will amend.

martincostello:
> The v1 API is still working, right?

It's still there, but it'll go away. I'm just not sure when.

>  If you plan to release the Auth0 provider

Auth0? Or did you mean the Okta one I added earlier? If so, I was planning on making that part of 3.1.2 as we've added new providers in patch releases before.

kevinchalet:
> Auth0? Or did you mean the Okta one I added earlier?

lol, yeah, Okta (I should really try to sleep... 🤣)

> If so, I was planning on making that part of 3.1.2 as we've added new providers in patch releases before.

Oh, then feel free to ship this change as part of 3.1.2.

PR #447, opened 2020-07-03
Only push packages from aspnet-contrib

martincostello:
Only attempt to push packages from aspnet-contrib, otherwise forks' dev branch will try to push to MyGet (and then fail).

This is something I missed on my own repos as they're not forks 😄


kevinchalet:
Hahaha, nice catch 😄

PR #448, opened 2020-07-04
Move build Id logic to workflow

martincostello:
Move the logic for calculating the build Id into the GitHub Actions workflow so the same Id is used consistently during a build, rather than changing if a build occurs across a 5 minute build boundary (https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/423#discussion_r449557811).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,23 @@ jobs:
     - name: Setup .NET Core SDK
       uses: actions/setup-dotnet@v1
 
+    # Arcade only allows the revision to contain up to two characters, and GitHub Actions does not roll-over
+    # build numbers every day like Azure DevOps does. To balance these two requirements, set the official
+    # build ID to be the same format as the built-in default from Arcade, except with the revision number
+    # being the number of the quarter hour of the current time of day (24 * 4 = 96, which is less than 100).
+    # So a build between 00:00 and 00:14 would have a revision of 1, and a build between 23:45 and 23:59:59
+    # would have a revision of 97.
+    - name: Set Build ID
+      if: ${{ startsWith(github.ref, 'refs/pull/') == false }}
+      shell: pwsh
+      run: |
+        $Now = (Get-Date).ToUniversalTime()
+        $Hours = $Now.Hour * 4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -32,6 +32,23 @@ jobs:
...
+    - name: Set Build ID
+      if: ${{ startsWith(github.ref, 'refs/pull/') == false }}
+      shell: pwsh
+      run: |
+        $Now = (Get-Date).ToUniversalTime()
+        $Hours = $Now.Hour * 4
+        $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
+        $Revision = $Hours + $QuarterHours + 1
+        $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
+        Write-Host "::set-env name=_AspNetContribBuildNumber::${BuildId}"
```
kevinchalet, line 50: Very nice! Could we remove the entire `PropertyGroup` block in `Directory.Build.props` if we directly set `OfficialBuild` and `OfficialBuildId` in this step?

kevinchalet (commented):
🎉

Directory.Build.props
```diff
@@ -23,22 +23,12 @@
   </PropertyGroup>
 
   <!--
-    Arcade only allows the revision to contain up to two characters, and GitHub Actions does not roll-over
-    build numbers every day like Azure DevOps does. To balance these two requirements, set the official
-    build ID to be the same format as the built-in default from Arcade, except with the revision number
-    being the number of the quarter hour of the current time of day (24 * 4 = 96, which is less than 100).
-    So a build between 00:00 and 00:14 would have a revision of 1, and a build between 23:45 and 23:59:59
-    would have a revision of 97.
+    These are set per-project so versioning is applied correctly, but are not set globally otherwise
```
kevinchalet, line 26: Oh? Interesting. Well, I guess that answers my question 😄

martincostello: Yeah, that's what I tried first but it go 💥 😅

https://github.com/martincostello/AspNet.Security.OAuth.Providers/runs/836762272?check_suite_focus=true#step:5:301

PR #449, opened 2020-07-04
Set DotNetPublishUsingPipelines to true and remove the workaround from Directory.Build.props

kevinchalet:
/cc @martincostello

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,15 +47,16 @@ jobs:
         $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
         $Revision = $Hours + $QuarterHours + 1
         $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
-        Write-Host "::set-env name=_AspNetContribBuildNumber::${BuildId}"
+        Write-Host "::set-env name=OfficialBuild::true"
+        Write-Host "::set-env name=OfficialBuildId::${BuildId}"
 
     - name: Build, Test and Package
-      run: eng\common\CIBuild.cmd -configuration Release -prepareMachine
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -47,15 +47,16 @@ jobs:
         $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
         $Revision = $Hours + $QuarterHours + 1
         $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
-        Write-Host "::set-env name=_AspNetContribBuildNumber::${BuildId}"
+        Write-Host "::set-env name=OfficialBuild::true"
+        Write-Host "::set-env name=OfficialBuildId::${BuildId}"
 
     - name: Build, Test and Package
-      run: eng\common\CIBuild.cmd -configuration Release -prepareMachine
+      run: eng\common\CIBuild.cmd -configuration Release -prepareMachine /p:DotNetPublishUsingPipelines=true
```
martincostello, line 54: Is this the _"I'm compiling in Open Source mode, not internal-with-all-the-security-process mode"_ switch?

kevinchalet: It's more like the "_I don't want Arcade to push the symbols (and likely some other stuff) as part of the build task to take care of that manually later in the pipeline in a dedicated publishing task_" mode 🤣 

I've tried a few other options, like adding a `Publishing.props` file with `PublishToSymbolServer` set to `false` or clearing the `PackagesToPublishToSymbolServer` or `FilesToPublishToSymbolServer` item groups, but `DotNetPublishUsingPipelines` is the only thing that works... 😭

kevinchalet: Anyway, thanks a lot for the time you spent on the build identifier computation script, it will make my life much easier. Merging PRs in the OpenIddict repo had become a total nightmare 😄

PR #417, opened 2020-05-04
Add SuperOffice provider

AnthonyYates:
Add SuperOffice provider and unit tests.
Update README.md

PR #452, opened 2020-07-07
Always publish logs and test results

martincostello:
Always try to publish build logs and test results, even if the build itself fails (https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/417#issuecomment-653909174).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,6 +60,7 @@ jobs:
 
     - name: Publish logs
       uses: actions/upload-artifact@v1
+      if: ${{ always() }}
       with:
         name: logs-${{ matrix.os_name }}
         path: ./artifacts/log/Release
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -60,6 +60,7 @@ jobs:
 
     - name: Publish logs
       uses: actions/upload-artifact@v1
+      if: ${{ always() }}
```
kevinchalet, line 63: Interesting. What does that mean for the actions that execute after a step decorated with `if: ${{ always() }}`? Do they execute even if the first step - but not the previous one - failed?

martincostello: It just applies to the specific step you decorate. The default if you don't specify anything is that it calls `success()`: [docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions)

kevinchalet: Very nice! Thanks 👏

PR #476, opened 2020-09-24
Use <DebugType>portable</DebugType> to avoid embedding the PDBs in the assemblies

### LINKED ISSUE #475: Determine whether we should stop using embedded PDBs
@martincostello all the aspnet-contrib OAuth 2.0/OpenID 2.0 providers currently ship with the PDBs embedded in the assemblies themselves, which makes deployments a bit bigger (something like +~10%, AFAICT).
That's because Arcade defaults to `embedded` PDBs when `OfficialBuild` is not set (and it's not something we use, as we hit issues with the automatic symbols upload to MSFT that requires authentication).

Instead, should we use `portable` and the new `snupkg` format? If we shipped our PDBs as `snupkg` files, we could push them to NuGet.org.

kevinchalet:
Fixes https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/475.

@martincostello should I target the `rel/5.0.0` branch for this PR?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,7 @@ jobs:
         path: ./artifacts/TestResults/Release
 
     - name: Push NuGet packages to aspnet-contrib MyGet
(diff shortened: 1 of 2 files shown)


martincostello (approved):
Either can go to dev and I'll make sure to incorporate into 5.0 for RC2, or can go straight into 5.0 if you're not worried about back-porting.

kevinchalet:
First option it will be, then 😄

PR #479, opened 2020-10-01
Kloudless provider

kYann:
Add Kloudless provider


samples/Mvc.Client/Startup.cs
```diff
@@ -60,7 +60,14 @@ public void ConfigureServices(IServiceCollection services)
                 options.EnterpriseDomain = Configuration["GitHub:EnterpriseDomain"];
                 options.Scope.Add("user:email");
             })
-
+            /*
+            .AddKloudless(options =>
```
martincostello: We don't generally add new ones anymore (unless they're unusual or very popular) as with 60+ providers now if we included them this file would be huge 😄

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods to add Kloudless authentication capabilities to an HTTP application pipeline.
+    /// </summary>
+    public static class KloudlessAuthenticationExtensions
+    {
+        /// <summary>
+        /// Adds <see cref="KloudlessAuthenticationHandler"/> to the specified
+        /// <see cref="AuthenticationBuilder"/>, which enables Imgur authentication capabilities.
```
martincostello:
```suggestion
        /// <see cref="AuthenticationBuilder"/>, which enables Kloudless authentication capabilities.
```

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                throw new HttpRequestException("An error occurred while retrieving the user profile.");
+            }
+
+            using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
+
+            var principal = new ClaimsPrincipal(identity);
+            var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
+            var accounts = payload.RootElement.GetProperty("objects");
+            context.RunClaimActions(accounts[0]);
```
martincostello, line 71: Can you harden this so it would throw a more informative error if the `objects` key weren't found, or if it didn't contain at least 1 item?

This is an array? If so it would be better to have the code explicitly deal with the JSON as an array, rather than just using the index on a `JsonValue` for the property.

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+        public KloudlessAuthenticationOptions()
+        {
+            ClaimsIssuer = KloudlessAuthenticationDefaults.Issuer;
+            CallbackPath = KloudlessAuthenticationDefaults.CallbackPath;
+
+            AuthorizationEndpoint = KloudlessAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = KloudlessAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = KloudlessAuthenticationDefaults.UserInformationEndpoint;
+
+            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "id");
```
martincostello, line 28: Just checking, there's no email-related claim here - is that correct, or an oversight?

test/AspNet.Security.OAuth.Providers.Tests/AspNet.Security.OAuth.Providers.Tests.csproj
```diff
@@ -17,6 +17,10 @@
     <ProjectReference Include="..\..\src\AspNet.Security.OAuth.*\*.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="Kloudless\bundle.json" />
+  </ItemGroup>
```
martincostello: This can be deleted as it's covered by a wildcard in the file already.

test/AspNet.Security.OAuth.Providers.Tests/Kloudless/KloudlessTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public override string DefaultScheme => KloudlessAuthenticationDefaults.AuthenticationScheme;
+
+        protected internal override void RegisterAuthentication(AuthenticationBuilder builder)
+        {
+            builder.AddKloudless(options => ConfigureDefaults(builder, options));
+        }
+
+        [Theory]
+        [InlineData(ClaimTypes.NameIdentifier, "my-id")]
```
martincostello, line 31: Could you extend the test to cover all of the custom claim mappings please?

martincostello (commented):
Thanks for your PR Yann - just a few comments otherwise this looks good 👍

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationOptions.cs
kYann: As value are provided by multi different provider, email address is not guaranteed.

martincostello: But if it is provided it won't be mapped. If there's a consistent claim name for it if it is present, then it would be automatically present rather than needing the user to do it explicitly?

samples/Mvc.Client/Startup.cs
kYann: I wanted to add an example, because if you don't provide a scope it won't work and there is no "default scope" as it really depends on what you want to use in kloudless.
Is there a better way to provide an example ?

test/AspNet.Security.OAuth.Providers.Tests/Kloudless/KloudlessTests.cs
kYann: I've added the tests for as much mapping as I could, but as `AuthenticateUserAsync` returns a Dictionary, I can't add mapping test for array (apis property)

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationOptions.cs
kYann: Sorry I was not clear enough. Field "account" may contain an email or a different type of "name" depending of the part and/or provider that you use on kloudless.
So I mapped it as "Name".

samples/Mvc.Client/Startup.cs
martincostello: You could add a Kloudless-specific document here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev/docs

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -54,7 +55,19 @@ public class KloudlessAuthenticationHandler : OAuthHandler<KloudlessAuthenticati
 
             var principal = new ClaimsPrincipal(identity);
             var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
-            var accounts = payload.RootElement.GetProperty("objects");
+
+            var isObjectsExists = payload.RootElement.TryGetProperty("objects", out var accounts);
+            if (!isObjectsExists)
+            {
+                throw new ArgumentException("Property objects is missing on accounts JSON");
```
martincostello: This isn't really an argument exception. If there isn't a better candidate, suggest using `InvalidOperationException`.

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -54,7 +55,19 @@ public class KloudlessAuthenticationHandler : OAuthHandler<KloudlessAuthenticati
...
+            }
+
+            var hasAtLeastOneAccount = accounts.GetArrayLength() > 0;
+            if (!hasAtLeastOneAccount)
+            {
+                throw new ArgumentException("Accounts JSON does not contains any account");
```
martincostello: Same here.

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -54,7 +55,19 @@ public class KloudlessAuthenticationHandler : OAuthHandler<KloudlessAuthenticati
 
             var principal = new ClaimsPrincipal(identity);
             var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
-            var accounts = payload.RootElement.GetProperty("objects");
+
+            var isObjectsExists = payload.RootElement.TryGetProperty("objects", out var accounts);
```
martincostello: `doesObjectsExist`

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -56,16 +56,16 @@ public class KloudlessAuthenticationHandler : OAuthHandler<KloudlessAuthenticati
...
             {
-                throw new ArgumentException("Property objects is missing on accounts JSON");
+                throw new InvalidOperationException("Property objects is missing on accounts JSON");
             }
 
             var hasAtLeastOneAccount = accounts.GetArrayLength() > 0;
             if (!hasAtLeastOneAccount)
             {
-                throw new ArgumentException("Accounts JSON does not contains any account");
+                throw new InvalidOperationException("Accounts JSON does not contains any account");
```
martincostello:
```suggestion
                throw new InvalidOperationException("Accounts JSON does not contain any accounts.");
```

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationHandler.cs
```diff
@@ -56,16 +56,16 @@ public class KloudlessAuthenticationHandler : OAuthHandler<KloudlessAuthenticati
             var principal = new ClaimsPrincipal(identity);
             var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
 
-            var isObjectsExists = payload.RootElement.TryGetProperty("objects", out var accounts);
-            if (!isObjectsExists)
+            var doesObjectsExist = payload.RootElement.TryGetProperty("objects", out var accounts);
+            if (!doesObjectsExist)
             {
-                throw new ArgumentException("Property objects is missing on accounts JSON");
+                throw new InvalidOperationException("Property objects is missing on accounts JSON");
```
martincostello:
```suggestion
                throw new InvalidOperationException("Property objects is missing on accounts JSON.");
```

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationConstants.cs
```diff
@@ -30,5 +30,17 @@ public static class Claims
...
             public const string Active = "urn:kloudless:active";
         }
+
+        public static class Scopes
+        {
+            public const string Any = "any";
+            public const string Storage = "storage";
+            public const string Calendar = "calendar";
+            public const string Email = "email";
+            public const string CRM = "crm";
```
martincostello:
```suggestion
            public const string Crm = "crm";
```

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationConstants.cs
```diff
@@ -30,5 +30,17 @@ public static class Claims
...
+            public const string Messaging = "messaging";
+            public const string ITSM = "itsm";
```
martincostello:
```suggestion
            public const string Itsm = "itsm";
```

martincostello (commented):
Two minor naming changes but otherwise this looks good to go! 🥳

samples/Mvc.Client/Startup.cs
kYann: Added documentation and set default scope (like it's done in Deezer provider)

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationConstants.cs
kYann: This is an acronym, that's why I'm using all uppercase here

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationConstants.cs
kYann: This is an acronym, that's why I'm using all uppercase here

src/AspNet.Security.OAuth.Kloudless/KloudlessAuthenticationConstants.cs
martincostello: .NET convention is to use first-letter capital and the rest lowercase ([docs](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1709?view=vs-2019)) for acronyms of 3 letters in length or more.

> By convention, two-letter acronyms use all uppercase letters, and acronyms of three or more characters use Pascal casing. The following examples use this naming convention: 'DB', 'CR', 'Cpa', and 'Ecma'. The following examples violate the convention: 'Io', 'XML', and 'DoD', and for non-parameter names, 'xp' and 'cpl'.

martincostello (approved):

martincostello:
Thanks @kYann! I'll merge and release this later on today as part of the 3.1.4 release. 🍰

martincostello:
This package is now available from [NuGet.org](https://www.nuget.org/packages/AspNet.Security.OAuth.Kloudless/3.1.4) - thanks again!

PR #484, opened 2020-10-11
Use IMemoryCache to cache client secrets for Apple provider

### LINKED ISSUE #396: Replace Sign in with Apple DefaultAppleClientSecretGenerator's `_clientSecret` with `IMemoryCache`

martincostello:
* Use `IMemoryCache` to cache client secrets for the Apple provider so they can be sharded by different client/key/team IDs.
  * Remove the configuration of `JwtSecurityTokenHandler` from the DI container and instead make it a property on `AppleAuthenticationOptions`.
  * Rather than using an existing `CryptoProviderFactory` if one is registered with DI, which could potentially create a conflict between other usage and ours, explicitly use a custom implementation that ensures the cache is disabled.
  * Fixes a log message that was not using structured logging.

Replaces #397, resolves #396.


src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
```diff
@@ -76,15 +74,14 @@ public static AuthenticationBuilder AddApple([NotNull] this AuthenticationBuilde
             [CanBeNull] string caption,
             [NotNull] Action<AppleAuthenticationOptions> configuration)
         {
+            builder.Services.AddMemoryCache();
             builder.Services.TryAddSingleton<AppleClientSecretGenerator, DefaultAppleClientSecretGenerator>();
```
kevinchalet: Should we make all these services options while we're at it? (to allow folks have different implementations per instance)

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
```diff
@@ -76,15 +74,14 @@ public static AuthenticationBuilder AddApple([NotNull] this AuthenticationBuilde
...
             builder.Services.TryAddSingleton<AppleIdTokenValidator, DefaultAppleIdTokenValidator>();
             builder.Services.TryAddSingleton<AppleKeyStore, DefaultAppleKeyStore>();
-            builder.Services.TryAddSingleton<JwtSecurityTokenHandler>();
 
             // Use a custom CryptoProviderFactory so that keys are not cached and then disposed of, see below:
             // https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1302
-            builder.Services.TryAddSingleton(
-                (_) => new CryptoProviderFactory() { CacheSignatureProviders = false });
+            builder.Services.AddSingleton<CustomCryptoProviderFactory>();
```
kevinchalet: I'd use an option for this one too. This way, you wouldn't need a custom derived type.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
```diff
@@ -144,7 +138,8 @@ protected virtual IEnumerable<Claim> ExtractClaimsFromToken([NotNull] string tok
         {
             try
             {
-                var securityToken = _tokenHandler.ReadJwtToken(token);
+                var tokenHandler = Options.JwtSecurityTokenHandler ?? new System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler();
```
kevinchalet: Is it worth having a null check here?

Also, it might be the right time to move to the newer `JsonWebTokenHandler` stack.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
```diff
@@ -86,7 +104,8 @@ public override async Task<string> GenerateAsync([NotNull] AppleGenerateClientSe
             {
                 tokenDescriptor.SigningCredentials = CreateSigningCredentials(context.Options.KeyId!, algorithm);
 
-                clientSecret = _tokenHandler.CreateEncodedJwt(tokenDescriptor);
+                var tokenHandler = context.Options.JwtSecurityTokenHandler ?? new JwtSecurityTokenHandler();
```
kevinchalet: I wouldn't care much about this option being null. We can still enforce it at validation time if we want to be 100% sure.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
```diff
@@ -18,24 +18,23 @@ internal sealed class DefaultAppleIdTokenValidator : AppleIdTokenValidator
...
             _keyStore = keyStore;
-            _tokenHandler = tokenHandler;
             _cryptoProviderFactory = cryptoProviderFactory;
             _logger = logger;
         }
 
         public override async Task ValidateAsync([NotNull] AppleValidateIdTokenContext context)
         {
-            if (!_tokenHandler.CanValidateToken)
+            var tokenHandler = context.Options.JwtSecurityTokenHandler ?? new JwtSecurityTokenHandler();
```
kevinchalet: Same here.

kevinchalet (commented):
🎉

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
martincostello: Implementers can already do that by overriding the DI registrations before calling `AddApple()` because the `TryAdd...()` methods are used.

Making them options would make it more explicit to override what is in DI, but we'd still need to register them with DI to create the defaults (because they need the clock, a logger etc.).

Given no-one has asked for this functionality in the last year, I don't think we need to further bloat the options with another way of doing something that's already possible.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
martincostello: I consider this an internal implementation detail to work around the caching behavior they changed which caused the bug it's working around in the first place, so I don't think we should expose it.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
martincostello: I'm not a big fan of making the code assume validation elsewhere too much, as if it's wrong/changed later then you get harder-to-diagnose `NullReferenceException`s instead of it either "just working", or giving you a more useful error.

I was just changing it locally to assign the default as a `PostConfigure()` and remove the null checks, but that then broke some of the unit tests that don't go through the full OAuth pipeline as the post-configure code doesn't run, so they just throw `NullReferenceException` rather than work, which sort of fits with why I'm not a fan of not-checking it in the first place.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
martincostello: As above, taking this out locally breaks some tests because the validation code doesn't always run unless it's a full in-the-HTTP-pipeline end-to-end test.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
kevinchalet: > Implementers can already do that by overriding the DI registrations before calling AddApple() because the TryAdd...() methods are used.

Not really. Services are shared between all schemes/instances, so you can't have one `AppleClientSecretGenerator` per Apple authentication handler instance (e.g `scheme1` and `scheme2`).

> Making them options would make it more explicit to override what is in DI, but we'd still need to register them with DI to create the defaults (because they need the clock, a logger etc.).

No, you can instantiate the "services-as-options" in an `IPostConfigureOptions` class, which is a very classical pattern in ASP.NET Core (e.g https://github.com/dotnet/aspnetcore/blob/master/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectPostConfigureOptions.cs#L35).

That said, if you think it's not worth it, that's fine.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
kevinchalet: Fine (that said, you can also have internal members in your options classes 😄)

martincostello: I did think of doing that, but then I thought it was a bit bleurgh. If you're happy with that, then I'll do that instead.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
kevinchalet: Ignoring the JWT handler from the options if it cannot be found and falling back to a default instance here - which is hard to detect - seems weird to me. Why not throwing an exception instead?

Regarding the tests, it's not clear to me why you'd get a null-ref: the property is always assigned a default value using the `{ get; set; } = new ...` syntax.

martincostello: Not in the commit I pushed just now where I copied the `= default!;` pattern from the AspNetCore repo which you hadn't seen where it's initialised by a `PostConfigure()` call.

But I fixed it by just explicitly setting the JwtSecurityTokenHandler property in the 3 tests it affected.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
martincostello: Yeah I was referencing [the OAuth version](https://github.com/dotnet/aspnetcore/blob/ab973d23fb5fc3b00d37ba63182b5bb14dc02803/src/Security/Authentication/OAuth/src/OAuthPostConfigureOptions.cs#L33) where you can provide a data protection provider, but it falls back to the one in DI (not doing a `new ...()` where you need to get hold of all of _those_ things' dependencies too).

martincostello: This is now all handled in `ApplePostConfigureOptions`.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationExtensions.cs
martincostello: This is still internal only, but now handled in `ApplePostConfigureOptions` rather than DI.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
martincostello: I refactored the tests so that the `PostConfigure()` got called which fixed the NREs.

I looked at changing over to `JsonWebTokenHandler`, but that broke all the tests as there's no longer a hook the tests can use to override the lifetime validation of the JWT without adding additional functionality to allow the application to configure the token validation parameters.

I'll leave that for now as it's a bigger change and unrelated to the original issue, plus would need the SuperOffice provider to change as well as that uses the same handler for JWTs.

src/AspNet.Security.OAuth.Apple/AppleAuthenticationEvents.cs
```diff
@@ -22,17 +21,15 @@ public class AppleAuthenticationEvents : OAuthEvents
         /// </summary>
         public Func<AppleGenerateClientSecretContext, Task> OnGenerateClientSecret { get; set; } = async context =>
         {
-            var provider = context.HttpContext!.RequestServices!.GetRequiredService<AppleClientSecretGenerator>();
-            context.Options.ClientSecret = await provider.GenerateAsync(context);
+            context.Options.ClientSecret = await context.Options.ClientSecretGenerator.GenerateAsync(context);
```
kevinchalet, line 24: It's probably off-topic for this PR, but we shouldn't mutate options during request handling. Maybe we should add a `ClientSecret` property to `AppleGenerateClientSecretContext` instead of mutating the one in the options.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
```diff
@@ -18,24 +18,21 @@ internal sealed class DefaultAppleIdTokenValidator : AppleIdTokenValidator
...
             _keyStore = keyStore;
-            _tokenHandler = tokenHandler;
             _cryptoProviderFactory = cryptoProviderFactory;
             _logger = logger;
         }
 
         public override async Task ValidateAsync([NotNull] AppleValidateIdTokenContext context)
         {
-            if (!_tokenHandler.CanValidateToken)
+            if (!context.Options.JwtSecurityTokenHandler!.CanValidateToken)
```
kevinchalet: Is the null-forgiving operator necessary here?

kevinchalet (approved):

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
martincostello: Won't compile without it because we're no longer checking for null.

kevinchalet: But the `JwtSecurityTokenHandler` is not "nullable", right? It shouldn't be needed here (if it is, it may be another Roslyn bug).

martincostello: Looks like that bug was fixed in RC2.

PR #488, opened 2020-10-19
Email is a mandatory claim in yahoo.

farajfarook:
Email is a mandatory claim. Add it by default to claims.


martincostello:
Thanks for the PR @farajfarook - could you add a test case for the new claim please?

Should be as simple as adding a new `[InlineData(...)]` here:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/33a57e3da11b9c72d20b697f75187a58251cbcb1/test/AspNet.Security.OAuth.Providers.Tests/Yahoo/YahooTests.cs#L30-L36

The email will need adding to the test JSON payload here:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/33a57e3da11b9c72d20b697f75187a58251cbcb1/test/AspNet.Security.OAuth.Providers.Tests/Yahoo/bundle.json#L18-L24

martincostello:
This change is now available from [NuGet.org](https://www.nuget.org/packages/AspNet.Security.OAuth.Yahoo/3.1.5).

farajfarook:
@martincostello I saw you have already added the tests. Thank you very much.  🥇

PR #490, opened 2020-10-19
Update staging subdomain

### LINKED ISSUE #489: AspNet.Security.OAuth.Providers.SuperOffice package expects incorrect issuer in "stage"-enviroment
**Describe the bug**

When using the package `AspNet.Security.OAuth.Providers.SuperOffice`, you are expected to populate the relevant SuperOffice environment to authenticate for (development, stage or production). The provider is then configured with appropriate settings and URLs for that environment.

However, for the stage environment, the package does not work, because the issuer-value which the provider expects is not the actual issuer-value used in this environment.

**Steps To reproduce**

Create a minimal Asp .Net Core application and configure it to use SuperOffice authentication in the SuperOffice stage-environment. Launch app, attempt to authenticate.

````
$ dotnet new web
$ dotnet add package AspNet.Security.OAuth.SuperOffice --version 3.1.4
$ cat Startup.cs
...
  public void ConfigureServices(IServiceCollection services)
  {
      services.AddAuthentication()
              .AddSuperOffice(SuperOfficeAuthenticationDefaults.AuthenticationScheme, options =>
      {
          options.Environment = SuperOfficeAuthenticationEnvironment.Stage;
          // etc etc
      });
  }
...
$ dotnet run
````

**Expected behaviour**

Expected website to use SuperOffice stage environment to provide authenticated user.

**Actual behaviour**

Unable to authenticate user because package expects `"stage.superoffice.com"` as issuer of claims-identity, while real issuer used is `"qaonline.superoffice.com"`.

This can be confirmed independently in the well-known openid-configuration document found here:

https://stage.superoffice.com/login/.well-known/openid-configuration

image

**System information:**

 - OS: N/A
 - Library Version: 3.1.4
 - .NET version: 3.1.201

**Additional context**

This breaks the release/deployment flow of any application which attempts to follow the DEV, STAGE, PROD deployment flow with the SuperOffice authentication-package, so it would be nice to get it fixed before .NET 5.0 RTM ships.

AnthonyYates:
Updated staging subdomain, which was cause problems validating id_token due to different issue.

Issue 489
https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/489

### DIFF
--- a/src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationOptions.cs
@@ -138,7 +138,7 @@ namespace AspNet.Security.OAuth.SuperOffice
             return _environment switch
             {
                 SuperOfficeAuthenticationEnvironment.Development => "sod",
-                SuperOfficeAuthenticationEnvironment.Stage => "stage",
+                SuperOfficeAuthenticationEnvironment.Stage => "qaonline",
                 SuperOfficeAuthenticationEnvironment.Production => "online",
                 _ => throw new NotSupportedException("Environment property must be set to either Development, Stage or Production.")
             };


martincostello (approved):

martincostello:
Thanks for the PR @AnthonyYates - I'll look at pushing out a 3.1.5 version with the fix either today or tomorrow.

PR #500, opened 2020-11-09
Use new syntax to set environment variable

martincostello:
Use new syntax to set environment variable value.

See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
         $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
         $Revision = $Hours + $QuarterHours + 1
         $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
-        Write-Host "_AspNetContribBuildNumber=${BuildId}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        Write-Output "_AspNetContribBuildNumber=${BuildId}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
     - name: Build, Test and Package
       if: ${{ runner.os == 'Windows' }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
         $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
         $Revision = $Hours + $QuarterHours + 1
         $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
-        Write-Host "::set-env name=_AspNetContribBuildNumber::${BuildId}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
         $Revision = $Hours + $QuarterHours + 1
         $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
-        Write-Host "::set-env name=_AspNetContribBuildNumber::${BuildId}"
+        Write-Host "_AspNetContribBuildNumber=${BuildId}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```
kevinchalet: I'm definitely not a PowerShell expert so I may be wrong, but doesn't `Write-Host` directly write to the console output?
Instead, shouldn't we do something like:

```
"_AspNetContribBuildNumber=${BuildId}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```

Or:

```
Add-Content -Path $env:GITHUB_PATH -Encoding utf8 -Value "_AspNetContribBuildNumber=${BuildId}"
```

martincostello: ¯\_(ツ)_/¯

I just copied this and replaced the `echo` alias: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

kevinchalet: I suspect the two syntaxes don't behave exactly the same way in this regard:

image

martincostello: Looks like I just mixed up what `echo` was the alias for: https://stackoverflow.com/a/17623887/1064169

martincostello:
Hmm...

I'm not sure that actually worked correctly. The revision is coming out as `1` even though the code that calculates the number hasn't changed and it seeme to work as expected if I run it locally:

```powershell
C:\Coding> $Now = (Get-Date).ToUniversalTime()
C:\Coding> $Hours = $Now.Hour * 4
C:\Coding> $QuarterHours = [Math]::Floor($Now.Minute / 15.0)
C:\Coding> $Revision = $Hours + $QuarterHours + 1
C:\Coding> $BuildId = $Now.ToString("yyyyMMdd") + "." + $Revision
C:\Coding> Write-Output "_AspNetContribBuildNumber=${BuildId}"
_AspNetContribBuildNumber=20201109.90
```

kevinchalet:
> I'm not sure that actually worked correctly.

Is that because you used `$env:GITHUB_PATH` instead of `$env:GITHUB_ENV`?

PR #518, opened 2021-02-02
Updated Lichess scopes to include bulk challenges

tanczosm:
Updates Lichess authentication scope constants to include recently added scope for "bulk challenges"

https://lichess.org/api#tag/Bulk-pairings


martincostello:
Can you rebase please? Looks like this PR includes loads of unrelated commits.

For future changes if you need to make them, it's better to create a branch, rather than work on the `dev` branch or you'll have to be constantly changing what your dev branch is as it will divert from ours.

PR #519, opened 2021-02-02
Added new Lichess bulk challenge OAuth scope

tanczosm:
Updates Lichess authentication scope constants to include recently added scope for "bulk challenges"

https://lichess.org/api#tag/Bulk-pairings


martincostello:
Thanks @tanczosm - as this is just adding a constant, this will just be released whenever we next do a release.

martincostello:
This change is now available in v5.0.2 of the provider, thanks again.

https://www.nuget.org/packages/AspNet.Security.OAuth.Lichess/5.0.2

PR #387, opened 2020-02-15
Add Alipay provider

vicenteyu:
- Add Alipay provider and unit test.
- Update README.md

PR #522, opened 2021-02-24
Add Alipay provider

### LINKED ISSUE #520: About Alipay Provider

I notice that the PR #387 has been closed. Alipay(of Alibaba) is a quite common external provider for Chinese users, which is similar to QQ or WeChat, so I think it’s better to add alipay provider here.

The difference between Alipay provider and others is that it uses an unified gateway endpoint with many self-defined request params and it currently provides a dedicated [SDK for .NET](https://github.com/alipay/alipay-easysdk/tree/master/csharp). So I want to make sure if I can rely on the SDK which is maintained actively by the official. 
The SDK provides two benefits:  make develop work simpler(for example, the way to get value of 'sign' by RSA2) and more maintainable for the future (when outh-apis change, you need to upgrade the SDK only).

Also, I'm able to give my own implementation according to the document of Alipay without the help of official SDK, but I'm not sure I have the time to keep it work in the future.

**Additional context**
Dingding Talk(another production of Alibaba) is similar to this situation, but I'm not interested about that because it's not very popular as an external provider.

LeaFrock:
Related to issue #520 .

Based on a part of the closed PR #387 .

### DIFF
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
@@ -1,0 +1,30 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+namespace AspNet.Security.OAuth.Alipay
+{
+    /// <summary>
+    /// Contains constants specific to the <see cref="AlipayAuthenticationHandler"/>.
+    /// </summary>
+    public static class AlipayAuthenticationConstants
+    {
+        public static class Claims
+        {
+            public const string Avatar = "urn:alipay:avatar";
+
+            public const string Province = "urn:alipay:province";
+
+            public const string City = "urn:alipay:city";
+
+            public const string Nickname = "urn:alipay:nick_name";
+
+            /// <summary>
+            /// The user's gender. F: Female; M: Male;
+            /// </summary>
+            public const string Gender = "urn:alipay:gender";
+        }
+    }
+}
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
@@ -22,7 +22,7 @@ namespace AspNet.Security.OAuth.Alipay
             public const string Nickname = "urn:alipay:nick_name";
 
             /// <summary>
-            /// The user's gender. F: Female; M: Male;
+            /// The user's gender. F: Female; M: Male.
             /// </summary>
             public const string Gender = "urn:alipay:gender";
         }
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
@@ -1,0 +1,227 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Claims;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.OAuth;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace AspNet.Security.OAuth.Alipay
+{
+    /// <summary>
+    /// Defines a handler for authentication using Alipay.
+    /// </summary>
+    public class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthenticationOptions>
+    {
+        public AlipayAuthenticationHandler(
+            [NotNull] IOptionsMonitor<AlipayAuthenticationOptions> options,
+            [NotNull] ILoggerFactory logger,
+            [NotNull] UrlEncoder encoder,
+            [NotNull] ISystemClock clock)
+            : base(options, logger, encoder, clock)
+        {
+        }
+
+        protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
+        {
+            var query = Request.Query;
+            if (query.TryGetValue("auth_code", out var authCode))
+            {
+                // The base `HandleRemoteAuthenticateAsync` requires that `Request.Query` must contain the key called `code`,
+                // which is actually the same as `auth_code` by Alipay's design, but `Request.Query` does not have `Add` operation.
+                // So here is a trick to get the private `Store` dictionary of `QueryCollection`.
+                var queryStore = query.ToDictionary(c => c.Key, c => c.Value, StringComparer.OrdinalIgnoreCase);
+                queryStore["code"] = authCode;
+                Request.QueryString = QueryString.Create(queryStore);
+            }
+
+            return base.HandleRemoteAuthenticateAsync();
+        }
+
+        protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
+        {
+            // See https://opendocs.alipay.com/apis/api_9/alipay.system.oauth.token for details.
+            var sortedParams = new SortedDictionary<string, string>()
+            {
+                ["app_id"] = Options.ClientId,
+                ["charset"] = "utf-8",
+                ["code"] = context.Code,
+                ["format"] = "JSON",
+                ["grant_type"] = "authorization_code",
+                ["method"] = "alipay.system.oauth.token",
+                ["sign_type"] = "RSA2",
+                ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
+                ["version"] = "1.0",
+            };
+            sortedParams.Add("sign", GetRSA2Signature(sortedParams, Options.GetPrivateKeyBytes()));
+
+            string address = QueryHelpers.AddQueryString(Options.TokenEndpoint, sortedParams!);
+
+            using var response = await Backchannel.GetAsync(address, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+            if (!response.IsSuccessStatusCode)
+            {
+                Logger.LogError("An error occurred while retrieving an access token: the remote server " +
+                                "returned a {Status} response with the following payload: {Headers} {Body}.",
+                                /* Status: */ response.StatusCode,
+                                /* Headers: */ response.Headers.ToString(),
+                                /* Body: */ await response.Content.ReadAsStringAsync());
+
+                return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+            }
+
+            using var stream = await response.Content.ReadAsStreamAsync(Context.RequestAborted);
+            using var document = JsonDocument.Parse(stream);
+
+            var mainElement = document.RootElement.GetProperty("alipay_system_oauth_token_response");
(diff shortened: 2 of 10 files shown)


src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -0,0 +1,30 @@
...
+            public const string Avatar = "urn:alipay:avatar";
+
+            public const string Province = "urn:alipay:province";
+
+            public const string City = "urn:alipay:city";
+
+            public const string Nickname = "urn:alipay:nick_name";
+
+            /// <summary>
+            /// The user's gender. F: Female; M: Male;
```
martincostello:
```suggestion
            /// The user's gender. F: Female; M: Male.
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+using static AspNet.Security.OAuth.Alipay.AlipayAuthenticationConstants;
+
+namespace AspNet.Security.OAuth.Alipay
+{
+    /// <summary>
+    /// Defines a set of options used by <see cref="AlipayAuthenticationHandler"/>.
+    /// </summary>
+    public class AlipayAuthenticationOptions : OAuthOptions
+    {
+        public byte[] GetPrivateKeyBytes() => Convert.FromBase64String(ClientSecret);
```
martincostello: I would just make this an implementation detail of the class that is using it, rather than putting this here.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+            var sortedParams = new SortedDictionary<string, string>()
+            {
+                ["app_id"] = Options.ClientId,
+                ["charset"] = "utf-8",
+                ["code"] = context.Code,
+                ["format"] = "JSON",
+                ["grant_type"] = "authorization_code",
+                ["method"] = "alipay.system.oauth.token",
+                ["sign_type"] = "RSA2",
+                ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
```
martincostello, line 70: Does Alipay support having `Z` on the end of the timestamp string so you can just use the `"u"` format string instead?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+                ["app_id"] = Options.ClientId,
+                ["auth_token"] = tokens.AccessToken,
+                ["charset"] = "utf-8",
+                ["format"] = "JSON",
+                ["method"] = "alipay.user.info.share",
+                ["sign_type"] = "RSA2",
+                ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
+                ["version"] = "1.0",
+            };
+            sortedParams.Add("sign", GetRSA2Signature(sortedParams, Options.GetPrivateKeyBytes()));
```
martincostello: You always use the same key, so might as well fold it into the `GetRSA2Signature()` method.

```suggestion
            sortedParams.Add("sign", GetRSA2Signature(sortedParams));
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+            var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, mainElement);
+
+            context.RunClaimActions();
+
+            await Options.Events.CreatingTicket(context);
+
+            return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
+        }
+
+        protected override string FormatScope() => FormatScope(Options.Scope);
```
martincostello: You don't need this override, it's [already what the base class does](https://github.com/dotnet/aspnetcore/blob/50c635f95f0fcbe30f30175d2884f79e693f48a7/src/Security/Authentication/OAuth/src/OAuthHandler.cs#L339-L340).

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+        protected override string FormatScope(IEnumerable<string> scopes) => string.Join(',', Options.Scope);
+
+        /// <summary>
+        /// Check the code sent back by server for potential server errors.
+        /// </summary>
+        /// <param name="e">Main part of json document from response</param>
+        /// <param name="code">Returned code from server</param>
+        /// <remarks>See https://opendocs.alipay.com/open/common/105806 for details.</remarks>
+        /// <returns>True if succeed, otherwise false.</returns>
+        private static bool ValidateReturnCode(JsonElement e, out string code)
```
martincostello: Rename `e` to `element`.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        /// <summary>
+        /// Gets the RSA2(SHA256 with RSA) signature.
+        /// </summary>
+        /// <param name="sortedPairs">Sorted key-value pairs</param>
+        /// <param name="privateKeyBytes">Private key</param>
+        private static string GetRSA2Signature([NotNull] SortedDictionary<string, string> sortedPairs, byte[] privateKeyBytes)
+        {
+            var builder = new StringBuilder(128);
```
martincostello:
```suggestion
        private static string GetRSA2Signature([NotNull] SortedDictionary<string, string> sortedPairs)
        {
            byte[] privateKeyBytes = Convert.FromBase64String(Options.ClientSecret);
            var builder = new StringBuilder(128);
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+                builder.Append(pair.Key)
+                       .Append('=')
+                       .Append(pair.Value)
+                       .Append('&');
+            }
+
+            string plainText = builder.ToString();
+            byte[] plainBytes = Encoding.UTF8.GetBytes(plainText, 0, plainText.Length - 1); // Skip the last '&'
+
+            // Use ThreadLocal<RSA> instead?
```
martincostello: No, the RSA should be created and disposed like this.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,227 @@
...
+            using var rsa = RSA.Create();
+            try
+            {
+                rsa.ImportRSAPrivateKey(privateKeyBytes, out int _);
+            }
+            catch (Exception)
+            {
+                throw;
+            }
```
martincostello: The try-catch is redundant.

```suggestion
            rsa.ImportRSAPrivateKey(privateKeyBytes, out int _);
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: As the doc site says, the answer is 'no'.
image

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: Oh！It's careless of me.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: That‘s OK.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
LeaFrock: Done.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
LeaFrock: How about make it `internal`? In fact, I hope to optimize it if possible in the future. The problem is that `Convert.FromBase64String(ClientSecret)` each time may be a huge waste of perf because `ClientSecret` is rarely changed after once setted.
At the very original version, I wrote codes like below,
``` C#

private byte[] _privateKeyBytes;

public byte[] PrivateKeyBytes()
{
    if(_privateKeyBytes is null)
    {
        _privateKeyBytes = Convert.FromBase64String(ClientSecret);
    }
    return _privateKeyBytes;
}

```
However, `ClientSecret` is not a readonly property. So that solution is not ideal.
I thought putting this here would be a good placeholder. That's why I want to check with you.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: Well.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: Just confirm. If so, `GetRSA2Signature` will be non static due to `Options`.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: It's my mistake as there was a `log` operation here and I deleted it.
Now just throw exception by itself if `ClientSecret` is invalid.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
LeaFrock: Done.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
martincostello: The `FromBase64String()` overhead is going to be pretty trivial. Also, the providers support online configuration change (e.g. from a database, Azure Key Vault etc.) as it accepts an `IOptionsMonitor<T>` instance. If the key was cached like this and only computed once, such a configuration change wouldn't take effect and would require an application restart to be applied.

As such, compute it each time it's needed as suggested.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: Correct.

martincostello (commented):
This looks OK to me as-is, but I'll leave final approval to @kevinchalet as he had specific feedback about this provider in #520.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
LeaFrock: Fine.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -0,0 +1,217 @@
...
+                throw new HttpRequestException("An error occurred while retrieving user information.");
+            }
+
+            using var stream = await response.Content.ReadAsStreamAsync(Context.RequestAborted);
+            using var document = JsonDocument.Parse(stream);
+            var rootElement = document.RootElement;
+
+            if (!rootElement.TryGetProperty("alipay_user_info_share_response", out var mainElement))
+            {
+                string errorCode = rootElement.GetProperty("error_response").GetProperty("code").GetString() !;
```
kevinchalet, line 142: nit: please use `var` and remove unnecessary spaces before `!` (here and elsewhere)

kevinchalet (approved):
Thanks for your contribution, @LeaFrock!

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: (The `{space}!` is a StyleCop workaround)

kevinchalet: Crap... 😢 
Thanks for the info!

LeaFrock: Done.

PR #526, opened 2021-03-09
Added documentation on Sign In with Apple on iOS

### LINKED ISSUE #525: Apple Signin Documentation
There are two ways of authenticating an app with Apple, OAuth and the iOS "native" signin workflow. The first one is covered by the Apple OAuth Package, the second is not really addressed here.

An update to the OAuth2 Apple documentation or integration of code handling the iOS Apple Signin Workflow

I've implemented handling of the iOS Signin Workflow in my ASP.Net Core Program and could create a separate repository for that but I would say 80-90 % of the code would be the same than in the Apple OAuth Package (generating client secret, validating tokens)

**Additional context**

The sequence of native iOS Apple Signin workflow is as follows:

1. Receive Authorization Code within App
2. Send Authorization Code to own servers
3. Generate or retrieve cached client secret 
4. Send client secret, client id and authorization code to [Apple Endpoint](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens)
5. Retrieve idtoken, validate token and parse claims for user user information

The parts not covered by the Apple OAuth are sending the request with authorization code and extracting the claims but that would be a very small code addition, however it is true that is not really OAuth related. I am open to adding to the documentation or providing the code supporting the native signin workflow but since it might be out of scope I wanted to ask before creating a pull request. I created an issue #523 for this a few days ago as well

j-bbr:
Documentation update as discussed in #525

### DIFF
--- a/docs/sign-in-with-apple.md
+++ b/docs/sign-in-with-apple.md
@@ -61,6 +61,12 @@ Below are links to some issues raised against this repository that were related 
   * [Allow passing in private key as string instead of p8 file (#385)](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/385 "Allow passing in private key as string instead of p8 file")
   * [Apple Signin redirects to a blank page 404 Error (#390)](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/390 "Apple Signin redirects to a blank page 404 Error")
 
+## Sign in with Apple on iOS
+
+When using _Sign In with Apple_ on an iOS 13+ Device, [Apple provides a different authentication workflow](https://developer.apple.com/documentation/authenticationservices) that returns the validation response to the App instead of in a server callback. Using that response to authenticate a User against your own backend requires sending the response to your servers and [communicating with the Apple authentication endpoint from there](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens). 
+
+This workflow is out of the scope of this package but client secret generation and token validation can provide a starting point for an ASP.NET.Core Integration. Note that the `ClientId` in this case is the App Id where the authentication was requested not your Services Id.
+
 ## Further Reading
 
 Below are links to a number of other documentation sources, blog posts and sample code about Sign in with Apple.
--- a/docs/sign-in-with-apple.md
+++ b/docs/sign-in-with-apple.md
@@ -65,7 +65,7 @@ Below are links to some issues raised against this repository that were related 
 
 When using _Sign In with Apple_ on an iOS 13+ Device, [Apple provides a different authentication workflow](https://developer.apple.com/documentation/authenticationservices) that returns the validation response to the App instead of in a server callback. Using that response to authenticate a User against your own backend requires sending the response to your servers and [communicating with the Apple authentication endpoint from there](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens). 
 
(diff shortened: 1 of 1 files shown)


docs/sign-in-with-apple.md
```diff
@@ -61,6 +61,12 @@ Below are links to some issues raised against this repository that were related
...
   * [Apple Signin redirects to a blank page 404 Error (#390)](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/390 "Apple Signin redirects to a blank page 404 Error")
 
+## Sign in with Apple on iOS
+
+When using _Sign In with Apple_ on an iOS 13+ Device, [Apple provides a different authentication workflow](https://developer.apple.com/documentation/authenticationservices) that returns the validation response to the App instead of in a server callback. Using that response to authenticate a User against your own backend requires sending the response to your servers and [communicating with the Apple authentication endpoint from there](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens). 
+
+This workflow is out of the scope of this package but client secret generation and token validation can provide a starting point for an ASP.NET.Core Integration. Note that the `ClientId` in this case is the App Id where the authentication was requested not your Services Id.
```
martincostello:
```suggestion
This workflow is out of the scope of this package but client secret generation and token validation can provide a starting point for an ASP.NET.Core integration. Note that the `ClientId` in this case is the App Id where the authentication was requested, not your Services Id.
```

martincostello: Is _"Services Id"_ the correct term here? Just checking as I'm not familiar with the term.

docs/sign-in-with-apple.md
```diff
@@ -61,6 +61,12 @@ Below are links to some issues raised against this repository that were related
   * [Allow passing in private key as string instead of p8 file (#385)](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/385 "Allow passing in private key as string instead of p8 file")
   * [Apple Signin redirects to a blank page 404 Error (#390)](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/390 "Apple Signin redirects to a blank page 404 Error")
 
+## Sign in with Apple on iOS
+
+When using _Sign In with Apple_ on an iOS 13+ Device, [Apple provides a different authentication workflow](https://developer.apple.com/documentation/authenticationservices) that returns the validation response to the App instead of in a server callback. Using that response to authenticate a User against your own backend requires sending the response to your servers and [communicating with the Apple authentication endpoint from there](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens).
```
martincostello:
```suggestion
When using _Sign In with Apple_ on an iOS 13+ Device, [Apple provides a different authentication workflow](https://developer.apple.com/documentation/authenticationservices) that returns the validation response to the app instead of in a server callback. Using that response to authenticate a user against your own backend requires sending the response to your servers and [communicating with the Apple authentication endpoint from there](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens). 
```

docs/sign-in-with-apple.md
j-bbr: That's how it is referred to in the Apple Documentation:

> The identifier (App ID or Services ID) for your app.

from the same [Apple Documentation Page](https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens).

martincostello (approved):

PR #497, opened 2020-11-02
Add EVEOnlineV2 provider

Dusty-Meg:
Add provider for the V2 routes of the EVE Online SSO.

The V1 (current) routes are still valid, but they will eventually be phased out so having the V2 routes available will be useful. 
The V2 login pattern is the recommended route from the developers.

PR #540, opened 2021-04-23
updating the out-of-date EveOnline provider

seraphx2:
The Eve Online provider is woefully out of date. It no longer supports Singularity server and also has a v2 option for OAuth provider.
I have removed Singularity as an option and added in Serenity. Also added an enum option to support choosing between versions of Oauth.

Both v1 and v2 options have been tested and verified that the built-in Microsoft External Login Provider will acknowledge both versions of OAuth and it knows the user is the same, even switching between versions within the same project's authorization scheme.

Please consider updating this because whoever originally added this has not been keeping up with it and several changes have obviously been made by the SSO owners (CCP) and this would be most useful to the Eve Online community.


seraphx2:
OK. I guess I should have checked to see if someone was already working on this. My bad.

martincostello:
#497 has been stuck since February pending this comment https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/497#issuecomment-771649716 - are you able to shed any more light on things so we can maybe move forward with it?

PR #532, opened 2021-03-26
Added Moodle support

FerdinandSu:
Added Moodle support for plugin [projectestac/moodle-local_oauth](https://github.com/projectestac/moodle-local_oauth).

Details:

1. Added project **AspNet.Security.OAuth.Moodle**
2. Added docs _moodle.md_
3. Updated README.md

PR #543, opened 2021-05-09
added Keycloak provider

### LINKED ISSUE #498: Can we have AddKeycloak()?

antosubash:
this pull adds support for the keycloak which was asked here #498 

I have added the test but the keycloak is self-hosted and I'm not sure how to make the test work. please advise on how to fix the tests. I don't want to put some random instance and make it work and then it will break later.


martincostello:
Will aim to try and review this next weekend.

martincostello:
> I have added the test but the keycloak is self-hosted and I'm not sure how to make the test work. please advise on how to fix the tests. I don't want to put some random instance and make it work and then it will break later.

If you take a look at the Moodle provider (#532), you can see an example of testing a self-hosted provider. There are also similar examples for Okta and GitHub Enterprise where the domain of the service is not static.

src/AspNet.Security.OAuth.Keycloak/AspNet.Security.OAuth.Keycloak.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
martincostello: Our dev branch now targets .NET 5.0.

```suggestion
    <TargetFramework>net5.0</TargetFramework>
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            ClaimsIssuer = KeycloakAuthenticationDefaults.Issuer;
+            CallbackPath = KeycloakAuthenticationDefaults.CallbackPath;
+
+            AuthorizationEndpoint = KeycloakAuthenticationDefaults.AuthorizationEndpoint;
+            TokenEndpoint = KeycloakAuthenticationDefaults.TokenEndpoint;
+            UserInformationEndpoint = KeycloakAuthenticationDefaults.UserInformationEndpoint;
+            Scope.Add("openid");
+            Scope.Add("email");
+            Scope.Add("roles");
```
martincostello:
```suggestion

            Scope.Add("openid");
            Scope.Add("email");
            Scope.Add("roles");

```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+            ClaimActions.MapAll();
```
martincostello: Why do you need to call this?

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+            ClaimActions.MapJsonKey(ClaimTypes.Email, "email");
+            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "sub");
+            ClaimActions.MapJsonKey(ClaimTypes.Name, "name");
+            ClaimActions.MapJsonKey(ClaimTypes.GivenName, "given_name");
+            ClaimActions.MapJsonKey(ClaimTypes.Role, "roles");
+        }
+
+        public string BaseUrl { get; set; } = string.Empty;
```
martincostello: Please add `///` comments explaining what this property is for.

src/AspNet.Security.OAuth.Keycloak/KeycloakPostConfigureOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class KeycloakPostConfigureOptions : IPostConfigureOptions<KeycloakAuthenticationOptions>
+    {
+        public void PostConfigure([NotNull] string name, [NotNull] KeycloakAuthenticationOptions options)
+        {
+            if (!string.IsNullOrWhiteSpace(options.BaseUrl))
+            {
+                options.AuthorizationEndpoint = options.BaseUrl + KeycloakAuthenticationDefaults.AuthorizationEndpoint;
+                options.TokenEndpoint = options.BaseUrl + KeycloakAuthenticationDefaults.TokenEndpoint;
+                options.UserInformationEndpoint = options.BaseUrl + KeycloakAuthenticationDefaults.UserInformationEndpoint;
```
martincostello: Please use the same pattern the other post-configure implementations use to ensure that URLs are constructed correctly, rather than using simple string concatenation.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dad5906d49eea20886aa35fada099ad8c5043221/src/AspNet.Security.OAuth.GitHub/GitHubPostConfigureOptions.cs#L20-L44

test/AspNet.Security.OAuth.Providers.Tests/Keycloak/KeycloakTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public override string DefaultScheme => KeycloakAuthenticationDefaults.AuthenticationScheme;
+
+        protected internal override void RegisterAuthentication(AuthenticationBuilder builder)
+        {
+            builder.AddKeycloak(options => ConfigureDefaults(builder, options));
+        }
+
+        [Theory]
+        [InlineData(ClaimTypes.NameIdentifier, "my-id")]
```
martincostello: Please add a test case for each of the claims mapped by the default options.

test/AspNet.Security.OAuth.Providers.Tests/Keycloak/bundle.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://raw.githubusercontent.com/justeat/httpclient-interception/master/src/HttpClientInterception/Bundles/http-request-bundle-schema.json",
+  "items": [
+    {
+      "uri": "https://{domain}/v1/oauth/token",
```
martincostello: Use a `.local` domain with a fixed value like the Moodle, GitHub Enterprise and Okta tests do.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dad5906d49eea20886aa35fada099ad8c5043221/test/AspNet.Security.OAuth.Providers.Tests/Okta/bundle.json#L6

martincostello (commented):
Please add the new provider to the README and add a file to the `/docs` folder following the convention used by the existing files to document what the additional property in the Options class is for.

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
martincostello: Also given it's probably the domain of the self-hosted Keycloak server, for consistency please name it `Domain`. The post-options class would then turn it into a valid URI.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dad5906d49eea20886aa35fada099ad8c5043221/src/AspNet.Security.OAuth.Moodle/MoodleAuthenticationOptions.cs#L48-L52

test/AspNet.Security.OAuth.Providers.Tests/Keycloak/KeycloakTests.cs
antosubash: test cases are added for all the mapped claims

src/AspNet.Security.OAuth.Keycloak/KeycloakPostConfigureOptions.cs
antosubash: I have updated this.

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
antosubash: changed base URL property to domain.

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
antosubash: MapAll() is required to map custom claims. others were the default scopes.

antosubash:
@martincostello I have made the required changes. please take a look again when you have time.

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
martincostello: I'm still not convinced this is needed. None of our other providers call it.

This feels like maybe something specific to your use case that would be configured in your call to `AddKeycloak()` when you configure the provider, rather than something it does by default.

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -37,6 +39,10 @@ public KeycloakAuthenticationOptions()
             ClaimActions.MapJsonKey(ClaimTypes.Role, "roles");
         }
 
-        public string BaseUrl { get; set; } = string.Empty;
+        /// <summary>
+        /// Gets or sets the Keycloak domain (Org URL) to use for authentication.
+        /// For example: 'keycloakdomain.com'.
```
martincostello:
```suggestion
        /// For example: <c>keycloakdomain.com</c>.
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -29,6 +29,8 @@ public KeycloakAuthenticationOptions()
             Scope.Add("openid");
```
martincostello, line 30:
```suggestion

            Scope.Add("openid");
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
antosubash: I followed the Okta options from here https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/97c1f79f4c09c492d10d1b7aec37aea2ceb49c64/src/AspNet.Security.OAuth.Okta/OktaAuthenticationOptions.cs#L27

but you are right you can add these from `AddKeycloak()` i have removed the default scopes.

antosubash:
@martincostello I have removed the default scopes and added the doc file. please take a look again when you have time.

docs/keycloak.md
```diff
@@ -0,0 +1,28 @@
...
+                    options.Scope.Add("roles");
+                    options.SaveTokens = true;
+                })
+```
+
+## Required Additional Settings
+
+| Property Name | Property Type | Description                                                                         | Default Value |
+| :------------ | :------------ | :---------------------------------------------------------------------------------- | :------------ |
+| `Domain`      | `string?`     | The Keycloak domain to use for authentication. This is the url of realm in Keycloak | `null`        |
```
martincostello:
```suggestion
| `Domain`      | `string?`     | The Keycloak domain to use for authentication. This is the URL of the realm in Keycloak. | `null`        |
```

docs/keycloak.md
```diff
@@ -0,0 +1,28 @@
...
+                {
+                    options.SignInScheme = IdentityConstants.ExternalScheme;
+                    options.ClientId = "my-client-id";
+                    options.ClientSecret = "my-client-secret";
+                    options.Domain = "http://{my-url}/auth/realms/{realm-name}";
+                    options.Scope.Add("openid");
+                    options.Scope.Add("email");
+                    options.Scope.Add("roles");
+                    options.SaveTokens = true;
+                })
```
martincostello:
```suggestion
        .AddKeycloak(options =>
        {
            options.SignInScheme = IdentityConstants.ExternalScheme;
            options.ClientId = "my-client-id";
            options.ClientSecret = "my-client-secret";
            options.Domain = "http://{my-url}/auth/realms/{realm-name}";
            options.Scope.Add("openid");
            options.Scope.Add("email");
            options.Scope.Add("roles");
            options.SaveTokens = true;
        })
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
martincostello: Is there a minimum scope required, or does the provider work out-of-the-box with exactly zero configured scopes?

The general idea is that the minimum scopes required to log in at all are configured in the provider, and any other additional scopes are configured by the user in their application.

My original point still stands though - none of the other providers need to call `MapAll()`, so I don't think this call should be here at all.

antosubash: yes, the minimum of `openid` is required for the login to work out of the box. I will update it and remove the `MapAll()`. It is needed if you have a custom claim in your app.

antosubash:
@martincostello removed the `MapAll()` and added the default `openid` scope which is required for the app to work out of the box. Thanks for the explanation and help. please take a look again when you are free.

martincostello (approved):

martincostello:
PR looks good now, but the tests are failing.

antosubash:
@martincostello  fixed the test.

martincostello:
Provider now available from [NuGet.org](https://www.nuget.org/packages/AspNet.Security.OAuth.Keycloak/) - thanks again for your contribution.

PR #548, opened 2021-05-19
Removes additional scopes from GitLab provider

### LINKED ISSUE #545: GitLab Provider broken
**Describe the bug**
When using `AspNet.Security.OAuth.GitLab` the sign-in middleware is broken and results in 
```
Exception: Correlation failed.
Unknown location

Exception: An error was encountered while handling the remote login.
Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler<TOptions>.HandleRequestAsync()
```

Another issue is that the default `Scope` is outdated as GitLab only allows these scopes:
`api, read_user, read_api, read_repository...` and therefore the default `openid profile email` are unsupported and should be replaced with `read_user`.

When the scope are correctly set to `read_user` the redirect to GitLab works fine but when allowed and redirected to the fallback url `/signin-gitlab` route fails with an exception.

**Steps To reproduce**
Register GitLab Provider and try to login.
```c#
       services.AddAuthentication(options =>
                {
                    options.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                }).AddGitLab(options =>
                {
                    options.ClientId = Configuration["Gitlab:ClientId"];
                    options.ClientSecret = Configuration["Gitlab:ClientSecret"];
                    options.Scope.Clear();
                    options.Scope.Add("read_user");
                }).AddCookie();
```

**Expected behaviour**
GitLab fallbacks to `/signin-gitlab` after login which parses the data and successfully authenticates user.

**Actual behaviour**
When redirected to `/signin-gitlab` an exception is thrown.
```
System.Exception: An error was encountered while handling the remote login.
 ---> System.Exception: Correlation failed.
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Swashbuckle.AspNetCore.SwaggerUI.SwaggerUIMiddleware.Invoke(HttpContext httpContext)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

**System information:**
- OS Name:     pop
- OS Version:  20.10
- OS Platform: Linux
- .NET SDKs installed: 5.0.203
- .NET runtimes installed:
	- Microsoft.AspNetCore.App 5.0.6
  	- Microsoft.NETCore.App 5.0.6

Regards Artur

S0PEX:
This PR addresses the issue https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/545 as currently by default additional scopes e.g: openid, profile, email are included.  This isn't optimal as Gitlab will deny the authentication requests that include these additional scopes, if the target Gitlab application settings doesn't have them set as well.
Therefore, it would be better to specify them explicitly when needed.

### DIFF
--- a/src/AspNet.Security.OAuth.GitLab/GitLabAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.GitLab/GitLabAuthenticationOptions.cs
@@ -26,9 +26,7 @@ namespace AspNet.Security.OAuth.GitLab
             TokenEndpoint = GitLabAuthenticationDefaults.TokenEndpoint;
             UserInformationEndpoint = GitLabAuthenticationDefaults.UserInformationEndpoint;
 
-            Scope.Add("openid");
-            Scope.Add("profile");
-            Scope.Add("email");
+            // Available scopes: https://docs.gitlab.com/ee/integration/oauth_provider.html
             Scope.Add("read_user");
 
             ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "id");


martincostello (approved):

martincostello:
Thanks for the PR @S0PEX 🎈 

This will ship in the 5.0.5 release, but I'm not sure when that's going to publish yet as I'd like to include a number of the open PRs we have to add new providers included as part of the next release to NuGet.org.

In the meantime, if you'd like to remove your workaround you can use aprereleasee version of the 5.0.5 GitLab provider including this change from our MyGet feed.

PR #561, opened 2021-06-05
Use ConfigurationManager in Apple provider

### LINKED ISSUE #421: Consider using IdentityModel's ConfigurationManager in the Apple provider
Apple now supports standard OIDC provider discovery (https://appleid.apple.com/.well-known/openid-configuration), so we could muchly simplify our keys retrieval and caching story by using IdentityModel's `ConfigurationManager`.

Since it will likely require adopting API breaking changes, it's probably something that will have to wait until 5.0.

Related: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/417#issuecomment-624648902

martincostello:
Switch to using the `ConfigurationManager` and Apple's OpenID Connect discovery endpoint instead of the custom implementation.

Also changes `PrivateKeyBytes` to accept a `CancellationToken`.

Resolves #421.


src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
```diff
@@ -5,62 +5,67 @@
...
             {
-                CryptoProviderFactory = _cryptoProviderFactory,
-                IssuerSigningKeys = keySet.Keys,
-                ValidAudience = context.Options.ClientId,
-                ValidIssuer = context.Options.TokenAudience,
-            };
+                throw new InvalidOperationException($"Token validation parameters have not been set on the {nameof(AppleAuthenticationOptions)} instance.");
+            }
+
+            OpenIdConnectConfiguration configuration = await context.Options.ConfigurationManager.GetConfigurationAsync(context.HttpContext.RequestAborted);
```
kevinchalet: You need moar `var` love in your life 😃

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
```diff
@@ -5,62 +5,67 @@
...
+
+            context.Options.TokenValidationParameters.IssuerSigningKeys = configuration.JsonWebKeySet.Keys;
```
kevinchalet: Assuming this method can be concurrently called, this is a risky operation. Instead, you may want to clone `context.Options.TokenValidationParameters` and mutate the cloned instance.

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleIdTokenValidator.cs
```diff
@@ -5,62 +5,67 @@
...
 
             try
             {
-                context.Options.JwtSecurityTokenHandler.ValidateToken(context.IdToken, parameters, out _);
+                var result = context.Options.SecurityTokenHandler.ValidateToken(context.IdToken, context.Options.TokenValidationParameters);
+
+                if (result.Exception != null || !result.IsValid)
```
kevinchalet: `!=` -> `is not`? 😃

src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs
```diff
@@ -133,7 +130,7 @@ private SigningCredentials CreateSigningCredentials(string keyId, ECDsa algorith
 
             // Use a custom CryptoProviderFactory so that keys are not cached and then disposed of, see below:
             // https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1302
-            return new SigningCredentials(key, SecurityAlgorithms.EcdsaSha256Signature)
+            return new SigningCredentials(key, SecurityAlgorithms.EcdsaSha256)
```
martincostello, line 133: This change is because when `JwtSecurityTokenHandler` was used, its internals include some code to coerce algorithm names to be those supported for JWTs:

  * [Mapping table](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L82)
  * [Usage](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L551)
  * [Coercion](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/System.IdentityModel.Tokens.Jwt/JwtHeader.cs#L101-L102)

`JsonWebTokenHandler` does not do that, and [uses the algorithm name verbatim](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/b45a5b4cf4e918ca2b2c12ff1951f1be7b592bac/src/Microsoft.IdentityModel.JsonWebTokens/JsonWebTokenHandler.cs#L147), which was causing the wrong algorithm name to be sent to Apple, causing it to be rejected.

martincostello:
Are you happy with this as-is Kévin?

kevinchalet:
Looks great! 👍🏻

PR #579, opened 2021-07-09
Add support for Dropbox response_type and token_access_type params

### LINKED ISSUE #578: Add support for Dropbox token_access_type and response_type
The Dropbox provider currently does not support the ability to specify the response_type or token_access_type parameters.  These two parameters allow you to specify Code authentication flow and enable the ability to grant offline access and receive refresh tokens.

I have submitted a pull request along side this feature request that implements the necessary changes for this new feature.

kiddailey:
This pull request address Feature Request #578 and adds support for the Dropbox token_access_type and response_type parameters:

**ResponseType** allows specification of the grant type. Per the [documentation](https://www.dropbox.com/developers/documentation/http/documentation):

> The grant type requested, either token or code. 

**AccessType** allows for the ability to request offline access and receive refresh_tokens with authentication tokens.   Per the [documentation](https://www.dropbox.com/developers/documentation/http/documentation):

>  If this parameter is set to offline, then the access token payload returned by a successful /oauth2/token call will contain a short-lived access_token and a long-lived refresh_token that can be used to request a new short-lived access token as long as a user's approval remains valid. If set to online then only a short-lived access_token will be returned. If omitted, the response will default to returning a long-lived access_token if they are allowed in the app console. If long-lived access tokens are disabled in the app console, this parameter defaults to online

### DIFF
--- a/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
@@ -20,11 +20,6 @@ namespace AspNet.Security.OAuth.Dropbox
         /// </summary>
         public string? AccessType { get; set; }
 
-        /// <summary>
-        /// Gets or sets the Response Type requested from Dropbox:  Token or Code
-        /// </summary>
-        public string? ResponseType { get; set; }
-
         public DropboxAuthenticationOptions()
         {
             ClaimsIssuer = DropboxAuthenticationDefaults.Issuer;
--- a/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
  * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
  * for more information concerning the license and the contributors participating to this project.
@@ -15,6 +15,16 @@ namespace AspNet.Security.OAuth.Dropbox
     /// </summary>
     public class DropboxAuthenticationOptions : OAuthOptions
     {
+        /// <summary>
+        /// Gets or sets what the response type from Dropbox should be:  Online, Offline or Legacy
+        /// </summary>
+        public string? AccessType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Response Type requested from Dropbox:  Token or Code
+        /// </summary>
+        public string? ResponseType { get; set; }
+
         public DropboxAuthenticationOptions()
         {
             ClaimsIssuer = DropboxAuthenticationDefaults.Issuer;
--- a/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
@@ -16,7 +16,7 @@ namespace AspNet.Security.OAuth.Dropbox
     public class DropboxAuthenticationOptions : OAuthOptions
     {
         /// <summary>
-        /// Gets or sets what the response type from Dropbox should be:  Online, Offline or Legacy
+        /// Gets or sets what the response type from Dropbox should be: Online, Offline or Legacy.
         /// </summary>
         public string? AccessType { get; set; }
 
--- a/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
@@ -16,7 +16,7 @@ namespace AspNet.Security.OAuth.Dropbox
     public class DropboxAuthenticationOptions : OAuthOptions
     {
         /// <summary>
-        /// Gets or sets what the response type from Dropbox should be: Online, Offline or Legacy.
+        /// Gets or sets what the response type from Dropbox should be: online, offline or legacy.
         /// </summary>
         public string? AccessType { get; set; }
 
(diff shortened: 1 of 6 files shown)


src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
```diff
@@ -15,6 +15,16 @@ namespace AspNet.Security.OAuth.Dropbox
     /// </summary>
     public class DropboxAuthenticationOptions : OAuthOptions
     {
+        /// <summary>
+        /// Gets or sets what the response type from Dropbox should be:  Online, Offline or Legacy
```
martincostello:
```suggestion
        /// Gets or sets what the response type from Dropbox should be: Online, Offline or Legacy.
```

src/AspNet.Security.OAuth.Dropbox/DropboxAuthenticationOptions.cs
```diff
@@ -15,6 +15,16 @@ namespace AspNet.Security.OAuth.Dropbox
...
+        /// </summary>
+        public string? AccessType { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Response Type requested from Dropbox:  Token or Code
```
martincostello:
```suggestion
        /// Gets or sets the response type requested from Dropbox: Token or Code.
```

martincostello (commented):
Thanks for the contribution 👍

Just a few extra things needed:

1. Please add a `dropbox.md` file to the `docs` folder for the additional parameters based on the format used by the existing provider documentation.
2. Please add at least one additional test for the Dropbox provider using both of the new options to cover the new code paths.

Could you also confirm you've tested this works as expected with the real Dropbox APIs?

kevinchalet:
@kiddailey thanks for your PR!

It's worth noting that `response_type=token` cannot work with server-side clients unless the OAuth 2.0 server supports the `response_mode=form_post` extension, as `response_type=token` implies returning the token material as part of the URL fragment, which is not sent by browsers to HTTP servers. I don't think Dropbox supports `response_mode=form_post` and if even if it did, it's not something the OAuth 2.0 base handler supports anyway (and it doesn't even support the implicit flow).

Personally, I'd remove the `ResponseType` thing and focus on the `AccessType` addition. We could even go one step further and set it to `offline` to encourage developers to use refresh tokens instead of the long-lived access tokens that were deprecated by Dropbox (which is a good thing).

kiddailey:
Thanks for the great feedback!

@martincostello: I'll take care of adding these as soon as I can.  I have tested it partially against Dropbox.  Specifically, in a Xamarin app using XamarinEssential's WebAuthenticator with my own server facilitating.  Everything appears to work as expected, but I still need to finish testing and implement the refresh_token handling, which I should get to today.  In any case, Dropbox appears to be responding as expected.

Edit: Confirmed my previous tests that I can receive tokens and refresh tokens as expected with code/offline and can authenticate with them.   ~~Testing the fetch of a new token using the provided refresh is still pending.~~ The Dropbox API reportedly automatically handles renewing the token if refresh token is provided on instantiation.  ~~Testing to confirm with my use case (takes 4 hours for short-lived tokens to expire).~~ Yes, the refresh token I received was able to be used to authenticate again.

@kevinchalet: Agreed.  As I specifically needed code flow and offline refresh tokens, I mainly added this so that I didn't break existing behavior.  Would it be acceptable to include the information you mention, note the deprecation in the readme, and leave this in, or would it be preferred to note the deprecation and make offline a constant?

kiddailey:
Readme and tests added.  And to confirm from my earlier post, I have tested this against Dropbox actual in an "offline" app that refreshes tokens as needed and it works as expected.

You'll probably note that I removed the ResponseType support per kevinchalet's suggestion.  In writing the tests and looking more at things, I realized that the response_type was already being set to "code" (obviously) and that "token" was likely not going to work properly without more effort.   For those reasons, I went ahead and removed it as suggested.

If the consensus is to make "offline" the default, let me know and I'll make that change as well -- and probably note it in the readme as it may break things for some.

test/AspNet.Security.OAuth.Providers.Tests/Dropbox/DropboxTests.cs
```diff
@@ -42,5 +45,71 @@ public async Task Can_Sign_In_Using_Dropbox(string claimType, string claimValue)
...
+            void ConfigureServices(IServiceCollection services)
+            {
+                services.PostConfigureAll<DropboxAuthenticationOptions>((options) =>
+                {
+                    options.AccessType = value;
+                    options.Events = new OAuthEvents
+                    {
+                        OnRedirectToAuthorizationEndpoint = ctx =>
+                        {
+                            accessTypeIsSet = ctx.RedirectUri.Contains($"token_access_type={value}", StringComparison.InvariantCulture);
```
martincostello:
```suggestion
                            accessTypeIsSet = ctx.RedirectUri.Contains($"token_access_type={value}", StringComparison.OrdinalIgnoreCase);
```

martincostello (commented):
Just one minor change. Otherwise looks good to me.

Let's not change the default behaviour as I'd consider that to be a breaking change that we wouldn't do outside of a major release.

martincostello (approved):

kevinchalet (approved):
Looks great, thanks!

Do you think it's worth introducing a `DropboxAuthenticationConstants` class containing a nested `TokenAccessTypes` class with online/offline/legacy constants?

kiddailey:
> Do you think it's worth introducing a `DropboxAuthenticationConstants` class containing a nested `TokenAccessTypes` class with online/offline/legacy constants?

That is a great question!  I originally did write it that way, but when looking at the style of the other providers, it seemed that many were using simple strings so I went with that.

Personally I do prefer doing it with constants, though that does remove a little bit of flexibility should the dropbox API change before the provider can be updated.  Let me know if you'd like me to change it and I'll take care of it.

Thanks!

martincostello:
I think for now we can just take this as-is, rather than add some extra constants, unless anyone feels strongly the other way.

martincostello:
This change is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Dropbox/5.0.9

PR #582, opened 2021-07-14
Improvement for LINE provider

linmasaki:
Recently, I discovered that if users didn't log in to the LINE first at the browser, the login page would not show for users.

Set the prompt value to an empty string can resolve this issue.


martincostello:
This seems a reasonable fix based on the description. As we can't validate this behaviour with our CI tests, can you confirm that running this change against the real Line service locally results in the expected behaviours for the two scenarios you've identified?

linmasaki:
Yes, I tested it on real LINE service and confirmed this alteration didn't break the old application.

martincostello:
Thanks for the fix - this will go out with the 5.0.9 release. Until that releases (after #579 is merged), you can consume the fix from our MyGet feed.

martincostello:
This change is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Line/5.0.9

linmasaki:
> This change is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Line/5.0.9

Thanks!

Issue #584, opened 2021-08-03
Discord OAuth provider returns .png avatar url for animated avatars


ndelta0:
**Describe the bug**
Claim for Discord Avatar Url always has an extension of `.png` even if the avatar is animated.

**Steps To reproduce**
n/a

**Expected behaviour**
Claim for avatar url ends in `.gif` for animated avatars/

**Actual behaviour**
Claim always ends with `.png`

**System information:**
 - OS: n/a
 - Library version 5.0.9
 - .NET version: n/a

**Additional context**
As per [Discord Documentation](https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints) users that have an animated avatar have a avatar hash that always starts with `a_`.

martincostello:
The _"best"_ way to fix this would be in version 6 as it requires a change to a constant, which is a breaking change.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/f9059d3f80ddd4158b298820a0008329545ee45c/src/AspNet.Security.OAuth.Discord/DiscordAuthenticationConstants.cs#L17

For earlier versions you can work around this by either:

1. Changing the format string used to create the avatar claim: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/f9059d3f80ddd4158b298820a0008329545ee45c/src/AspNet.Security.OAuth.Discord/DiscordAuthenticationOptions.cs#L25-L30
2. Adding your own custom claims action that does what you would like it to do: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/f9059d3f80ddd4158b298820a0008329545ee45c/src/AspNet.Security.OAuth.Discord/DiscordAuthenticationOptions.cs#L55-L61

martincostello:
Rather than try and maintain the code to do this and keep in sync with Discord, we've removed this functionality from the v6 release and won't be fixing this in the v5 provider.

Here is a snippet from the updated provider documentation for v6 that shows how to configure this for users who want/need the claim.

---

## Avatars as Claims

Versions of the Discord provider before version `6.0.0` would automatically map the user's avatar URL as the `urn:discord:avatar:url` claim.

[...]

PR #585, opened 2021-08-05
Remove claim for Discord avatar

martincostello:
Rather than maintain logic to understand how Discord builds up avatar URLs and handles different image formats, this PR removes the computed "avatar URL" claim for the v6 release while we have the opportunity to make breaking changes.

Consumers can instead author their own `MapCustomJson(...)` on the `ClaimActions` and derive the URI based on their own requirements.

As this is pretty easy to do using the extensibility that already exists, I think this is a long-term more preferable and maintainable solution than us trying to keep in step with whatever it is that Discord is doing for their avatar image storage.

If we decide this is the route we want to go down, I'll add an example implementation that mirrors what previous versions did with a tweak to support the animated case into the Markdown document. Users can then copy-paste that into their app configuration if they need it.

We could also optionally add `[Obsolete]` to these properties in the dev branch for v5 to give some advance notice to it being removed in v6.

Relates to #584.

### DIFF
--- a/docs/discord.md
+++ b/docs/discord.md
@@ -25,7 +25,7 @@ _None._
 
 Versions of the Discord provider before version `6.0.0` would automatically map the user's avatar URL as the `urn:discord:avatar:url` claim.
 
-This functionality is no longer built-in (see #584 and #585), but can be added to your application with some extra code similar to that shown in the sample below.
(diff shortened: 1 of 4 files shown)


kevinchalet (approved):
Sounds like the most reasonable approach 👍🏻

PR #590, opened 2021-08-24
Bump default API version for VK

### LINKED ISSUE #589: Need to bump VK API version
There is VK notification that library is passing obsolete API version https://vk.com/dev/lifecycle_notification
It will be shutdown at 2nd September, 2021

Change [VkontakteAuthenticationDefaults.ApiVersion](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/ed0c8bfd527aee56fd821c0e8d8e2908f16679b9/src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs#L55) to 5.131 

Every user can override correct version in Options, but it will be additional work for users of library

leotsarev:
https://vk.com/dev/lifecycle_notification

fixes #589

### DIFF
--- a/src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs
@@ -52,6 +52,6 @@ namespace AspNet.Security.OAuth.Vkontakte
         /// <summary>
         /// Default API version.
         /// </summary>
-        public const string ApiVersion = "5.78";
+        public const string ApiVersion = "5.131";
     }
 }
--- a/test/AspNet.Security.OAuth.Providers.Tests/Vkontakte/bundle.json
+++ b/test/AspNet.Security.OAuth.Providers.Tests/Vkontakte/bundle.json
@@ -13,7 +13,7 @@
       }
     },
     {
-      "uri": "https://api.vk.com/method/users.get.json?access_token=secret-access-token&v=5.78&fields=id,first_name,last_name,photo_rec,photo,hash",
+      "uri": "https://api.vk.com/method/users.get.json?access_token=secret-access-token&v=5.131&fields=id,first_name,last_name,photo_rec,photo,hash",
       "contentFormat": "json",
       "contentJson": {
         "response": [


src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs
```diff
@@ -52,6 +52,6 @@ public static class VkontakteAuthenticationDefaults
         /// <summary>
         /// Default API version.
         /// </summary>
-        public const string ApiVersion = "5.78";
+        public const string ApiVersion = "5.131";
```
martincostello, line 55: As this is a constant, technically this is a breaking change. Anything depending on this library wouldn't pick up the fix without explicit re-compilation.

I think for v5 it would be better to have the options class not use the constant, and instead assign this value directly.

Then in v6 we can make a breaking change and make this a property or field, rather than a constant.

martincostello:
The tests need updating too:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/ed0c8bfd527aee56fd821c0e8d8e2908f16679b9/test/AspNet.Security.OAuth.Providers.Tests/Vkontakte/bundle.json#L16

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs
leotsarev: >Anything depending on this library wouldn't pick up the fix without explicit re-compilation.

I don't see that as problem. They need to recompile anyway to use new version. 

Scenarios that it won't be picked is:
1. ASP.NET project (Nuget version for provider updated) → Some other library (not recompiled) → Updated provider
2. ASP.NET project (not recompiled) → Provider (updated as binary directly on production)

I see both scenarios as extremely unlikely. Am I missing something?

leotsarev:
I'll fix unit test after discussion above will be resolved.

src/AspNet.Security.OAuth.Vkontakte/VkontakteAuthenticationDefaults.cs
martincostello: Sorry, it's early, so what I said isn't completely accurate.

The fix is fine, and not using the constant in the library itself would make no difference, but it wouldn't fix it for any library _using the constant_ as that would need recompiling against the fixed version to fix itself.

I agree that isn't likely (app depending on library X, which depends on this), but is a possible scenario. That said, there's no way to fix that retroactively without making a breaking change anyway. Of course any scenario in that predicament can just override the API version in code anyway.

I'll merge this once the test is fixed, then I'll merge this into the v6 branch and review all the usage of public constants and change them to fields where appropriate so the edge case for this doesn't persist going forwards.

leotsarev:
Fixed unit tests and verified changes locally in my project.

I appreciate that will be merged eariler than sooner, because we are on tight schedule before 2021-09-02

martincostello (approved):

leotsarev:
@martincostello thanks for this awesome project and for fast review!

martincostello:
No problem - the fix is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Vkontakte/5.0.11

PR #599, opened 2021-09-22
Add notion user workspace ownership parameter

mifopen:
https://developers.notion.com/changelog/space-level-integrations-will-be-deprecated-soon-migrate-your-oauth-flows#how-to-prepare-for-this-change

### DIFF
--- a/src/AspNet.Security.OAuth.Notion/NotionAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.Notion/NotionAuthenticationDefaults.cs
@@ -37,7 +37,7 @@ namespace AspNet.Security.OAuth.Notion
         /// <summary>
         /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


martincostello:
It's seems to me that this change is not needed - the documentation seems to suggest that the new query string parameter is a _temporary_ measure for the user to opt-in to the new behaviour for testing before it is enforced so they can check they're compatible with the new token/scoping requirements.

> What to expect on October 19
On October 19, we will migrate all existing workspace-level tokens to user-level tokens. We will also default to creating user-level tokens when a user goes through OAuth, **regardless of the owner parameter in the OAuth URL**.

If that's the case, you would just append the query string parameter in your own config temporarily and the provider does not need to change at all.

```csharp
services.AddAuthentication(options => /* Auth configuration */)
        .AddNotion(options =>
        {
            options.ClientId = "my-client-id";
            options.ClientSecret = "my-client-secret";
            options.AuthorizationEndpoint += "?owner=user";
        });
```

mifopen:
@martincostello Oh, you're absolutely right. Sorry for bothering you. Thank you!

martincostello:
No problem - thanks for raising it!

PR #601, opened 2021-09-24
Add test for OnCreatingTicket with EventsType

martincostello:
Add a unit test to validate #600 for all providers.


martincostello:
I audited the other usages of `Options.Events` instead of `Events` and found an occurrence in the Apple provider that lead to a similar bug. That's also now been fixed and tests added for those custom events too.

PR #611, opened 2021-10-08
Fix Keycloak throwing exception for Secret Key if we are using public access type

### LINKED ISSUE #610: Keycloak's "ClientSecret" is mandatory, but Keycloak with access type of "public" has no secret.
**Describe the bug**
If we set the Keycloak to use access type of "public", it will have no secret.

But Validate will fail if there is no "ClientSecret".

**Steps To reproduce**
1. Set a Keycloak server with 'public' access type. This is under the Client settings.

image

2. At ConfigureServices, do not set the ClientSecret.

            services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
            .AddJwtBearer(o => 
            {
                o.Authority = "http://localhost:8080/auth/realms/vuedemo";
                o.RequireHttpsMetadata = false;
            })
            .AddKeycloak(o => 
            {
                o.ClientId = "hello-vue";
                o.TokenEndpoint = "http://localhost:8080/auth";
            });

**Expected behaviour**
Validate should succeed.

**Actual behaviour**
Validate is throwing exception "The 'ClientSecret' must be provided".

**System information:**
 - OS: Win10
 - Library Version 5.0.15
 - .NET version 5.0.401

FaridAhamat:
This is a fix for #610 

I added a new property for Keycloak's access type. It will swallow the exception thrown for missing Secret Key if the access type is public, otherwise everything stays the same.

### DIFF
--- a/src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
+++ b/src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
@@ -4,6 +4,7 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
+using System;
 using System.Security.Claims;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.OAuth;
@@ -34,6 +35,48 @@ namespace AspNet.Security.OAuth.Keycloak
             ClaimActions.MapJsonKey(ClaimTypes.Name, "name");
             ClaimActions.MapJsonKey(ClaimTypes.GivenName, "given_name");
             ClaimActions.MapJsonKey(ClaimTypes.Role, "roles");
+        }
+
+        /// <summary>
+        /// Gets or sets the value for Keycloak client's access type
+        /// </summary>
+        public KeycloakAuthenticationAccessType AccessType { get; set; }
+
+        /// <inheritdoc />
+        public override void Validate()
+        {
+            try
+            {
+                // HACK
+                // We want all of the base validation except for ClientSecret,
+                // so rather than re-implement it all, catch the exception thrown
+                // for that being null and only throw if we aren't using public access type.
+                // This does mean that three checks have to be re-implemented
+                // because the won't be validated if the ClientSecret validation fails.
+                base.Validate();
+            }
+            catch (ArgumentException ex) when (ex.ParamName == nameof(ClientSecret))
+            {
+                if (AccessType != KeycloakAuthenticationAccessType.Public)
+                {
+                    throw;
+                }
+            }
+
+            if (string.IsNullOrEmpty(AuthorizationEndpoint))
+            {
+                throw new ArgumentException($"The '{nameof(AuthorizationEndpoint)}' option must be provided.", nameof(AuthorizationEndpoint));
+            }
+
+            if (string.IsNullOrEmpty(TokenEndpoint))
+            {
+                throw new ArgumentException($"The '{nameof(TokenEndpoint)}' option must be provided.", nameof(TokenEndpoint));
+            }
+
+            if (!CallbackPath.HasValue)
+            {
+                throw new ArgumentException($"The '{nameof(CallbackPath)}' option must be provided.", nameof(CallbackPath));
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -36,6 +37,48 @@ public KeycloakAuthenticationOptions()
...
+                // because the won't be validated if the ClientSecret validation fails.
+                base.Validate();
+            }
+            catch (ArgumentException ex) when (ex.ParamName == nameof(ClientSecret))
+            {
+                if (AccessType != KeycloakAuthenticationAccessType.Public)
+                {
+                    throw;
+                }
+            }
```
martincostello, line 64: This could be simplified to just swallow in the required case, rather than need to throw when it isn't.

```suggestion
            catch (ArgumentException ex) when (ex.ParamName == nameof(ClientSecret) && AccessType == KeycloakAuthenticationAccessType.Public)
            {
                // No client secret is required for a public key.
                // See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/610
            }
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -36,6 +37,48 @@ public KeycloakAuthenticationOptions()
...
+        /// </summary>
+        public KeycloakAuthenticationAccessType AccessType { get; set; }
+
+        /// <inheritdoc />
+        public override void Validate()
+        {
+            try
+            {
+                // HACK
+                // We want all of the base validation except for ClientSecret,
```
martincostello, line 51: This makes the Task List in Visual Studio more informative.

image

vs.

image


```suggestion
                // HACK We want all of the base validation except for ClientSecret,
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -36,6 +37,48 @@ public KeycloakAuthenticationOptions()
             ClaimActions.MapJsonKey(ClaimTypes.Role, "roles");
         }
 
+        /// <summary>
+        /// Gets or sets the value for Keycloak client's access type
```
martincostello, line 41:
```suggestion
        /// Gets or sets the value for the Keycloak client's access type.
```

src/AspNet.Security.OAuth.Keycloak/KeycloakAuthenticationOptions.cs
```diff
@@ -36,6 +37,48 @@ public KeycloakAuthenticationOptions()
...
+        /// </summary>
+        public KeycloakAuthenticationAccessType AccessType { get; set; }
+
+        /// <inheritdoc />
+        public override void Validate()
```
martincostello, line 46: Also please put this method after all properties.

martincostello:
Could you add the unit test below when you rebase and address the other feedback please?

Also please add the new setting to the optional section of the Keycloak provider documentation as a new table following the format of the table in the section above:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/324262f8a66d8844a31686ab254f5a4338b7187d/docs/keycloak.md#L43

```csharp
[Theory]
[InlineData(ClaimTypes.NameIdentifier, "995c1500-0dca-495e-ba72-2499d370d181")]
[InlineData(ClaimTypes.Email, "john@smith.com")]
[InlineData(ClaimTypes.GivenName, "John")]
[InlineData(ClaimTypes.Role, "admin")]
[InlineData(ClaimTypes.Name, "John Smith")]
public async Task Can_Sign_In_Using_Keycloak_Public_AccessType(string claimType, string claimValue)
{
    // Arrange
    static void ConfigureServices(IServiceCollection services)
    {
        services.PostConfigureAll<KeycloakAuthenticationOptions>((options) =>
        {
            options.AccessType = KeycloakAuthenticationAccessType.Public;
            options.ClientSecret = string.Empty;
            options.Domain = "keycloak.local";
        });
    }

    using var server = CreateTestServer(ConfigureServices);

    // Act
    var claims = await AuthenticateUserAsync(server);

    // Assert
    AssertClaim(claims, claimType, claimValue);
}
```

martincostello:
Hi @FaridAhamat - thanks again for your PR and fix. As I'd like to release this change and another for Keycloak today, I've opened a new PR which uses your initial commit and then applies the feedback I gave here in a new PR: #613.

You're a co-author on the original commit, so you'll still get credit for the original change 😃

PR #615, opened 2021-10-14
Update default Vkontakte API version in docs

alexbagirov:
As per code, default Vkontakte API version used is 5.131, however docs claims that it is lower. This PR updates docs default value.


martincostello (approved):

martincostello:
Thanks!

PR #619, opened 2021-10-28
Updated LiChess to use required pkce authentication

tanczosm:
LiChess updated their OAuth requirements to require PKCE.  This revision sets the default options to use PKCE, updates the TokenEndPoint in defaults, and adds an override for ExchangeCodeAsync that follows their required mechanism for authentication.

[https://lichess.org/api#tag/OAuth](https://lichess.org/api#tag/OAuth)

### DIFF
--- a/src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
@@ -4,11 +4,8 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
-using System;
-using System.Collections.Generic;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using System.Net.Mime;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
 using System.Text.Json;
--- a/src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
@@ -4,8 +4,11 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
+using System;
+using System.Collections.Generic;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Mime;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
 using System.Text.Json;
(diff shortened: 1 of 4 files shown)


martincostello (approved):

martincostello:
This looks OK to me - I'll look at merging and releasing this tomorrow. Thanks for the contribution!

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
```diff
@@ -39,6 +42,48 @@ public class LichessAuthenticationHandler : OAuthHandler<LichessAuthenticationOp
         {
         }
 
+        protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
```
kevinchalet: Do we absolutely need to override this method? Isn't the native PKCE support in the OAuth 2.0 enough for this provider?

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
```diff
@@ -39,6 +42,48 @@ public class LichessAuthenticationHandler : OAuthHandler<LichessAuthenticationOp
...
+        {
+            var tokenRequestParameters = new Dictionary<string, string?>()
+            {
+                ["grant_type"] = "authorization_code",
+                ["client_id"] = Options.ClientId,
+                ["code"] = context.Code,
+                ["redirect_uri"] = context.RedirectUri
```
kevinchalet: I don't see `client_secret`. Does this provider only support public clients?

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
tanczosm: I will verify if this is the case.  I think when testing with the default implementation I had one of the endpoints misconfigured as they updated the token endpoint.  It doesn't use client_secret but this may not matter.

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
tanczosm: It does not use client_secret.  I'll check if I can just revert to the default token exchange.

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
tanczosm: It does work using native PKCE so I'm removing this method.

src/AspNet.Security.OAuth.Lichess/LichessAuthenticationHandler.cs
tanczosm: It does work using native PKCE so I'm removing this method.

tanczosm:
I have removed the ExchangeCodeAsync method as it turns out not to need to be overridden to work with LiChess.  Updating the endpoints and setting PKCE to be required by default was sufficient to get this working again.

It is ready to be merged.

martincostello:
Thanks for your contribution once again @tanczosm - this change is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Lichess/5.0.18

PR #633, opened 2022-01-03
Allow package baseline override

martincostello:
Allow individual packages to override the baseline for any new providers added since the last major release.

For example, a new provider added tomorrow could specify a baseline of 6.0.2, but disable baseline validation until it was published.

```xml
  <PropertyGroup>
    <PackageValidationBaselineVersion>6.0.2</PackageValidationBaselineVersion>
  </PropertyGroup>
```

Otherwise, a new provider wouldn't build because the package validation analyzer would fail because it wouldn't be able to find it in NuGet.org.

In another PR I'll look at adding a list of released providers so that this doesn't need any thought by contributors and "just works", becoming a small piece of extra release process admin for new providers instead.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -5,7 +5,7 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>2</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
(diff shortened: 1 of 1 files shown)


martincostello:
Prototype of the approach mentioned above here: https://github.com/martincostello/AspNet.Security.OAuth.Providers/commit/2762c3660770f61df9f83c2daf3719ce86d6e39b

It _mostly_ works, but doesn't work well if the NuGet packages that are the baselines are not on disk locally already.

PR #635, opened 2022-01-05
Refactor BuildChallengeUrl overrides

martincostello:
Refactor the overrides of `BuildChallengeUrl()` to still leverage the base class' functionality for PKCE ([code](https://github.com/dotnet/aspnetcore/blob/4e5cdf955be6cca550e3db38e5de98f1643a67ac/src/Security/Authentication/OAuth/src/OAuthHandler.cs#L291-L323)) and reduce duplication/re-implementation. While these providers might not actually support PKCE, any that do cannot leverage it even if `UsePkce` is set to true on these providers' options in the consuming applications.

Also updates the .NET SDK to 6.0.101.

### DIFF
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
@@ -221,7 +221,7 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
         if (Options.UsePkce)
         {
             byte[] bytes = RandomNumberGenerator.GetBytes(32);
-            string codeVerifier = Microsoft.AspNetCore.WebUtilities.Base64UrlTextEncoder.Encode(bytes);
+            string codeVerifier = WebEncoders.Base64UrlEncode(bytes);
 
             // Store this for use during the code redemption.
             properties.Items.Add(OAuthConstants.CodeVerifierKey, codeVerifier);
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
@@ -51,7 +51,7 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
     protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
     {
         // See https://opendocs.alipay.com/apis/api_9/alipay.system.oauth.token for details.
-        var sortedParams = new SortedDictionary<string, string?>()
+        var tokenRequestParameters = new SortedDictionary<string, string?>()
         {
             ["app_id"] = Options.ClientId,
             ["charset"] = "utf-8",
@@ -63,9 +63,16 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
             ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
             ["version"] = "1.0",
         };
-        sortedParams.Add("sign", GetRSA2Signature(sortedParams));
+        tokenRequestParameters.Add("sign", GetRSA2Signature(tokenRequestParameters));
 
-        string address = QueryHelpers.AddQueryString(Options.TokenEndpoint, sortedParams!);
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        string address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);
 
         using var response = await Backchannel.GetAsync(address, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
         if (!response.IsSuccessStatusCode)
@@ -93,7 +100,7 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
         [NotNull] OAuthTokenResponse tokens)
     {
         // See https://opendocs.alipay.com/apis/api_2/alipay.user.info.share for details.
-        var sortedParams = new SortedDictionary<string, string?>()
+        var parameters = new SortedDictionary<string, string?>()
         {
             ["app_id"] = Options.ClientId,
             ["auth_token"] = tokens.AccessToken,
@@ -104,9 +111,9 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
             ["timestamp"] = Clock.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture),
             ["version"] = "1.0",
         };
-        sortedParams.Add("sign", GetRSA2Signature(sortedParams));
+        parameters.Add("sign", GetRSA2Signature(parameters));
 
-        string address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, sortedParams!);
+        string address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, parameters);
 
         using var response = await Backchannel.GetAsync(address, Context.RequestAborted);
 
@@ -203,7 +210,7 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
         var scopeParameter = properties.GetParameter<ICollection<string>>(OAuthChallengeProperties.ScopeKey);
         var scope = scopeParameter != null ? FormatScope(scopeParameter) : FormatScope();
 
-        var parameters = new Dictionary<string, string>
+        var parameters = new Dictionary<string, string?>
         {
             ["app_id"] = Options.ClientId, // Used instead of "client_id"
             ["scope"] = scope,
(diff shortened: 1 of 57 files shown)


martincostello:
While looking at this I also found some other things that didn't seem quite right.

1. Some providers had an override for `FormatScopes()` but not `FormatScopes(IEnumerable<string>)`. As the former just calls the latter in `OAuthHandler<T>`, these providers should only have been overriding the latter. Providers that didn't now do, but as removing an override is technically a breaking change, I've left them in with the appropriate behaviour that doesn't duplicate code. We can remove them in v7.
2. Various handlers that had an override for `BuildChallengeUrl()` did not honour `UsePkce`, further various handlers that had an override for `ExchangeCodeAsync()` also did not honour the presence of `OAuthConstants.CodeVerifierKey` in the properties. This meant that these handlers would either ignore `UsePkce` entirely and/or not consume the value when the token was exchanged. These handlers now implement the same behaviour as in the relevant `OAuthHandler<T>` methods, regardless of whether a given provider _actually_ supports PKCE.
3. The Shopify and Weixin handlers didn't follow the same pattern as other handlers/`OAuthHandler<T>` and use `OAuthChallengeProperties.ScopeKey` if present in the `AuthenticationProperties`. This has been fixed.
4. The Mixcloud handler had copy-pasted the whole implementation of `OAuthHandler<T>.HandleRemoteAuthenticateAsync()`, with a small change in the middle of the method. Instead, it now overrides `ExchangeCodeAsync()` and applies the behaviour change there. This is technically a behaviourally breaking change if someone overrides `ExchangeCodeAsync()` themselves, but I think this one is unlikely. If it does cause anyone any issues, we can always put the copy-pasta back and apply the change in v7 instead.

martincostello:
As this ended up being a bigger change than I originally expected, would be appreciated if you could give it a review when you have some time please @kevinchalet.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -213,22 +220,20 @@ protected override string BuildChallengeUrl([NotNull] AuthenticationProperties p
 
         if (Options.UsePkce)
         {
-            var bytes = RandomNumberGenerator.GetBytes(32);
-            var codeVerifier = Microsoft.AspNetCore.Authentication.Base64UrlTextEncoder.Encode(bytes);
+            byte[] bytes = RandomNumberGenerator.GetBytes(32);
```
kevinchalet: Curious: is there any particular reason for replacing `var` by explicit types?

kevinchalet: Nit: I'd probably replace `32` by `256 / 8` so it's immediately apparent we're generating a 256-bit random value.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -213,22 +220,20 @@ protected override string BuildChallengeUrl([NotNull] AuthenticationProperties p
...
+            string codeVerifier = Microsoft.AspNetCore.WebUtilities.Base64UrlTextEncoder.Encode(bytes);
```
kevinchalet: If you're interested in removing inconsistencies, you may want to use `WebEncoders.Base64UrlEncode()` instead.

(both `Microsoft.AspNetCore.WebUtilities.Base64UrlTextEncoder` and `Microsoft.AspNetCore.Authentication.Base64UrlTextEncoder` are simple wrappers around `WebEncoders`).

src/AspNet.Security.OAuth.Automatic/AutomaticAuthenticationHandler.cs
```diff
@@ -53,14 +53,15 @@ public partial class AutomaticAuthenticationHandler : OAuthHandler<AutomaticAuth
...
-        {
-            ["client_id"] = Options.ClientId,
-            ["response_type"] = "code",
-            ["scope"] = FormatScope(),
-            ["state"] = Options.StateDataFormat.Protect(properties)
-        });
+        var challengeUrl = base.BuildChallengeUrl(properties, redirectUri);
+
+        // The redirect_uri parameter is not allowed by Automatic and MUST NOT be sent.
+        var challengeUri = new Uri(challengeUrl);
```
kevinchalet: To be super explicit, should we specify `UriKind.Absolute`?

src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
```diff
@@ -66,7 +63,16 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
-        request.Content = new FormUrlEncodedContent(parameters!);
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/x-www-form-urlencoded"));
```
kevinchalet, line 76: I see it's not a change introduced in this PR but are we sure we use the right value here? The OAuth 2.0 specification requires returning JSON responses and we use `JsonDocument.Parse()` later in the method, so it's not clear to me why formurl encoding is used here. Just a mistake?

src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationHandler.cs
```diff
@@ -115,35 +126,61 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
...
     {
-        string stateValue = Options.StateDataFormat.Protect(properties);
+        var scopeParameter = properties.GetParameter<ICollection<string>>(OAuthChallengeProperties.ScopeKey);
+        var scope = scopeParameter != null ? FormatScope(scopeParameter) : FormatScope();
+
+        var parameters = new Dictionary<string, string?>()
+        {
+            ["appid"] = Options.ClientId,
+            ["scope"] = scope,
+            ["response_type"] = "code",
```
kevinchalet, line 138: Nit: unnecessary `,`.

src/AspNet.Security.OAuth.Yandex/YandexAuthenticationHandler.cs
```diff
@@ -54,18 +54,25 @@ public partial class YandexAuthenticationHandler : OAuthHandler<YandexAuthentica
...
-        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-        request.Headers.Authorization = CreateAuthorizationHeader();
-
-        var parameters = new Dictionary<string, string>
+        var tokenRequestParameters = new Dictionary<string, string>
         {
             ["grant_type"] = "authorization_code",
             ["redirect_uri"] = context.RedirectUri,
-            ["code"] = context.Code
+            ["code"] = context.Code,
```
kevinchalet, line 61: Here too 😃

kevinchalet (approved):
I'm not sure any of these providers supports PKCE so it's likely no one will ever benefit from your hard work but it doesn't hurt, so 🤷🏻‍♂️ 

Overall, looks great (you'll probably want to fix the `var` vs explicit types for consistency tho').

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: I can change it, but the 32 was lifted from [`OAuthHandler<T>`](https://github.com/dotnet/aspnetcore/blob/de8ee82692989d46d54d1e1be2d3254e4436d720/src/Security/Authentication/OAuth/src/OAuthHandler.cs#L306).

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: I'll update all of these to use `WebEncoders.Base64UrlEncode(challengeBytes)` the same as `OAuthHandler<T>` - this choice was arbitrary as I copied it from somewhere else (maybe that was in turn copied from an earlier version of `OAuthHandler<T>`) and there's two `Base64UrlTextEncoder` in scope, so needed the namespace to resolve the conflict and I just picked one.

src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
martincostello: I'll review it and check, but without having a working AzDo OAuth setup to verify it with, if it's not doing any harm I'd just leave it as is (maybe with a comment to review it with a working integration at some point).

src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationHandler.cs
martincostello: Again personal choice - I prefer to _always_ have the comma because:

1. Consistency
2. Adding a new item is a 1 line change, not a 2 line change

If you feel strongly about it I can take them back out.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: I prefer it (and it's set up that way in [EditorConfig](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/1dcd2404d00226cd1f77e2af5f1686d2af4204f1/.editorconfig#L88), but I can turn it off if you want) when the type isn't obvious when you're reading the code in GitHub.

In this case it probably only changed because I copy-pasted it from somewhere where it wasn't using `var` already as part of this clean up, rather than an intentional change.

src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
kevinchalet: Makes sense 👍🏻

src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationHandler.cs
kevinchalet: I find it a bit weird but I'm not strongly opposed to adding a `,` on each line if you prefer 😃
That said, things will now be inconsistent if you don't adopt this code style everywhere, which is not ideal.

martincostello: I've made the others match to have a trailing `,`, and which got you some more bonus `var` usage 😄

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
martincostello: We could use `SHA256.HashSizeInBits` instead when 7 lands: https://github.com/dotnet/runtime/pull/63685

kevinchalet: We could: technically, it would do exactly same thing. But semantically, it would be super weird: `HashSizeInBits` represents the size of the message digest (i.e the output of the hash method) while the random array we generate here is used as the message (i.e the input given to the hash method). SHA256 can accept inputs of any length up to 2^64-1 and there's no relationship between the size of the input and the output so it would be weird to mix the two things.

FWIW, 256 comes from the PKCE spec itself, that recommends generating code verifiers with at least 256 bits of entropy: https://datatracker.ietf.org/doc/html/rfc7636#section-7.1 (256 bits of entropy already give a very high level of security so not sure why anyone would want to opt for a higher value 😃)

PR #671, opened 2022-04-18
Fix the value declared by the weixin provider NameIdentifier claim

### LINKED ISSUE #670: GetExternalLoginInfoAsync returns null
**Describe the bug**

Provider: Wexin

① Call to _signInManager.GetExternalLoginInfoAsync() always returns null 

See [SignInManager](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Identity/Core/src/SignInManager.cs)

```csharp

 var providerKey = auth.Principal.FindFirstValue(ClaimTypes.NameIdentifier);
 var provider = items[LoginProviderKey] as string;
 if (providerKey == null || provider == null)
 {
       return null;
}

```

If the WeChat official account is not associated with the open platform，the unionid is null, so, the providerKey always return null

See [WeixinAuthenticationOptions](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationOptions.cs)

```csharp

  ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "unionid");

```

② If I don't use _signInManager.GetExternalLoginInfoAsync(),  but use HttpContext.AuthenticateAsync(IdentityConstants.ExternalScheme), the userIdClaim is also null

See [ExternalController](https://github.com/IdentityServer/IdentityServer4.Templates/blob/main/src/IdentityServer4AspNetIdentity/Quickstart/Account/ExternalController.cs)

```csharp

      private async Task<(ApplicationUser user, string provider, string providerUserId, IEnumerable<Claim> claims)>
            FindUserFromExternalProviderAsync(AuthenticateResult result)
        {
            var externalUser = result.Principal;

            // try to determine the unique id of the external user (issued by the provider)
            // the most common claim type for that are the sub claim and the NameIdentifier
            // depending on the external provider, some other claim type might be used
            var userIdClaim = externalUser.FindFirst(JwtClaimTypes.Subject) ??
                              externalUser.FindFirst(ClaimTypes.NameIdentifier) ??
                              throw new Exception("Unknown userid");

            // remove the user id claim so we don't include it as an extra claim if/when we provision the user
            var claims = externalUser.Claims.ToList();
            claims.Remove(userIdClaim);

            var provider = result.Properties.Items["scheme"];
            var providerUserId = userIdClaim.Value;

            // find external user
            var user = await _userManager.FindByLoginAsync(provider, providerUserId);

            return (user, provider, providerUserId, claims);
        }

```

Of course,  we can get the providerKey from the external provider as needed, but each provider's statement of storing the providerKey is inconsistent, if there are many providers, it requires us to do a lot of work, so I think it is not very friendly

So what I want to say is can the important declarations be unified?

**Steps To reproduce**

```csharp

            services.AddAuthentication()
                .AddWeixin("WeChat", "WeChat", options => {
                    options.ClientId = "wxb5f420cb0efc95a5";
                    options.ClientSecret = "";
                    options.AuthorizationEndpoint = "https://open.weixin.qq.com/connect/oauth2/authorize";
                    options.CallbackPath = "/signin-wechat";
                });

```

**Expected behaviour**

Call to _signInManager.GetExternalLoginInfoAsync()  returns not null 

**Actual behaviour**

Call to _signInManager.GetExternalLoginInfoAsync() always returns null 

**System information:**
 - OS: Windows 10
 - Library Version  6.0.4
 - .NET version  6.0.200

**Additional context**
IdentityServer4
AspIdentity

ArcherTrister:
See [aspnet-contrib/AspNet.Security.OAuth.Providers#670](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/670)


src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationHandler.cs
```diff
@@ -79,6 +79,16 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
 
+        (var openId, var unionId) = GetUserIdentifier(payload.RootElement);
+        if (!string.IsNullOrWhiteSpace(unionId))
+        {
+            identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, unionId, ClaimValueTypes.String, Options.ClaimsIssuer));
+        }
+        else
+        {
+            identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, openId!, ClaimValueTypes.String, Options.ClaimsIssuer));
+        }
```
martincostello: I think something like this would be a bit more DRY:

```suggestion
        (var openId, var unionId) = GetUserIdentifier(payload.RootElement);
        var nameIdentifier = !string.IsNullOrWhiteSpace(unionId) ? unionId : openId;
        identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, nameIdentifier!, ClaimValueTypes.String, Options.ClaimsIssuer));
```

martincostello (commented):
Could you add a test case for the Open ID scenario please?

src/AspNet.Security.OAuth.Weixin/WeixinAuthenticationHandler.cs
ArcherTrister: Yes, this is great, I'm very busy right now, I try to take the time to provide a test case

test/AspNet.Security.OAuth.Providers.Tests/WechatHasUnionId/WechatHasUnionIdTests.cs
```diff
@@ -0,0 +1,50 @@
+﻿/*
```
martincostello: Using the [Apple tests](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/test/AspNet.Security.OAuth.Providers.Tests/Apple/AppleTests.cs) as a basis, these new test cases can be placed in the existing `WeixinTests` class.

test/AspNet.Security.OAuth.Providers.Tests/WechatHasUnionId/bundle.json
```diff
@@ -0,0 +1,35 @@
+{
```
martincostello: If you use the [Apple provider's tests](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/test/AspNet.Security.OAuth.Providers.Tests/Apple/AppleTests.cs) as a guide, the HTTP registrations can all be put in the existing Weixin bundle file.

test/AspNet.Security.OAuth.Providers.Tests/WechatHasUnionId/WechatHasUnionIdTests.cs
ArcherTrister: The test cases in the WeixinTests class correspond to the weixin open platform, so I put the test cases in the WechatTests class

martincostello: But you've changed the code in the Weixin provider - the tests are organised by the namespace and assembly.

ArcherTrister: The WechatTests class already exists before, so it should be removed.

martincostello: Fair enough, I hadn't noticed. I'll move them around post-merge as it's confusing.

ArcherTrister: it's the norm,You're right.

martincostello:
Thanks for your contribution @ArcherTrister - this change has been published to [NuGet.org](https://www.nuget.org/packages/AspNet.Security.OAuth.Weixin/6.0.6).

PR #675, opened 2022-05-05
Add Xiaomi provider

ArcherTrister:
Add an OAuth 2.0 Xiaomi provider.

### DIFF
--- a/samples/Mvc.Client/Startup.cs
+++ b/samples/Mvc.Client/Startup.cs
@@ -44,8 +44,8 @@ public class Startup
 
         .AddTwitter(options =>
         {
-            options.ClientId = Configuration["Twitter:ConsumerKey"];
-            options.ClientSecret = Configuration["Twitter:ConsumerSecret"];
+            options.ClientId = Configuration["Twitter:ClientId"];
+            options.ClientSecret = Configuration["Twitter:ClientSecret"];
         })
 
         .AddGitHub(options =>
--- a/samples/Mvc.Client/Startup.cs
+++ b/samples/Mvc.Client/Startup.cs
@@ -44,8 +44,8 @@ public class Startup
 
         .AddTwitter(options =>
         {
-            options.ClientId = Configuration["Twitter:ClientId"];
-            options.ClientSecret = Configuration["Twitter:ClientSecret"];
+            options.ClientId = Configuration["Twitter:ConsumerKey"];
+            options.ClientSecret = Configuration["Twitter:ConsumerSecret"];
         })
 
         .AddGitHub(options =>
--- a/src/AspNet.Security.OAuth.XiaoMi/AspNet.Security.OAuth.XiaoMi.csproj
+++ b/src/AspNet.Security.OAuth.XiaoMi/AspNet.Security.OAuth.XiaoMi.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
+    <PackageValidationBaselineVersion>6.0.7</PackageValidationBaselineVersion>
   </PropertyGroup>
 
   <!-- TODO Remove once published to NuGet.org -->
--- a/src/AspNet.Security.OAuth.XiaoMi/AspNet.Security.OAuth.XiaoMi.csproj
+++ b/src/AspNet.Security.OAuth.XiaoMi/AspNet.Security.OAuth.XiaoMi.csproj
@@ -1,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
+  </PropertyGroup>
+
+  <!-- TODO Remove once published to NuGet.org -->
+  <PropertyGroup>
+    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>ASP.NET Core security middleware enabling XiaoMi authentication.</Description>
+    <Authors>ArcherTrister</Authors>
+    <PackageTags>aspnetcore;authentication;oauth;security;xiaomi</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="JetBrains.Annotations" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
--- a/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
@@ -1,0 +1,20 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+namespace AspNet.Security.OAuth.XiaoMi;
+
+/// <summary>
+/// Contains constants specific to the <see cref="XiaoMiAuthenticationHandler"/>.
+/// </summary>
+public static class XiaoMiAuthenticationConstants
+{
+    public static class Claims
+    {
+        public const string MiliaoIcon = "urn:xiaomi:miliaoIcon";
+        public const string MiliaoNick = "urn:xiaomi:miliaoNick";
+        public const string UnionId = "urn:xiaomi:unionId";
+    }
+}
--- a/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
+++ b/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
@@ -4,12 +4,12 @@
  * for more information concerning the license and the contributors participating to this project.
  */
 
-namespace AspNet.Security.OAuth.XiaoMi;
+namespace AspNet.Security.OAuth.Xiaomi;
 
 /// <summary>
-/// Contains constants specific to the <see cref="XiaoMiAuthenticationHandler"/>.
+/// Contains constants specific to the <see cref="XiaomiAuthenticationHandler"/>.
 /// </summary>
-public static class XiaoMiAuthenticationConstants
+public static class XiaomiAuthenticationConstants
 {
     public static class Claims
     {
--- a/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
@@ -1,0 +1,199 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using System.Net;
+using System.Security.Claims;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace AspNet.Security.OAuth.XiaoMi;
+
+public partial class XiaoMiAuthenticationHandler : OAuthHandler<XiaoMiAuthenticationOptions>
+{
+    public XiaoMiAuthenticationHandler(
+        [NotNull] IOptionsMonitor<XiaoMiAuthenticationOptions> options,
+        [NotNull] ILoggerFactory logger,
+        [NotNull] UrlEncoder encoder,
+        [NotNull] ISystemClock clock)
+        : base(options, logger, encoder, clock)
+    {
+    }
+
+    protected override async Task<AuthenticationTicket> CreateTicketAsync(
+        [NotNull] ClaimsIdentity identity,
+        [NotNull] AuthenticationProperties properties,
+        [NotNull] OAuthTokenResponse tokens)
+    {
+        var parameters = new Dictionary<string, string?>
+        {
+            ["token"] = tokens.AccessToken,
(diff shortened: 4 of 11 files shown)


samples/Mvc.Client/Startup.cs
```diff
@@ -44,8 +44,8 @@ public void ConfigureServices(IServiceCollection services)
 
         .AddTwitter(options =>
         {
-            options.ClientId = Configuration["Twitter:ClientId"];
-            options.ClientSecret = Configuration["Twitter:ClientSecret"];
+            options.ClientId = Configuration["Twitter:ConsumerKey"];
+            options.ClientSecret = Configuration["Twitter:ConsumerSecret"];
```
martincostello: I think the preferred fix for this would be to change the names in appsettings.json - then the settings match the property names.

src/AspNet.Security.OAuth.XiaoMi/AspNet.Security.OAuth.XiaoMi.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```
martincostello, line 4:
```suggestion
    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
    <PackageValidationBaselineVersion>6.0.7</PackageValidationBaselineVersion>
```

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+    {
+        var parameters = new Dictionary<string, string?>
+        {
+            ["token"] = tokens.AccessToken,
+            ["clientId"] = Options.ClientId
+        };
+
+        var address = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, parameters);
+
+        using var response = await Backchannel.GetAsync(address);
```
martincostello:
```suggestion
        using var response = await Backchannel.GetAsync(address, Context.RequestAborted);
```

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+            throw new HttpRequestException("An error occurred while retrieving user information.");
+        }
+
+        var json = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+        using var payload = JsonDocument.Parse(json);
+        var rootElement = payload.RootElement;
+
+        if (!rootElement.TryGetProperty("data", out JsonElement dataElement))
+        {
+            var errorCode = rootElement.GetProperty("code").GetString()!;
```
martincostello:
```suggestion
            var errorCode = rootElement.GetString("code")!;
```

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        var address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);
+
+        using var response = await Backchannel.GetAsync(address);
```
martincostello:
```suggestion
        using var response = await Backchannel.GetAsync(address, Context.RequestAborted);
```

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+        if (!response.IsSuccessStatusCode)
+        {
+            await Log.ExchangeCodeErrorAsync(Logger, response, Context.RequestAborted);
+            return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+        }
+
+        var json = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+        var payload = JsonDocument.Parse(json.Replace("&&&START&&&", string.Empty, StringComparison.OrdinalIgnoreCase));
```
martincostello: What's this `&&&START&&&` for?

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+        if (!string.IsNullOrEmpty(errorCode))
+        {
+            Log.ExchangeCodeErrorCode(Logger, errorCode, response.Headers.ToString(), json);
+            return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+        }
+
+        return OAuthTokenResponse.Success(payload);
+    }
+
+    protected override string BuildChallengeUrl([NotNull] AuthenticationProperties properties, [NotNull] string redirectUri)
```
martincostello: This implementation _seems_ to be the same as the one [in the base class](https://github.com/dotnet/aspnetcore/blob/d5f487900782d7d66fb106578577467ef9666e21/src/Security/Authentication/OAuth/src/OAuthHandler.cs#L291-L323) - is this actually needed?

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        parameters["state"] = Options.StateDataFormat.Protect(properties);
+
+        return QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, parameters);
+    }
+
+    /// <inheritdoc/>
+    protected override string FormatScope()
+        => FormatScope(Options.Scope); // TODO This override is the same as the base class' and can be removed in the next major version
+
```
martincostello:
```suggestion
```

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        ClaimsIssuer = XiaoMiAuthenticationDefaults.Issuer;
+        CallbackPath = XiaoMiAuthenticationDefaults.CallbackPath;
+
+        AuthorizationEndpoint = XiaoMiAuthenticationDefaults.AuthorizationEndpoint;
+        TokenEndpoint = XiaoMiAuthenticationDefaults.TokenEndpoint;
+        UserInformationEndpoint = XiaoMiAuthenticationDefaults.UserInformationEndpoint;
+
+        // Scope.Add("1");
+        // Scope.Add("3");
```
martincostello:
```suggestion
```

test/AspNet.Security.OAuth.Providers.Tests/XiaoMi/XiaoMiTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        var claims = await AuthenticateUserAsync(server);
+
+        // Assert
+        AssertClaim(claims, claimType, claimValue);
+    }
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task BuildChallengeUrl_Generates_Correct_Url(bool usePkce)
```
martincostello: If `BuildChallengeUrl` doesn't need to be overridden, then this whole test can be deleted.

test/AspNet.Security.OAuth.Providers.Tests/XiaoMi/XiaoMiTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        actual.ToString().ShouldStartWith("https://account.xiaomi.com/oauth2/authorize?");
+
+        var query = QueryHelpers.ParseQuery(actual.Query);
+
+        query.ShouldContainKey("state");
+        query.ShouldContainKeyAndValue("client_id", options.ClientId);
+        query.ShouldContainKeyAndValue("redirect_uri", redirectUrl);
+        query.ShouldContainKeyAndValue("response_type", "code");
+
+        // query.ShouldContainKeyAndValue("scope", "1,3");
```
martincostello: If `BuildChallengeUrl` _is_ needed, then remove this comment.

```suggestion
```

samples/Mvc.Client/Startup.cs
ArcherTrister: I think so too, but when I saw the custom prefix in the configuration file, I thought it was special, so I didn't change it like this.

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
ArcherTrister: You are right, It was my oversight, I will revise and resubmit.

samples/Mvc.Client/Startup.cs
martincostello: I think it's a copy-paste from the Facebook provider somewhere, as that has differently-named aliases.

martincostello:
Just one outstanding comment, otherwise this looks good: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/675#discussion_r865689455

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
ArcherTrister: This is required by [the documentation](https://dev.mi.com/console/doc/detail?pId=707), but I think it's bad.

martincostello: Wow 😶

martincostello: I agree that's also _not good_, however I think the best way to deal with it would be:

1. Do a `StartsWith()` and then trim it off if present. Then there's no danger (even if theoretical) of it replacing _other_ things in the payload that contain that string that shouldn't.
2. Add a comment explaining why the code works that way with a link to that documentation.

ArcherTrister: I agree too, I will do it, thank you!

kevinchalet:
Thanks for your contribution!

In Europe and in the USA, the "official" casing is generally "Xiaomi", not "XiaoMi". Any chance you could change that?

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
kevinchalet: Are we 100% sure it's not a typo in the docs? If you use Fiddler to intercept the request or put a breakpoint at line 99, do you see the `&&&START&&&` prefix in the returned JSON payload?

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
```diff
@@ -0,0 +1,166 @@
...
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        var address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);
+
+        using var response = await Backchannel.GetAsync(address, Context.RequestAborted);
```
kevinchalet, line 90: I see the documentation says the token endpoint uses GET, but have you tried using POST to see what it does?

ArcherTrister: Thanks, I will rename to Xiaomi, I didn't try to use post request, can you tell me the necessity of using it? I can try.

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationHandler.cs
ArcherTrister: @kevinchalet  Yes, I can be sure that the document is ok, when I debug I can see the prefix in the returned JSON payload.

image

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
```diff
@@ -0,0 +1,20 @@
+﻿/*
```
martincostello, line 1: I can't comment on file-names in git diffs, but you also need to fix the casing of the directory and file names.

martincostello (commented):
I just realised you also haven't updated the README to add the new provider to the table.

src/AspNet.Security.OAuth.XiaoMi/XiaoMiAuthenticationConstants.cs
martincostello: This is why the ubuntu build is failing due to the case-sensitive file system.

ArcherTrister: I confirmed that it was all renamed, but nothing changed after the commit, and I don't know why, so I decided to re-commit, and I will update the README to add the new provider to the table.

xiaomiprovider

martincostello: Windows is a case-insensitive file system, so just changing the case in the file names won't make them change in git.

You can achieve a rename either by renaming to _something else_ and then back to the intended casing, e.g. `XiaoMiAuthenticationOptions2` -> `XiaomiAuthenticationOptions`, or by using the `git mv` command, e.g. `git mv XiaoMiAuthenticationOptions.cs XiaomiAuthenticationOptions.cs`.

ArcherTrister: I know the first way but don't like it, I will try the second way in the future, thanks.

PR #677, opened 2022-05-10
Support trimming

martincostello:
I've been playing around with trimming in some applications this afternoon, and thought I'd try it out with the provider assemblies. Going through [the documentation](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming) this involved:

- Enabling the trimming analyzer.
- Marking shipping assemblies as trimmable.
- Fixing JSON serializer usage that is not supported by trimming.

Thoughts about this @kevinchalet? For the small amount  of work involved in getting it to build I thought it was at least worth pushing up a draft to seek an opinion. Note that I haven't tried this out in an actual real trimmed ASP.NET Core application yet.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -50,10 +50,12 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" $(RepoRelativeProjectDir.Contains('src')) ">
+    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <IncludeSymbols>true</IncludeSymbols>
     <IsPackable>true</IsPackable>
     <IsShipping>true</IsShipping>
+    <IsTrimmable>true</IsTrimmable>
     <Serviceable>false</Serviceable>
   </PropertyGroup>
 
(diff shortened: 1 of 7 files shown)


kevinchalet:
> Thoughts about this @kevinchalet? For the small amount of work involved in getting it to build I thought it was at least worth pushing up a draft to seek an opinion.

Given the modular nature of the aspnet-contrib providers, I'd expect the gain to be marginal as pretty much all the plumbing code is needed once you register a provider. That said, since the cost seems absurdly low (it's even too easy to be true 🤣), there's no reason not to do it.

AFAIK, the OAuth 2.0 base handler hasn't been updated yet to be linker-friendly and it's not clear to me whether it's something that will happen in 7.0. Would that affect our trimming story?

Tratcher:
> AFAIK, the OAuth 2.0 base handler hasn't been updated yet to be linker-friendly and it's not clear to me whether it's something that will happen in 7.0. Would that affect our trimming story?

It's on the list, we'll see how far we get. I agree that there's not a lot to trim here, you need most of the functionality.

martincostello:
Will come back to this in the future if there's a need to.

PR #681, opened 2022-05-23
Add issue template for new providers 

martincostello:
* Add a GitHub issue template for new providers (the bit it I'd like reviewed, the first commit).
* Tidy up the existing issue templates (use markdown headings and use comments for hints).

I've intentionally not gone down the path of trying to use GitHub issue forms at this point, I just did a quick pass on what we already have while adding the new template with some more focused questions and information.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_provider.md
+++ b/.github/ISSUE_TEMPLATE/new_provider.md
@@ -1,0 +1,45 @@
+---
+name: New provider
+about: Suggest a new OAuth provider for AspNet.Security.OAuth.Providers
+labels: new provider
+
+---
+
+### Which provider would you like support added for?
+
+<!--
+The name of the OAuth 2.0 provider you'd like to support added for.
+Providers that deviate significantly from the OAuth 2.0 specification
+with non-standard extensions and/or behavior are unlikely to be accepted.
+-->
+
+### Provider Documentation
+
+<!--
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/new_provider.md
```diff
@@ -0,0 +1,45 @@
...
+from to implement the provider code with.
+-->
+
+### Are you willing to implement this provider yourself?
+
+<!--
+Are you willing to implement this provider yourself and contribute it to the
+repository as a pull request?
+
+If you are not able or willing to implement this provider yourself, the repository
```
kevinchalet, line 30: :+1:

kevinchalet (approved):
Looks good.

> I've intentionally not gone down the path of trying to use GitHub issue forms at this point

Makes sense for this PR, but I guess we'll really want to use them at some point. It makes things look much nicer (e.g https://github.com/openiddict/openiddict-core/issues/new?assignees=&labels=bug&template=bug_report.yml). I guess we'll be able to re-use the OpenIddict new provider issue template once I add it 🤣

martincostello:
> Makes sense for this PR, but I guess we'll really want to use them at some point.

Yeah I haven't properly looked into them since they moved out of beta or properly tested out how they'd behave in the face of trying to file real issues with them.

I might circle back to that at some point before we ship 7.0.0.

PR #710, opened 2022-08-20
Add Kroger provider

lukemoore495:
Add Kroger provider.


src/AspNet.Security.OAuth.Kroger/AspNet.Security.OAuth.Kroger.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```
martincostello, line 6: These are needed for the project to build until the provider ships to NuGet.org.

```suggestion
    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
    <PackageValidationBaselineVersion>6.0.9</PackageValidationBaselineVersion>
    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```

src/AspNet.Security.OAuth.Kroger/KrogerAuthenticationOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        CallbackPath = KrogerAuthenticationDefaults.CallbackPath;
+
+        AuthorizationEndpoint = KrogerAuthenticationDefaults.AuthorizationEndpoint;
+        TokenEndpoint = KrogerAuthenticationDefaults.TokenEndpoint;
+        UserInformationEndpoint = KrogerAuthenticationDefaults.UserInformationEndpoint;
+
+        Scope.Add("profile.compact");
+        Scope.Add("cart.basic:write");
+        Scope.Add("product.compact");
```
martincostello: Are all three of these needed by default? Looking at [the documentation](https://developer.kroger.com/reference/#section/Authentication/CustomerContext) it seems like only `profile.compact` is required.

src/AspNet.Security.OAuth.Kroger/KrogerAuthenticationOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "id");
```
martincostello: Looking at the [documentation](https://developer.kroger.com/reference/#operation/profileGet), should this be `MapJsonSubKey(ClaimTypes.NameIdentifier, "data", "id")`?

test/AspNet.Security.OAuth.Providers.Tests/Kroger/bundle.json
```diff
@@ -0,0 +1,25 @@
...
+        "token_type": "bearer",
+        "expires_in": "1800"
+      }
+    },
+    {
+      "comment": "https://developer.kroger.com/reference#tag/Identity",
+      "uri": "https://api.kroger.com/v1/identity/profile",
+      "contentFormat": "json",
+      "contentJson": {
+        "id": "53990804-cfd1-43f3-8256-bdc9817a4fd0"
```
martincostello: The [example in the documentation](https://developer.kroger.com/reference/#operation/profileGet) doesn't look like the stub in the file, but like this:

```json
{
  "data": {
    "id": "53990804-cfd1-43f3-8256-bdc9817a4fd0"
  },
  "meta": {
    "pagination": {
      "total": 0,
      "start": 0,
      "limit": 0
    },
    "warnings": [
      "string"
    ]
  }
}
```

martincostello (commented):
Thanks for your contribution.

Have you tested this code with the real Kroger service? The discrepancies between the documentation and the code suggest not.

src/AspNet.Security.OAuth.Kroger/KrogerAuthenticationOptions.cs
lukemoore495: You are correct. I'll fix that.

lukemoore495:
> Thanks for your contribution.
> 
> Have you tested this code with the real Kroger service? The discrepancies between the documentation and the code suggest not.

I have tested it against the Kroger service, and it worked for my purposes. But the way I'm requesting scopes isn't right, so I'll change that and test some more. Thanks for the feedback.

lukemoore495:
I fixed the scopes and tested the code against the live Kroger service to make sure it's all working.

martincostello:
Thanks again for your contribution. This package is now available from [NuGet.org](https://www.nuget.org/packages/AspNet.Security.OAuth.Kroger/).

Issue #713, opened 2022-08-31
legacy apple login flow may leads to account compromised


royzhang666:
### Describe the bug

The current apple login flow may respond client's email, name in clear text, which can be intercepted and changed with the victim's accounts during the apple login process. The application developer have no way to authenticate its validity (considering it is neither encrypted nor protected with signature).

https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms

This library has a critical vulnerability that the untrusted email/names be added into the user claims which can lead to customer's account fully compromised.

### Steps To reproduce

1. create a new apple account
2. login to a 3rd party website with Apple OAuth authentication
3. intercept the apple login callback url: /signin-apple, change the email in user parameter to a victim's address
4. continue login, the attacker can pretend to be the victims account in 3rd part website.

### Expected behaviour

exclude the clear text `user` parameter in the callback, or provide an option to client to bypass this parameter.

### Actual behaviour

A cracker can disguise as victim's apple account.

### System information

all systems

### Additional context

https://developer.apple.com/forums/thread/118209

martincostello:
Please provide an example, including any relevant server responses, demonstrating this issue.

As far as I recall from implementing this, the callback is protected by a state parameter as well as a cookie for anti-forgery. Further, the connection is secured with HTTPS which would mitigate MitM scenarios where the response is tampered with.

Also, what do you mean by "legacy" in the issue's title?

kevinchalet:
@royzhang666 thanks for reporting this security issue 👍🏻 

> As far as I recall from implementing this, the callback is protected by a state parameter as well as a cookie for anti-forgery. Further, the connection is secured with HTTPS which would mitigate MitM scenarios where the response is tampered with.

`state` (combined with the antiforgery cookie created by `RemoteAuthenticationHandler`) protects against session fixation and CSRF attacks (i.e where the attacker tries to force the victim to log under an account controlled by the attacker).

[...]

PR #714, opened 2022-08-31
#713 add an option to allow application developer to ignore the untrusted …

royzhang666:
addresses https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/713

### DIFF
--- a/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
@@ -342,17 +342,6 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
             properties.StoreTokens(authTokens);
         }
 
-        if (!Options.IgnoreUserParameterInCallback && parameters.TryGetValue("user", out var userJson))
-        {
-            using var user = JsonDocument.Parse(userJson);
-            var userClaims = ExtractClaimsFromUser(user.RootElement);
-
-            foreach (var claim in userClaims)
-            {
-                identity.AddClaim(claim);
-            }
-        }
-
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
```diff
@@ -342,17 +342,6 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
-        {
-            using var user = JsonDocument.Parse(userJson);
-            var userClaims = ExtractClaimsFromUser(user.RootElement);
-
-            foreach (var claim in userClaims)
-            {
-                identity.AddClaim(claim);
-            }
-        }
-
```
martincostello, line 355: This breaks things in the case of the form post flow (the default) where the parameters aren't in the query string.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/4b6591108b5d576005a7115e244b28f34339183c/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs#L177-L181

martincostello: I think the proper way to remove this code would be to have the override just call the base method, and the duplicated implementation that supports query and form can just be deleted.

royzhang666: it is still valid, as it ignores the user field from the form post deliberately.

royzhang666: Indeed, remove the boilerplate code and reuse the base would be a good idea. As long as the apple specific design be removed.

martincostello:
Thanks for your PR - we're resolving this through #716 and #718.

PR #715, opened 2022-08-31
Ignore user parameter for non-POST requests

### LINKED ISSUE #713: legacy apple login flow may leads to account compromised

martincostello:
Resolves #713 as an alternative to #714 by achieving the same goal without adding a new option.

I had a quick look at _removing_ the other code path, but I think it needs a bit more thought regarding compatibility to remove it completely (just deleting it as I suggested here would break things completely https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/714#discussion_r959389922) so I thought this might be a bit more expedient. It's still a breaking change for anyone not using `form_post` anyway as they'll no longer get some of the claims set.

/cc @royzhang666

### DIFF
--- a/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
@@ -172,26 +172,26 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
     /// <inheritdoc />
     protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
     {
-        bool trustUserResponse;
+        bool trustUserParameter;
         IEnumerable<KeyValuePair<string, StringValues>> source;
 
         // If form_post was used, then read the parameters from the form rather than the query string
         if (string.Equals(Request.Method, HttpMethod.Post.Method, StringComparison.OrdinalIgnoreCase))
         {
-            trustUserResponse = true;
+            trustUserParameter = true;
             source = Request.Form;
         }
         else
         {
             // Do not trust the user query string parameter to
             // prevent an elevation of privilege vulnerability.
-            trustUserResponse = false;
+            trustUserParameter = false;
             source = Request.Query;
         }
 
         var parameters = new Dictionary<string, StringValues>(source);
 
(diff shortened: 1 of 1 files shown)


kevinchalet:
Hum, I’m not sure I understand why a POST callback wouldn’t be affected by the same vulnerability: in this attack, the user starting and completing the flow is the attacker himself, so it’s very easy to craft a POST callback request with a valid `state`/`code` and a forged `user` parameter, no?

Having a breaking change is always annoying, but it seems completely unavoidable, IMHO.

martincostello:
Maybe I'm just not fully understanding the vulnerability then 😅

martincostello:
Right I think I properly understand it now.

What we need to do is find a way to get the email server-side like how we get it for GitHub, rather than accepting it in the callback?

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/07be2eab296bb12989185649636ba96a281713f7/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationHandler.cs#L49-L61

martincostello:
I'm going to close this and use my sample site to inspect the token [as noted here](https://developer.apple.com/forums/thread/118209?answerId=366668022#366668022) and rework it to extract the relevant claims from the validated ID token instead.

PR #716, opened 2022-08-31
Extract email from ID token not user parameter

### LINKED ISSUE #713: legacy apple login flow may leads to account compromised

martincostello:
Use the verified ID token JWT as the source of the email claim rather than the user parameter in the callback for the Apple provider.

~This retains the functionality to take the user's _name_ from the `user` callback parameter still, but the email is no longer sourced from there and instead comes from the claims.~

~This should make the change backwards compatible.~

As a result of this change, the `ClaimTypes.GivenName` and `ClaimTypes.Surname` claims are no longer set.

Resolves #713.

/cc @royzhang666

### DIFF
--- a/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
@@ -160,6 +160,7 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
     /// <returns>
     /// An <see cref="IEnumerable{Claim}"/> containing the claims extracted from the user information.
     /// </returns>
+    [Obsolete("This method is obsolete and will be removed in a future version.")]
     protected virtual IEnumerable<Claim> ExtractClaimsFromUser([NotNull] JsonElement user)
     {
         var claims = new List<Claim>();
@@ -168,6 +169,11 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
         {
             claims.Add(new Claim(ClaimTypes.GivenName, name.GetString("firstName") ?? string.Empty, ClaimValueTypes.String, ClaimsIssuer));
             claims.Add(new Claim(ClaimTypes.Surname, name.GetString("lastName") ?? string.Empty, ClaimValueTypes.String, ClaimsIssuer));
+        }
+
+        if (user.TryGetProperty("email", out var email))
+        {
+            claims.Add(new Claim(ClaimTypes.Email, email.GetString() ?? string.Empty, ClaimValueTypes.String, ClaimsIssuer));
         }
 
         return claims;
@@ -344,17 +350,6 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
             }
 
             properties.StoreTokens(authTokens);
-        }
-
-        if (parameters.TryGetValue("user", out var userJson))
-        {
-            using var user = JsonDocument.Parse(userJson);
-            var userClaims = ExtractClaimsFromUser(user.RootElement);
-
-            foreach (var claim in userClaims)
-            {
-                identity.AddClaim(claim);
-            }
         }
 
         var ticket = await CreateTicketAsync(identity, properties, tokens);
--- a/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
@@ -133,10 +133,19 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
         {
             var securityToken = Options.SecurityTokenHandler.ReadJsonWebToken(token);
 
-            return new List<Claim>(securityToken.Claims)
+            var claims = new List<Claim>(securityToken.Claims)
             {
                 new Claim(ClaimTypes.NameIdentifier, securityToken.Subject, ClaimValueTypes.String, ClaimsIssuer),
             };
+
+            var emailClaim = claims.Find((p) => string.Equals(p.Type, "email", StringComparison.Ordinal));
+
+            if (emailClaim is not null)
+            {
+                claims.Add(new Claim(ClaimTypes.Email, emailClaim.Value ?? string.Empty, ClaimValueTypes.String, ClaimsIssuer));
+            }
+
+            return claims;
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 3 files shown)


martincostello:
I've also manually tested the change from a local build using my [test site](https://signinwithapple.martincostello.com/).

kevinchalet:
> This retains the functionality to take the user's _name_ from the `user` callback parameter still, but the email is no longer sourced from there and instead comes from the claims.

I guess the user name cannot be retrieved from the identity token? In any case, we should consider eliminating the remaining code that uses the unsafe `user` parameter, even if it means fewer claims will be returned.

martincostello:
I can remove it if you think it's best, but as I didn't think the user display name would be used for security comapred to the email (note it's the user's name, e.g. _Martin_, not the username, e.g. `martin@system.foo` or `martincostello`) it could be left to not completely remove the ability to fetch the name.

I'm not aware of any other way to get the user's given name and surname, but I'll double check that I didn't see it with my testing earlier just because my sample site wasn't configured to return it or something. Now I've typed this out, it might also be the case that you only ever get the given name and first name on the first log in anyway.

martincostello:
I checked, the name is configured to be returned by default:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/07be2eab296bb12989185649636ba96a281713f7/src/AspNet.Security.OAuth.Apple/AppleAuthenticationOptions.cs#L32-L34

And I've not changed the default in my sample repo ([code](https://github.com/martincostello/SignInWithAppleSample/blob/a27ded6b1d4b4b708a0cf3e6ef382fdd025d367e/src/SignInWithApple/AuthenticationModule.cs#L30-L54)), so it would get the name claim if provided.

I didn't have access to my own name at all in the ID token claims when I was testing earlier (just the claims in the new test I added to re-generate the tokens for the provider tests), so I think my cloudy recollection of being able to get the name on the first registration through Sign in with Apple for an application is correct.

kevinchalet:
Thanks for the detailed information 😃 

> I can remove it if you think it's best, but as I didn't think the user display name would be used for security comapred to the email (note it's the user's name, e.g. _Martin_, not the username, e.g. `martin@system.foo` or `martincostello`) it could be left to not completely remove the ability to fetch the name.

Yeah, let's remove it: we can't guarantee this is really the value that was returned by Apple and some developers may think it's protected against forged requests and use this data to (pre-)fill sensitive forms/tables.

Once we have a patch available, we should probably post a security advisory: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/security/advisories/new. If you agree, I'll work on a draft explaining what the issue consists in, what the actual impact is and the recommended actions: in a nutshell, most applications using the Apple provider the intended way should be safe (the `sub` claim is not affected), but applications using `ClaimTypes.Email` or `ClaimTypes.Name` for security decisions are at risk)

martincostello:
Cool, yep that sounds reasonable.

I’ll get the patch finished shortly with those changes, and then get that merged over to the 7 branch.

I was going to suggest doing the security advisory too, so you’re ahead of me on that one 😄. I guess I finish the patch and get that up on NuGet, and then we publish it?

Should I backport this for 3.1 too, as it’s still supported for a few months. Feels like we should.

kevinchalet:
> Should I backport this for 3.1 too, as it’s still supported for a few months. Feels like we should.

Yeah, definitely patch-worthy 😃 

I started creating the draft and opted for the CVSS approach. In this case, the resulting score will only depend on whether we decide to classify the email claim as a critical piece of information or not:

image

In the first case, the severity would be "high":

image

In the second case, the severity would be "moderate":

image

Any opinion on that? 😃

martincostello:
Cool - will do the 3.1 patch shortly, I'll just branch from the last release's branch to start that off.

Looking at the diagram, I feel we should err towards High.

If an application was using the `sub` as the primary key into their own system with regards to what account of theirs maps to the Apple account, then they'd probably be unaffected (or less affected?). But we don't really have any insight to what people are doing with that value, which makes me think we go high so people are more likely to pay attention and patch anyway?

martincostello:
I've saved a draft of the 6.0.10 release notes [here](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/releases/edit/untagged-4e186ff1c4cd6308471e) - I can publish that as-is if you think it's correct to release the fix, then we can add in the security advisory link afterwards.

kevinchalet:
> Looking at the diagram, I feel we should err towards High.
> 
> If an application was using the `sub` as the primary key into their own system with regards to what account of theirs maps to the Apple account, then they'd probably be unaffected (or less affected?). But we don't really have any insight to what people are doing with that value, which makes me think we go high so people are more likely to pay attention and patch anyway?

Exactly my feeling so let's do that 👍🏻 


> I've saved a draft of the 6.0.10 release notes [here](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/releases/edit/untagged-4e186ff1c4cd6308471e) - I can publish that as-is if you think it's correct to release the fix, then we can add in the security advisory link afterwards.

Looks good, thanks!

martincostello:
Cool - I'll get the 3.1 patch ready now, then when they're both ready to publish from the branches I'll publish the two releases to NuGet.org.

martincostello:
PR to patch 3.1 is #718. Release notes [here](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/releases/tag/untagged-1021a88c2cccbe44411b) (copy paste of 6.0.10's with a tweaked PR number).

martincostello:
As I've never done one before I'm not sure what the fields available are in the security advisory (I've seen there's ranges in reading reports I've seen via dependabot updates), but for clarity the issue applies to all published versions of the `AspNet.Security.OAuth.Apple` NuGet package except 3.1.8 and 6.0.10, which includes these version ranges:

* 2.2.0-2.22
* 3.0.0
* 3.1.0-3.1.7
* 5.0.0-5.0.18
* 6.0.0-6.0.9

I guess once we've published the packages and the advisor, there's also NuGet functionality we can use to flag the vulnerability in the older versions too?

martincostello:
Patches that resolve this issue have been published to NuGet.org:

* [3.1.8](https://www.nuget.org/packages/AspNet.Security.OAuth.Apple/3.1.8)
* [6.0.10](https://www.nuget.org/packages/AspNet.Security.OAuth.Apple/6.0.10)

PR #729, opened 2022-10-08
Optimize the standardization for remote authenticate query

LeaFrock:
...of Alipay&Wechat

This PR contains some small perf improvements,

* Avoid repeated `Request.Query` calls
* Replace the LINQ `ToDictionary` with native `List` (capacity is predicable) and `foreach`
* Shorten the standardized `QueryString` to reduce string allocations
* More comments and clearer method name for current workarounds

### DIFF
--- a/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
@@ -37,7 +37,10 @@ public partial class AlipayAuthenticationHandler : OAuthHandler<AlipayAuthentica
 
     protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
     {
-        StandardizeRemoteAuthenticateQuery(Request);
+        if (TryStandardizeRemoteAuthenticateQuery(Request.Query, out var queryString))
+        {
+            Request.QueryString = queryString;
+        }
 
         return base.HandleRemoteAuthenticateAsync();
     }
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -236,6 +230,36 @@ protected override string BuildChallengeUrl([NotNull] AuthenticationProperties p
         return QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, parameters);
     }
 
+    private static void StandardizeRemoteAuthenticateQuery(HttpRequest request)
```
martincostello: Rather than make this a void that operates on the whole request, how about just passing in `request.Query` and then returning the new one to use so the callsite would be something like this? Would then be a bit clearer that it's mutating the reference it's being passed.

```csharp
request.Query = TryStandardizeRemoteAuthenticateQuery(request.Query);
```

LeaFrock: The point is this line, `request.QueryString = QueryString.Create(queryParams);`

That's how we do the trick to change `Query` and why I have to pass `Request`.

martincostello: I get that - I'm suggesting assigning it in the caller, rather than having a void, and returning the new query in the new method.

```csharp
private static QueryString TryStandardizeRemoteAuthenticateQuery(QueryString queryString)
{
    if (...)
    {
        return QueryString.Create(queryParams);
    }

    return queryString;
}
```

LeaFrock: How about `private static bool TryStandardizeRemoteAuthenticateQuery(IQueryCollection query, out QueryString queryString)`? So the call is like the following,

```csharp
if (TryStandardizeRemoteAuthenticateQuery(Request.Query, out var queryString))
{
      Request.QueryString = queryString;
}
```

martincostello (approved):

martincostello:
Thanks!

PR #733, opened 2022-10-13
Update OAuth Zalo for user access token v4

huanbd:
https://developers.zalo.me/docs/api/social-api/tham-khao/user-access-token-post-4316

Change `AuthorizationEndpoint , TokenEndpoint` 
Change `HttpMethod `and `Parameter `in `ZaloAuthenticationHandler.ExchangeCodeAsync`

### DIFF
--- a/src/AspNet.Security.OAuth.Zalo/ZaloAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Zalo/ZaloAuthenticationHandler.cs
@@ -69,9 +69,9 @@ public partial class ZaloAuthenticationHandler : OAuthHandler<ZaloAuthentication
         var tokenRequestParameters = new Dictionary<string, string?>
         {
             ["app_id"] = Options.ClientId,
-            ["app_secret"] = Options.ClientSecret,
             ["code"] = context.Code,
             ["redirect_uri"] = context.RedirectUri,
+            ["grant_type"] = "authorization_code"
         };
 
         // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Zalo/ZaloAuthenticationHandler.cs
```diff
@@ -81,9 +81,14 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
             context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
         }
 
-        var address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);
+        using var request = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint)
+        {
+            Content = new FormUrlEncodedContent(tokenRequestParameters)
+        };
+
+        Backchannel.DefaultRequestHeaders.Clear();
```
martincostello, line 89: Why specifically is this needed?

kevinchalet: 👍🏻 
It's likely not a safe operation anyway: `Backchannel.DefaultRequestHeaders` is not thread-safe so trying to mutate it from a place that can be invoked concurrently is not a good option.

auvansangit: I can login with the customize as bellow:
```c#
var address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);

using var request = new HttpRequestMessage(HttpMethod.Post, address);
request.Headers.Add("secret_key", Options.ClientSecret);
```

But please also take a look at the scope that Zalo allows to access, they not allow to access email anymore @@.
The following are the list of scopes they allowed:
- id: string
- name: string
- picture: Json object 

```json
    "id": "UserId",
    "name": "User Name",
    "picture": {
          "data": {
                "url": "User avatar url"
          }
     }
```

martincostello:
Closing due to lack of activity and broken CI. Please open a new pull request if you would like to continue with this contribution.

PR #734, opened 2022-10-14
Update StackExchange provider

martincostello:
- Use the JSON token exchange endpoint.
- Use the domains given in the [API documentation](https://api.stackexchange.com/docs/authentication).
- Update to version 2.3 of the API.
- Align the provider with the request key being required.
- Validate the Site once, rather than per request.

See https://github.com/openiddict/openiddict-core/pull/1531#discussion_r994929629.


kevinchalet (approved):
Isn't all this red so satisfying? 😄

PR #735, opened 2022-10-14
Update Arcade

martincostello:
Update Arcade to the latest v7 version used by ASP.NET Core (as of commit [`d50065c`](https://github.com/dotnet/aspnetcore/commit/d50065c4a4fe31a66a1cc2e1a31896d30464da13) of the [`release/7.0` branch](https://github.com/dotnet/aspnetcore/tree/release/7.0)).


kevinchalet:
I see there are new `.sh` files in this updated Arcade version. Did you make sure they were marked as executable so Linux users don't have permission issues? (in case you're not familiar with this procedure: https://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows)

Thanks.

martincostello:
Nope, forgot to do that. Will sort out in a bit.

PR #739, opened 2022-10-24
Added flexible subdomain support for UserInfo endpoint. Updated docs …

### LINKED ISSUE #738: SuperOffice provider: UserInfo endpoint doesn't support multiple subdomains
### Describe the bug

Today, the provider expects the UserInfo endpoint data will come from one *environmental* subdomain, i.e.:

development: https://sod.superoffice.com
stage: https://quonline.superoffice.com.
production: https://online.superoffice.com.

SuperOffice CRM Online has introduced an API cluster that will publish API calls on different subdomains, i.e.:

https://**sod**.superoffice.com/
https://**sod1**.superoffice.com/
https://**sod2**.superoffice.com/

This affects how the UserInfo endpoint resolves and obtains profile-related details. The authorize and token endpoints are not affected.

A [PR](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/739) to fix this issue is in place.

SuperOfficeDevNet:
[…link.](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/738)

This PR enabled backwards compatibility with current and very near future changes with how the SuperOffice provider resolves UserInfo endpoint data.

This PR also updates README link to the docs.


src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationHandler.cs
```diff
@@ -109,7 +122,7 @@ public partial class SuperOfficeAuthenticationHandler : OAuthHandler<SuperOffice
             }
         }
 
-        return contextIdentifier;
+        return await Task.FromResult((contextIdentifier, webApiUrl));
```
martincostello: You should be able to just return this:

```suggestion
        return (contextIdentifier, webApiUrl);
```

src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationHandler.cs
```diff
@@ -38,13 +38,20 @@ public partial class SuperOfficeAuthenticationHandler : OAuthHandler<SuperOffice
     {
         var contextId = await ProcessIdTokenAndGetContactIdentifierAsync(tokens, properties, identity);
```
martincostello: It would be nicer to just deconstruct this:

```csharp
(string tenantId, string webApiUrl) = await ...
```

martincostello (commented):
Could you add another test that covers the case where the `webapi_url` claim is used?

I also wonder if we're missing some code coverage here, as I'm surprised the change isn't causing any tests to fail as the `webapi_url` claim is required yet there's no test changes and they're still all passing.

It also kind of feels to me that for this change to be backwards compatible, if the `webapi_url` claim isn't found, it should just automatically fall back to the previous behaviour, rather than throw.

src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationHandler.cs
SuperOfficeDevNet: The webapiurl claim has always been there, this is not what has changed. What has changed is written in the [issue](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/738). 

In short, the subdomain (sub) was *fixed*. Now it can be sub or sub2 or sub3, etc. Because of this change, and because the UserInfo endpoint is an API call rather than IAW the standard, the UserInfo endpoint request can be on one of a few subdomains.

The webapiurl claim is guaranteed to be correct, and therefore validating it doesn't make sense.

src/AspNet.Security.OAuth.SuperOffice/SuperOfficeAuthenticationHandler.cs
SuperOfficeDevNet: I like this suggestion! Thanks.

martincostello:
Regarding the `webapi_url` claim, I was referring to the fact that there wasn't a property for that [in the test bundle](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/3e256c8cd116794b925b5042f26c84509dd8b7d1/test/AspNet.Security.OAuth.Providers.Tests/SuperOffice/bundle.json#L12). However, if it's because it's encoded into the `id_token` property then that would explain it.

SuperOfficeDevNet:
> Regarding the `webapi_url` claim, I was referring to the fact that there wasn't a property for that [in the test bundle](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/3e256c8cd116794b925b5042f26c84509dd8b7d1/test/AspNet.Security.OAuth.Providers.Tests/SuperOffice/bundle.json#L12). However, if it's because it's encoded into the `id_token` property then that would explain it.

Yes, that's exactly correct. I added an exception if that claim doesn't get returned from the ProcessIdTokenAndGetContactIdentifierAsync method, but this should *never* happen. But if we every really do make that mistake, at least the exception message would signal just exactly what we did (or didn't do) to upset all of our partners.

SuperOfficeDevNet:
Hi! Any particular reason why this hasn't been merged yet? Seems to have been flags for past milestones, but still just a PR. Any update is appreciate.

martincostello:
At first it was an oversight, but now the `dev` branch targets .NET 7 so there needs to be some additional manual work to port it to a .NET 6 branch for you (as that's what you were originally targeting), and then I just forgot about if I'm honest.

I'll try and sort this out at some point this week. Sorry for the delay.

martincostello:
Sorry, but I've had to revert this change as the unit tests are [broken by it](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/actions/runs/3704462244/jobs/6277214552#step:5:244).

Unfortunately this wasn't spotted before merging due to https://github.com/dotnet/arcade/issues/11589 not causing the builds to report themselves as failed.

Please submit a new PR to re-apply this change including the appropriate changes to the unit tests so they continue to pass (please also run the tests locally when developing, they would have surfaced this issue before you even opened the pull request).

Once submitted, I'll then look at re-merging and porting to a branch to release a version for .NET 6 and publishing the 6.0.15 and 7.0.1 releases.

Apologies for the inconvenience.

kevinchalet:
Hopefully https://github.com/dotnet/arcade/pull/11869 will be eventually merged, but in the meantime, should we include the same fix in our local copy of Arcade? (if so, we'll need to remember not to undo this change when we update Arcade... 😢)

martincostello:
Sounds like a good idea 👍

PR #741, opened 2022-10-27
Add Smartsheet as a provider.

refractivSteve:
Add an OAuth 2.0 Smartsheet provider. https://www.smartsheet.com/


src/AspNet.Security.OAuth.Smartsheet/AspNet.Security.OAuth.Smartsheet.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
+    <PackageValidationBaselineVersion>6.0.7</PackageValidationBaselineVersion>
```
martincostello:
```suggestion
    <PackageValidationBaselineVersion>6.0.13</PackageValidationBaselineVersion>
```

docs/smartsheet.md
```diff
@@ -0,0 +1,23 @@
...
+
+```csharp
+services.AddAuthentication(options => /* Auth configuration */)
+        .AddSmartsheet(options =>
+        {
+            options.ClientId = "my-client-id";
+            options.ClientSecret = "my-client-secret";
+
+            // At least one scope must be passed
+            options.Scope.Add("READ_SHEETS");
```
martincostello, line 13: Is there such a thing as a "default" scope that everyone would always be expected to provide? If so, then it would be best to configure it as a sensible default in the options class, and then we don't need to add this additional documentation.

If there is no sensible default, then this is fine as-is.

src/AspNet.Security.OAuth.Smartsheet/SmartsheetAuthenticationOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+        ClaimActions.MapCustomJson(Claims.ProfileImage, user => GetProfileImageUri(user));
+    }
+
+    private static string GetProfileImageUri(JsonElement user)
+    {
+        var profileImageUri = user.EnumerateObject()
+            .Where(u => u.Name == "profileImage")
+            .Select(u => u.Value)
+            .Select(p => p.GetProperty("imageId"))
+            .Select(i => i.GetString()?.Replace("g:", string.Empty, StringComparison.CurrentCulture))
```
martincostello:
```suggestion
            .Select(i => i.GetString()?.Replace("g:", string.Empty, StringComparison.Ordinal))
```

src/AspNet.Security.OAuth.Smartsheet/SmartsheetAuthenticationOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+        ClaimActions.MapJsonKey(ClaimTypes.Email, "email");
+        ClaimActions.MapCustomJson(Claims.ProfileImage, user => GetProfileImageUri(user));
+    }
+
+    private static string GetProfileImageUri(JsonElement user)
+    {
+        var profileImageUri = user.EnumerateObject()
+            .Where(u => u.Name == "profileImage")
+            .Select(u => u.Value)
+            .Select(p => p.GetProperty("imageId"))
```
martincostello: Is there guaranteed to always be a property called `imageId` if `profileImage` is found?

martincostello (commented):
Can you add the new provider to the table in the README please? There's a template row you can copy in a comment at the bottom.

src/AspNet.Security.OAuth.Smartsheet/AspNet.Security.OAuth.Smartsheet.csproj
martincostello:
```suggestion
    <PackageValidationBaselineVersion>6.0.14</PackageValidationBaselineVersion>
```

docs/smartsheet.md
refractivSteve: There is no default scope specified and from the [list](https://smartsheet.redoc.ly/#section/Authentication/OAuth2) I don't see a sensible default either.

src/AspNet.Security.OAuth.Smartsheet/SmartsheetAuthenticationOptions.cs
refractivSteve: Good point.  I've refactored to ensure it doesn't error if it's missing.

martincostello:
Thanks for your contribution, this looks good.

Can you confirm that you've tested this provider with the real SmartSheet service and that it works as intended with a running application?

refractivSteve:
I use a couple of other providers in the same project I'm writing this provider for, so it's nice to be able to give back!

I was able to sign in with the provider and also view the sheets, as expected, so it is works as intended.

martincostello:
Awesome - thanks.

I'll look at merging this PR and getting the new provider published to NuGet.org later on today.

martincostello:
Thanks again for your contribution - the new provider is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.Smartsheet/

refractivSteve:
That's great, thanks!

PR #747, opened 2022-11-09
Use NuGet Central Package Management

kevinchalet:
CP(V)M, _deux ans après, le retour_ :trollface: 

With transitive pinning being configurable in the recent .NET SDKs, we can consider bring back CP(V)M in aspnet-contrib.

OpenIddict was updated yesterday (https://github.com/openiddict/openiddict-core/blob/dev/Directory.Packages.props) and I haven't seen anything suspicious.


Directory.Build.props
```diff
@@ -12,6 +12,8 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
     <DefineConstants Condition=" '$(Configuration)' == 'Debug' ">$(DefineConstants);JETBRAINS_ANNOTATIONS</DefineConstants>
     <Nullable>enable</Nullable>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
```
kevinchalet, line 16: `false` is supposed to be the default value, but we never know what may happen in future versions, so... 😄

martincostello (approved):

PR #750, opened 2022-12-15
Revert: Added flexible subdomain support for UserInfo endpoint

martincostello:
Reverts aspnet-contrib/AspNet.Security.OAuth.Providers#739 due to test failures that went undetected before of merge (see https://github.com/dotnet/arcade/issues/11589).


martincostello:
Manually verified in the logs that the tests are now passing again on all three platforms.

PR #764, opened 2023-03-12
Upgrade Autodesk to v2

### LINKED ISSUE #760: Autodesk v2
### Describe the bug
Autodesk published v2 for their OAuth infrastructure as shown here: https://aps.autodesk.com/en/docs/oauth/v2/reference/http/
I tried to update the provider in this tool here: https://github.com/EMostafaAli/AspNet.Security.OAuth.Providers/tree/autodesk_v2
It is a simple update, only changing the endpoints URL. However, when I tried to test I got the following error:
`Exception: OAuth token endpoint failure: invalid_request;Description=The request is missing a required parameter 'code_verifier'.`

I know for a fact that Autodesk application has PKCE set to false (for web apps) and as far as I know 'code_verifier' is related to PKCE.

### Steps To reproduce

- Go to https://aps.autodesk.com and create a new application (Standard web server)
- Clone my fork found in the following link: https://github.com/EMostafaAli/AspNet.Security.OAuth.Providers/tree/autodesk_v2 (it is a fork from this repo with links updated to Autodesk v2)
- Create a new web app with the following code:
```
builder.Services.AddAuthentication(options =>
    {
        options.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
        options.DefaultChallengeScheme = "oidc";
    })
    .AddCookie(CookieAuthenticationDefaults.AuthenticationScheme)
    .AddAutodesk("oidc", options =>
    {
        options.ClientId = "CLIENT_ID";
        options.ClientSecret = "CLIENT_SECRET";
        options.UsePkce = false;
        options.SaveTokens = true;
        options.Scope.Clear();
        options.Scope.Add("user-profile:read");
        options.Scope.Add("data:read");
    });
```

### Expected behaviour
Authorization should work.

### Actual behaviour
The authentication fails with the following:
image

### System information

- OS: Windows 10
- .NET version
.NET SDK:
 Version:   7.0.103
 Commit:    276c71d299

EMostafaAli:
Update Autodesk provider for v2 https://aps.autodesk.com/en/docs/oauth/v2/developers_guide/overview/
Issue: #760

### DIFF
--- a/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
@@ -51,7 +51,7 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
-    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(OAuthCodeExchangeContext context)
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
     {
         var tokenRequestParameters = new Dictionary<string, string>
         {
--- a/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
@@ -6,6 +6,7 @@
 
 using System.Net.Http.Headers;
 using System.Security.Claims;
+using System.Text;
 using System.Text.Encodings.Web;
 using System.Text.Json;
 using Microsoft.Extensions.Logging;
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(OAuthCodeExchangeContext context)
+    {
+        var tokenRequestParameters = new Dictionary<string, string>
+        {
+            { "redirect_uri", context.RedirectUri },
+            { "code", context.Code },
+            { "grant_type", "authorization_code" },
+        };
+
+        var isNotPKCE = true;
+
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add("client_id", Options.ClientId);
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+            isNotPKCE = false;
+        }
+
+        var requestContent = new FormUrlEncodedContent(tokenRequestParameters!);
+
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        if (isNotPKCE)
+        {
+            var credentials = Convert.ToBase64String(Encoding.ASCII.GetBytes(
+                $"{Options.ClientId}:{Options.ClientSecret}"));
+            requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Basic", credentials);
+        }
+
+        requestMessage.Content = requestContent;
+        requestMessage.Version = Backchannel.DefaultRequestVersion;
+        using var response = await Backchannel.SendAsync(requestMessage, Context.RequestAborted);
+        if (!response.IsSuccessStatusCode)
+        {
+            await Log.ExchangeCodeAsync(Logger, response, Context.RequestAborted);
+            return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+        }
+
+        var body = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+        return OAuthTokenResponse.Success(JsonDocument.Parse(body));
+    }
+
     private static partial class Log
     {
         internal static async Task UserProfileErrorAsync(ILogger logger, HttpResponseMessage response, CancellationToken cancellationToken)
@@ -63,6 +108,22 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
 
         [LoggerMessage(1, LogLevel.Error, "An error occurred while retrieving the user profile: the remote server returned a {Status} response with the following payload: {Headers} {Body}.")]
         private static partial void UserProfileError(
+            ILogger logger,
+            System.Net.HttpStatusCode status,
+            string headers,
+            string body);
+
+        internal static async Task ExchangeCodeAsync(ILogger logger, HttpResponseMessage response, CancellationToken cancellationToken)
+        {
+            ExchangeCodeAsync(
+                logger,
+                response.StatusCode,
+                response.Headers.ToString(),
+                await response.Content.ReadAsStringAsync(cancellationToken));
+        }
+
+        [LoggerMessage(2, LogLevel.Error, "An error occurred while retrieving an access token: the remote server returned a {Status} response with the following payload: {Headers} {Body}.")]
+        static partial void ExchangeCodeAsync(
             ILogger logger,
             System.Net.HttpStatusCode status,
             string headers,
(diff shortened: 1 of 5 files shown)


src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(OAuthCodeExchangeContext context)
+    {
+        var tokenRequestParameters = new Dictionary<string, string>
+        {
+            { "redirect_uri", context.RedirectUri },
+            { "code", context.Code },
+            { "grant_type", "authorization_code" },
```
martincostello:
```suggestion
            ["redirect_uri"] = context.RedirectUri,
            ["code"] = context.Code,
            ["grant_type"] = "authorization_code",
```

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
...
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add("client_id", Options.ClientId);
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+            isNotPKCE = false;
+        }
+
+        var requestContent = new FormUrlEncodedContent(tokenRequestParameters!);
```
martincostello:
```suggestion
        using var requestContent = new FormUrlEncodedContent(tokenRequestParameters!);
```

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
...
+        var isNotPKCE = true;
+
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add("client_id", Options.ClientId);
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+            isNotPKCE = false;
+        }
```
martincostello:
```suggestion
        var isPkce = false;

        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
        {
            tokenRequestParameters["client_id"] = Options.ClientId;
            tokenRequestParameters[OAuthConstants.CodeVerifierKey] = codeVerifier!;
            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
            isPkce = true;
        }
```

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
...
+
+        var requestContent = new FormUrlEncodedContent(tokenRequestParameters!);
+
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        if (isNotPKCE)
```
martincostello:
```suggestion
        if (!isPkce)
```

martincostello (commented):
Thanks for your contribution - just some minor code style comments.

Have you verified that these changes work as expected in a real application with the Autodesk v2 API?

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,50 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(OAuthCodeExchangeContext context)
```
kevinchalet: The logic used in this overload for PCKE implies that it is used as an alternative to client authentication (since the `client_secret` isn't used when PKCE is). It's traditionally a server implementation mistake as PKCE has never been a way to replace client authentication (both should be used at the same time for a confidential application).

Are we 100% sure Autodesk rejects token requests that specify both PKCE parameters and `client_secret` (in either the form or the `Authorization` header)?

EMostafaAli: Yes, their implementation won't allow both (according to my tests).

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
EMostafaAli: Yes, I have tested it.

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
kevinchalet: Hum, I'm intrigued: I just added Autodesk to OpenIddict so I gave it a try and the "confidential client" + PKCE combo worked just fine.

Are you sure it's not because you're sending the client credentials twice: once via the `client_id` parameter and once via the `Authorization` header? What kind of client did you create via the Autodesk portal?

EMostafaAli: I create Traditional Web App, and unless Autodesk changed something recently. I recall it didn't work for me when I created the related issue.

kevinchalet: I just tested my theory and here's the error I get when trying to send the client identifier in both the header and the body (which is illegal):

```json
{
  "error": "invalid_request",
  "error_description": "The 'client_id' is not supported in the request body when Authorization headers are present."
}
```

Try to remove the `tokenRequestParameters.Add("client_id", Options.ClientId);` line and give it another try.

kevinchalet:
Note: you forgot to update the address of the userinfo endpoint to use the v2 userinfo API. See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/788/files.

EMostafaAli:
Isn't it done as per the following screenshot?
image

kevinchalet:
@EMostafaAli I don't see this change in this PR:

image

Are you sure you pushed the changes to the correct branch?

(also, you'll likely need to update the claim mapping in the options class)

EMostafaAli:
My bad, I was looking at different code. I will try to make the changes quickly.

EMostafaAli:
@kevinchalet , updated the user info endpoint

kevinchalet:
@EMostafaAli it looks like you didn't fix the points mentioned by @martincostello. We'll also need you to fix https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/764#discussion_r1241017181 before being able to merge this PR.

EMostafaAli:
@kevinchalet, I hope this new commit addresses your comments.

martincostello:
The tests are failing.

EMostafaAli:
@martincostello, Done

src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationHandler.cs
```diff
@@ -50,6 +51,41 @@ public partial class AutodeskAuthenticationHandler : OAuthHandler<AutodeskAuthen
...
+        {
+            tokenRequestParameters[OAuthConstants.CodeVerifierKey] = codeVerifier!;
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        using var requestContent = new FormUrlEncodedContent(tokenRequestParameters!);
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        var credentials = Convert.ToBase64String(Encoding.ASCII.GetBytes(
+            $"{Options.ClientId}:{Options.ClientSecret}"));
```
kevinchalet, line 74: Technically, the standard way of computing the `Authorization: Basic [...]` header requires formURL-encoding the client identifier and client secret individually before computing their base64 representation. That said, I see in [the Autodesk docs that they don't mention formURL-encoding](https://aps.autodesk.com/en/docs/oauth/v2/tutorials/get-2-legged-token/#step-1-convert-client-id-and-secret-to-base64-encoded-string) so their implementation is probably not 100% compliant. Let's keep it this way.

kevinchalet (approved):
Thanks for your PR!

I'll leave @martincostello decide what to do next, as this PR is technically a breaking behavior change, given the claims are changing 😃

Issue #766, opened 2023-03-12
EpicGames Provider


10GeekJames:
### Which provider would you like support added for?
EpicGames
### Provider Documentation
https://dev.epicgames.com/docs/epic-account-services/auth/auth-interface

### Are you willing to implement this provider yourself?
Possibly

### Additional context
They have recently launched a developer program that will likely bring much more attention to this needed behavior

kevinchalet:
I added Epic Games to the OpenIddict client stack and I didn't hit any major issue. Two remarks tho' if you're interested in contributing:

  - Even if it's not explicitly mentioned, Epic Games supports OpenID Connect (but doesn't list `openid` as a supported scope in its configuration/discovery document) and exposes a standard userinfo endpoint.
  - The basic client authentication method is not standard: you can't formURL-encode the client credentials (as required by the specification) before base64-encoding them. And of course, they do include things like `/` or `+` in their client secrets... 🤣

PR #767, opened 2023-03-13
Extract user object from form body

pieperu:
Related to #766

This change pulls the User object from the Form on the Request sent from the apple Authorize endpoint.

This is then provided to the `OAuthCreatingTicketContext` so that the user properties can be retrieved using the ticket created event.

### DIFF
--- a/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
@@ -101,7 +101,9 @@ public partial class AppleAuthenticationHandler : OAuthHandler<AppleAuthenticati
 
         var principal = new ClaimsPrincipal(identity);
 
-        var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, tokens.Response.RootElement);
+        var userJson = ExtractUserFromRequestForm(Request.Form["user"].ToString());
+        var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, userJson);
+
         context.RunClaimActions();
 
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR 😃 

Unfortunately, as @martincostello said, we deliberately stopped using this unsafe property for security reasons and it's unlikely we'll accept a PR that will expose it (not to mention that's changing the object passed to the context constructor is a behavior breaking change).

Closing.

pieperu:
Hi @kevinchalet 

Thanks for checking this over so quickly. I didn't get to poke anyone and explain :)

My intention wasn't for this to be merged but to demonstrate some odd behaviour. 

Martin did mention that the raw payload would be available on the ticket created event, but upon checking the context, the user json object holds this format:
```
{
"access_token": string,
"expires_in”: string,
"id_token”: string,
“refresh_token”: string,
“token_type”: string
}
```
Instead of the User payload with the user's names which comes in the form.

Is this not misleading, whether you intend to surface it or not? Should it just be an empty JsonElement if you're not intending on exposing the expected user payload on this event/context?

pieperu:
In terms of breaking behaviour @kevinchalet, yes absolutely, although the exact same properties are available on the `Tokens` property so you are just passing the same information in 2 different constructor params currently...

pieperu:
So just to clarify @kevinchalet - If i wanted to access a user's GivenName and Surname, given that I only have 1 opportunity to capture it - I could not do so using this library, because you will never pass it on, even in the CreateTicket event?

If this is your stance on it, that's fair enough, but I need access to these values, so I just need to know if i need to fork or not :D

fipster-dev:
Martin cleared everything up for me. I totally missed that the HTTP context was available in the event handler so i can pluck it straight from there.

PR #774, opened 2023-04-16
Add CodeQL

martincostello:
Add a workflow to run CodeQL against the codebase.


martincostello:
@kevinchalet Happy with adding this?

kevinchalet:
I'm not super familiar with CodeQL but it sounds like a good idea 👍🏻

PR #778, opened 2023-04-23
Add Xiaomi provider

ArcherTrister:
Add an OAuth 2.0 Xiaomi provider.


docs/xiaomi.md
```diff
@@ -0,0 +1,24 @@
...
+## Example
+
+```csharp
+services.AddAuthentication(options => /* Auth configuration */)
+        .AddXiaomi(options =>
+        {
+            options.ClientId = "my-client-id";
+            options.ClientSecret = "my-client-secret";
+
+            // Optionally request the other open data interface permissions, if needed
```
martincostello:
```suggestion
            // Optionally request the other open data interface permissions, if needed.
```

src/AspNet.Security.OAuth.Xiaomi/AspNet.Security.OAuth.Xiaomi.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
+  </PropertyGroup>
+
+  <!-- TODO Remove once published to NuGet.org -->
+  <PropertyGroup>
+    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
```
martincostello, line 9:
```suggestion
    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
    <PackageValidationBaselineVersion>7.0.3</PackageValidationBaselineVersion>
```

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        var json = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+        using var payload = JsonDocument.Parse(json);
+        var rootElement = payload.RootElement;
+
+        if (!rootElement.TryGetProperty("data", out JsonElement dataElement))
+        {
+            var errorCode = rootElement.GetString("code")!;
+            throw new Exception($"An error (Code:{errorCode}) occurred while retrieving user information.");
```
martincostello:
```suggestion
            throw new Exception($"An error (Code: {errorCode}) occurred while retrieving user information.");
```

martincostello (commented):
Just a few minor changes, otherwise this seems OK to me.

martincostello (approved):

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+        };
+
+        // PKCE https://tools.ietf.org/html/rfc7636#section-4.5, see BuildChallengeUrl
+        if (context.Properties.Items.TryGetValue(OAuthConstants.CodeVerifierKey, out var codeVerifier))
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        var address = QueryHelpers.AddQueryString(Options.TokenEndpoint, tokenRequestParameters);
```
kevinchalet: Does that work if you send the parameters in the request form (formURL-encoding) instead of in the query string?

kevinchalet (commented):
Thanks for your PR!

(nice to see POST works and helps avoid a super ugly workaround 😄)

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
ArcherTrister: Yes, it works, and I've noticed that many of the existing projects also send parameters in query string, Are they tested to verify that they can't use form?

martincostello: We have over 90 providers now (in fact, this PR when merged will be the 95th), and don't have real accounts for many of them, so we don't have the time to go back over them and re-verify after they're added. Further changes usually come from user reports from broken functionality as the provider evolves, or incompatibilities added in new versions of ASP.NET Core that we need to address.

martincostello:
Before I merge and release this, just to confirm, you've verified the latest changes all work in a real running application integrated against the live production Xiamomi services?

ArcherTrister:
> Before I merge and release this, just to confirm, you've verified the latest changes all work in a real running application integrated against the live production Xiamomi services?

Yes, I have verified. Here is the running screenshot of sample.



f4a573f3b845352e971d1622ea80c40

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
ArcherTrister: Agree with.

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            await Log.ExchangeCodeErrorAsync(Logger, response, Context.RequestAborted);
+            return OAuthTokenResponse.Failed(new Exception("An error occurred while retrieving an access token."));
+        }
+
+        var json = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+
+        var payload = JsonDocument.Parse(json);
+
+        var errorCode = payload.RootElement.GetString("errcode");
```
kevinchalet: I see in the docs (https://dev.mi.com/console/doc/detail?pId=707) that errors returned from the token endpoint are expected to include the standard `error` parameter. Are we sure `errcode` is really returned?

ArcherTrister: Sorry, this is last year's code, I didn't revalidate, this should be I accidentally wrote the same code as the defined parameter name, I now try to change the request parameters to verify, it always returns an error with status code 400 (Use get or post requests), so I can't be sure if it returns errcode, and I think it may not return an error message with status code 200 after it fails the request(The code that gets the error code may not be executed), I now modify it to error as per the docs.

65ab7f466c1c682f8aad2f9d036676a

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
```diff
@@ -101,7 +101,7 @@ protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OA
 
         var payload = JsonDocument.Parse(json);
 
-        var errorCode = payload.RootElement.GetString("errcode");
+        var errorCode = payload.RootElement.GetString("error");
         if (!string.IsNullOrEmpty(errorCode))
         {
             Log.ExchangeCodeErrorCode(Logger, errorCode, response.Headers.ToString(), json);
```
martincostello: Let's rename the variable, the log method and update the log message template to reflect this change.

src/AspNet.Security.OAuth.Xiaomi/XiaomiAuthenticationHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        var principal = new ClaimsPrincipal(identity);
+        var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, dataElement);
+        context.RunClaimActions();
+
+        await Events.CreatingTicket(context);
+        return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
+    }
+
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync([NotNull] OAuthCodeExchangeContext context)
```
kevinchalet, line 71: Now that the logic here is standard, we should probably remove this overload and just use the default implementation provided by `OAuthHandler`.

PR #779, opened 2023-05-14
Change InstagramAuthenticationDefaults.cs  to InstagramDefaults.cs

AbdulrahmanXSO25:
I have error when using this package that I've Controllers with this name "InstagramAuthenticationDefaults", so I changed it to "InstagramDefaults"

### DIFF
--- a/src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.Instagram/InstagramAuthenticationDefaults.cs
@@ -9,7 +9,7 @@ namespace AspNet.Security.OAuth.Instagram;
 /// <summary>
 /// Default values used by the Instagram authentication middleware.
 /// </summary>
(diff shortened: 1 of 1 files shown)


kevinchalet:
Thanks for your PR.

Doing so would be a major breaking change for all the users of this package, so it's not an option.
Instead, if you have a class with the same name, consider renaming it or using aliases.

AbdulrahmanXSO25:
I've did this already!😅

But it was very boring for me to edit all the project, as the files is too many, but the right is that I can't modify the package classes' names as you said.

Thank you kevin!

Also, I've a mini question if you don't matter..
As this was my first pull request here, you said "PR" for pull request, is that correct?

kevinchalet:
> Also, I've a mini question if you don't matter..
As this was my first pull request here, you said "PR" for pull request, is that correct?

Yep, that's correct 😃

PR #785, opened 2023-06-18
Remove the Beam/Mixer and Gitter providers

### LINKED ISSUE #762: Consider removing/unlisting the Beam provider
It seems https://beam.pro/ no longer exists. If the service is indeed dead, we should consider removing the Beam provider from the repository and unlisting the NuGet package, as we previously did for other services.

@Schlurcher 👋🏻 you're the one who wrote the provider 7 years ago. Do you happen to know what happened to Beam?

### LINKED ISSUE #763: Determine whether the Gitter provider should be removed/unlisted
The authorization/token and userinfo endpoints used by the aspnet-contrib Gitter provider are no longer available, very likely as a result of the Matrix migration. Since Matrix doesn't seem to use/support OAuth 2.0/OpenID Connect, I suspect there'll be no way to update our provider to make it work with the new https://app.gitter.im/ service.

If so, we'll need to remove it and unlist the NuGet package.

kevinchalet:
Fixes https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/762 (Beam/Mixer was shut down by MSFT in 2020).
Fixes https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/763.

I'll unlist the packages on NuGet.org once this PR is merged.

@martincostello any preference for the milestone? No longer shipping a package is technically a breaking change (something we generally want to avoid for a patch version), but since the services are no longer operating... 😄


martincostello:
If they don't even work anymore, then I think SemVer goes out the window 🤣

kevinchalet:
That was my feeling too 😄

martincostello:
I'll sync this with the 8.0 branch tomorrow so there's no more previews published to MyGet either.

kevinchalet:
Thanks! :heart:

kevinchalet:
Packages deprecated on NuGet.org:
  - https://www.nuget.org/packages/AspNet.Security.OAuth.Beam/
  - https://www.nuget.org/packages/AspNet.Security.OAuth.Gitter/

martincostello:
> I'll sync this with the 8.0 branch tomorrow so there's no more previews published to MyGet either.

Done.

PR #788, opened 2023-06-23
Update Autodesk Auth to new v2 API endpoints

ClaysonIO:
Autodesk is deprecating their v1 Auth API endpoints, per the following blog post: https://aps.autodesk.com/blog/authentication-v2-and-deprecation-v1

This PR updates the URLs, and adjusts the claims to match the new userInfo API response, per the following documentation links: 
* https://aps.autodesk.com/en/docs/oauth/v2/reference/http/authorize-GET/
* https://aps.autodesk.com/en/docs/oauth/v2/reference/http/gettoken-POST/
* https://aps.autodesk.com/en/docs/oauth/v2/reference/http/userinfo-GET/

### DIFF
--- a/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.Autodesk/AutodeskAuthenticationDefaults.cs
@@ -34,15 +34,15 @@ public static class AutodeskAuthenticationDefaults
     /// <summary>
     /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
     /// </summary>
-    public static readonly string AuthorizationEndpoint = "https://developer.api.autodesk.com/authentication/v1/authorize";
+    public static readonly string AuthorizationEndpoint = "https://developer.api.autodesk.com/authentication/v2/authorize";
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


kevinchalet:
Thanks for your PR!

One thing: their documentation explicitly mentions that for v2 token requests, the client credentials must now be sent as part of the `Authorization` header (aka `client_secret_basic`), which also seems to be confirmed by the fact their OIDC discovery document doesn't list `client_secret_post` as a supported client authentication method: https://developer.api.autodesk.com/.well-known/openid-configuration

Did you have a chance to test against the real implementation? Are we sure `client_secret_post` still works?

kevinchalet:
I forgot we already had a PR open to update the Autodesk, sorry. Feel free to join the discussion: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/764.

kevinchalet:
> Are we sure `client_secret_post` still works?

For the record, I just gave it a try and it indeed doesn't work: `client_secret_post` is no longer supported:

```json
{
  "error": "invalid_credentials",
  "error_description": "The client credentials are invalid."
}
```

I see you work for Autodesk as a Senior Technical Consultant ; do you happen to know why this change was made?

PR #797, opened 2023-08-15
Added JumpCloud support

AaronSadlerUK:
Added support for the JumpCloud OIDC application.

https://jumpcloud.com/support/sso-with-oidc


AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
martincostello, line 1: All code files for the provider need to go under `/src` directory, not in the root.

docs/jumpcloud.md
```diff
@@ -0,0 +1,21 @@
...
+        });
+```
+
+## Required Additional Settings
+
+_None._
+
+## Optional Settings
+
+_None._
```
martincostello: `Domain` appears to be a new value (though not clear if required or not), so needs documenting.

AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
AaronSadlerUK: Apologies, I thought I created it under /src

That's moved now

docs/jumpcloud.md
AaronSadlerUK: I have added the Domain details to the docs

martincostello (commented):
This looks good so far. Please add tests by following an example of one of the existing providers that has a custom domain, such as Okta.

AaronSadlerUK:
I'm probably missing something simple but I'm getting the following error when I try to build:

`NU1101	Unable to find package AspNet.Security.OAuth.JumpCloud. No packages exist with this id in source(s): dotnet-eng, dotnet-tools, nuget	AspNet.Security.OAuth.JumpCloud`

Of course I know it's not found as the package hasn't been published yet, how do I get past this?

src/AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>$(DefaultNetCoreTargetFramework)</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>ASP.NET Core security middleware enabling JumpCloud authentication.</Description>
+    <Authors>AaronSadlerUK</Authors>
+    <PackageTags>jumpcloud;aspnetcore;authentication;oauth;security</PackageTags>
+  </PropertyGroup>
```
martincostello: From our [scaffolder](https://github.com/aspnet-contrib/generator-aspnet-oauth/blob/3366bb744ea69a2c76c921e2eb476b9ea625f522/generators/app/templates/Project.csproj#L3-L11):

```suggestion
  <PropertyGroup>
    <PackageValidationBaselineVersion>7.0.4</PackageValidationBaselineVersion>
    <TargetFramework>$(DefaultNetCoreTargetFramework)</TargetFramework>
  </PropertyGroup>
  
  <!-- This property group is required until the new provider is first published to NuGet.org -->
  <PropertyGroup>
    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
  </PropertyGroup>

  <PropertyGroup>
    <Description>ASP.NET Core security middleware enabling JumpCloud authentication.</Description>
    <Authors>AaronSadlerUK</Authors>
    <PackageTags>jumpcloud;aspnetcore;authentication;oauth;security</PackageTags>
  </PropertyGroup>
```

AaronSadlerUK:
I have removed the custom authentication as looking through the documentation with JumpCloud there is no mention of it.

I have added unit tests, also it looks like the domain is fixed and cannot be changed.

src/AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFramework>$(DefaultNetCoreTargetFramework)</TargetFramework>
```
martincostello, line 5: The baseline needs to be set to the release this provider will first ship as. Otherwise the baseline will default to `7.0.0`, and we wouldn't have such a package version for JumpCloud.

```suggestion
    <PackageValidationBaselineVersion>7.0.4</PackageValidationBaselineVersion>
    <TargetFramework>$(DefaultNetCoreTargetFramework)</TargetFramework>
```

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationExtensions.cs
```diff
@@ -68,7 +68,7 @@ public static AuthenticationBuilder AddJumpCloud([NotNull] this AuthenticationBu
     public static AuthenticationBuilder AddJumpCloud(
         [NotNull] this AuthenticationBuilder builder,
         [NotNull] string scheme,
-         string? caption,
+        [CanBeNull] string caption,
```
martincostello, line 71: If it can be null it should still be nullable (`string?`) too.

test/AspNet.Security.OAuth.Providers.Tests/JumpCloud/JumpCloudAuthenticationOptionsTests.cs
```diff
@@ -0,0 +1,77 @@
+﻿/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using AspNet.Security.OAuth.Okta;
```
martincostello: This shouldn't be here - you're testing Okta by mistake 😃

AaronSadlerUK: Ahh good spot!

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationExtensions.cs
AaronSadlerUK: This doesn't accept nullable types:
image

I have followed the Okta example for this

src/AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
AaronSadlerUK: Added the base version

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationExtensions.cs
martincostello: OK - I thought we'd turned them on, but I guess not!

src/AspNet.Security.OAuth.JumpCloud/AspNet.Security.OAuth.JumpCloud.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageValidationBaselineVersion>7.0.4</PackageValidationBaselineVersion>
+
```
martincostello:
```suggestion
```

martincostello (approved):

martincostello:
Before I merge this and prepare a release, can you confirm that you've tested this as working with the real JumpCloud service?

AaronSadlerUK:
Yes I have this working, happy to provide a URL privately

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+        ClaimActions.MapJsonKey(ClaimTypes.GivenName, "given_name");
+        ClaimActions.MapJsonKey(ClaimTypes.Name, "name");
+        ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "sub");
+        ClaimActions.MapJsonKey(ClaimTypes.Surname, "family_name");
+    }
+
+    /// <summary>
+    /// Gets or sets the JumpCloud domain (Org URL) to use for authentication.
+    /// </summary>
+    public string? Domain { get; set; }
```
kevinchalet, line 38: Does JumpCloud support configuring your own domain?
  - If not, we shouldn't make the domain configurable and remove this property.
  - If so, it would likely make sense to use `https://oauth.id.jumpcloud.com/` as the default domain when the user doesn't specify an explicit value.

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="JumpCloudAuthenticationOptions"/> class.
+    /// </summary>
+    public JumpCloudAuthenticationOptions()
+    {
+        ClaimsIssuer = JumpCloudAuthenticationDefaults.Issuer;
+        CallbackPath = JumpCloudAuthenticationDefaults.CallbackPath;
+
+        Scope.Add("openid");
+        Scope.Add("profile");
```
kevinchalet, line 25: In general, we try to limit the default scopes to the absolute required ones. For a standard-compliant OIDC provider, only `openid` should be required.

kevinchalet:
@martincostello I guess I'm late to the party, but that would be nice to address my two remarks before shipping a public version of this provider.

kevinchalet:
Well... scratch that I guess 😅 

image

martincostello:
Sorry, I was running the commands whilst also in a meeting so didn't see the notifications until just now.

martincostello:
We could always unlist 7.0.4 and then do a 7.0.5 that has the changes you want made?

kevinchalet:
> Sorry, I was running the commands whilst also in a meeting so didn't see the notifications until just now.

Is that the secret sauce of being a x10 developer? Doing things in parallel? 🤣 

> We could always unlist 7.0.4 and then do a 7.0.5 that has the changes you want made?

Doesn't seem necessary to me: depending on @AaronSadlerUK's feedback, we'll still be able to give `Domain` a default value in a future minor version (or deprecate it if the domain is not configurable in JumpCloud). For the scopes, well, it would be a breaking change, so we'll have to live with it 😄

martincostello:
It would technically be a breaking change, but it's only been published for ~5 minutes and I think some one would need to be a 100x developer to have already integrated it and be broken by it.

AaronSadlerUK:
I can't see anywhere that a custom domain is documented... However, that's not to say the enterprise subscribers can't do it

kevinchalet:
> I can't see anywhere that a custom domain is documented... However, that's not to say the enterprise subscribers can't do it

I guess it would be more clearly documented if it was possible (and AFAIK, JumpCloud is cloud-only and not offered as an on-premises product, so you can't even self-host it).

@martincostello I'm not sure it's worth pushing a new release now. We can integrate these changes in a future version and document them, after all. As you said, folks are unlikely to massively start using this provider anyway 😄

kevinchalet:
Note: JumpCloud fully supports PKCE (for both public and confidential apps) so we'll also want to set `UsePkce = true` for increased security.

src/AspNet.Security.OAuth.JumpCloud/JumpCloudAuthenticationOptions.cs
kevinchalet: I just tested and it works fine with only `openid`.

martincostello:
I'll make these changes now as I've just rebased the new provider into the v8 branch anyway.

AaronSadlerUK:
@martincostello would it be possible to PR into the v6 version?

The site I plan on using this with uses .NET 6

martincostello:
We don't typically backport new providers to older versions because it's additional maintenance for us.

AaronSadlerUK:
Ah ok, this site runs on Umbraco 10 (LTS) which uses .NET 6 (LTS)

PR #806, opened 2023-10-01
Use C# 12 features

martincostello:
- Use primary constructors where relevant.
- Use collection literals where relevant.
- Add helper method to reduce duplicative tests.
- Apply some IDE suggestions.
- Remove redundant using statements.
- Fix some typos.
- Hide shared files from the Visual Studio project explorer.

### DIFF
--- a/samples/Mvc.Client/Startup.cs
+++ b/samples/Mvc.Client/Startup.cs
@@ -11,8 +11,6 @@ namespace Mvc.Client;
 
 public class Startup(IConfiguration configuration, IHostEnvironment hostingEnvironment)
 {
-    public IConfiguration Configuration { get; } = configuration;
-
     public void ConfigureServices(IServiceCollection services)
     {
         services.AddRouting();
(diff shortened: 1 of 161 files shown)


samples/Mvc.Client/Startup.cs
```diff
@@ -9,17 +9,9 @@
...
-    public Startup(IConfiguration configuration, IHostEnvironment hostingEnvironment)
-    {
-        Configuration = configuration;
-        HostingEnvironment = hostingEnvironment;
-    }
-
-    public IConfiguration Configuration { get; }
-
-    private IHostEnvironment HostingEnvironment { get; }
+    public IConfiguration Configuration { get; } = configuration;
```
kevinchalet: Should this property be removed too for consistency with the `IHostEnvironment HostingEnvironment` property?

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
```diff
@@ -283,19 +275,19 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
 
         if (Options.SaveTokens)
         {
-            var authTokens = new List<AuthenticationToken>()
-            {
-                new AuthenticationToken() { Name = "access_token", Value = tokens.AccessToken },
-            };
+            List<AuthenticationToken> authTokens =
```
kevinchalet, line 278: Assuming we like the new syntax mostly for its terseness, maybe we should put everything on the same line when there's a single item added?

```csharp
List<AuthenticationToken> authTokens = [new() { Name = "access_token", Value = tokens.AccessToken }];

File: src/AspNet.Security.OAuth.Apple/Internal/DefaultAppleClientSecretGenerator.cs [hunk 3]
```diff
@@ -55,25 +42,25 @@ public override async Task<string> GenerateAsync([NotNull] AppleGenerateClientSe
 
     private static string CreateCacheKey(AppleAuthenticationOptions options)
     {
-        var segments = new[]
-        {
+        string[] segments =
```
kevinchalet: In this specific case, the older syntax seems better to me 😅

kevinchalet (approved):
👍🏻

samples/Mvc.Client/Startup.cs
martincostello: This one was left because it's public and the other was removed because it was private. I can make them consistent with each other if you like (either both private or both public).

src/AspNet.Security.OAuth.Apple/AppleAuthenticationHandler.cs
martincostello: I'm not going to change this one, at least for now, as doing so needs me to suppress SA1010 due to [this bug](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3687).

PR #813, opened 2023-11-07
Use unified BattleNet server

### LINKED ISSUE #812: Cannot register all BattleNet Provider Regions at the same time.
### Describe the bug

The BattleNet Provider has 5 regions and in general if you plan on supporting one you probably want to support all of them.
This is currently impossible, because registering the same provider twice, with different `authenticationScheme` and `displayName`, 
results in the following error, if you try to login with a region that was not the first registered one.

```
    AuthenticationFailureException: The oauth state was missing or invalid.
    Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler<TOptions>.HandleRequestAsync()
    Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
    Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)
```

I guess it is because 
```cs
    public static  Microsoft.Extensions.DependencyInjection.AuthenticationBuilder AddOAuth<TOptions, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] THandler>(this AuthenticationBuilder builder, string authenticationScheme, string displayName, Action<TOptions> configureOptions)
        where TOptions : OAuthOptions, new()
        where THandler : OAuthHandler<TOptions>
    {
        builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<TOptions>, OAuthPostConfigureOptions<TOptions, THandler>>());
        return builder.AddRemoteScheme<TOptions, THandler>(authenticationScheme, displayName, configureOptions);
    }
```
Calls `         builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<TOptions>, OAuthPostConfigureOptions<TOptions, THandler>>()); ` with the same types multiple times `TOptions == BattleNetAuthenticationOptions` and `THandler == BattleNetAuthenticationHandler`.
This results in only the first instance being correctly registered and working.

### Steps To reproduce

1. Create a template Blazor WASM Hosted .NET 7 app.
2. Change this
```cs
builder.Services.AddAuthentication()
    .AddIdentityServerJwt();
```
to this
```cs
builder.Services.AddAuthentication()
    .AddIdentityServerJwt()
    .AddBattleNet(
        BattleNetAuthenticationDefaults.AuthenticationScheme + BattleNetAuthenticationRegion.America,
        BattleNetAuthenticationDefaults.DisplayName + BattleNetAuthenticationRegion.America,
        options =>
        {
            options.Region = BattleNetAuthenticationRegion.America;
            options.ClientId = "id";
            options.ClientSecret = "secret";
            options.Scope.Add("oidc");
        })
    .AddBattleNet(
        BattleNetAuthenticationDefaults.AuthenticationScheme + BattleNetAuthenticationRegion.Europe,
        BattleNetAuthenticationDefaults.DisplayName + BattleNetAuthenticationRegion.Europe,
        options =>
        {
            options.Region = BattleNetAuthenticationRegion.Europe;
            options.ClientId = "id";
            options.ClientSecret = "secret";
            options.Scope.Add("oidc");
        });
```
Try to login with BattleNetEurope.

### Expected behaviour

You should be able to register all regions independently of each other.

### Actual behaviour

If you try to register multiple providers with different regions it crashes.

### System information

.NET SDK:
 Version:   7.0.403
 Commit:    142776d834

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.403\

Host:
  Version:      8.0.0-rc.2.23479.6
  Architecture: x64
  Commit:       0b25e38ad3

### Additional context
I tried creating a proxy type, that just inherits from `BattleNetAuthenticationHandler`
```cs
public class BattleNetAuthenticationHandlerAmerica: BattleNetAuthenticationHandler
{
    public BattleNetAuthenticationHandlerAmerica(IOptionsMonitor<BattleNetAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock) : base(options, logger, encoder, clock)
    {
    }
}
```

```cs
    .AddBattleNet<BattleNetAuthenticationOptions, BattleNetAuthenticationHandlerAmerica>(
        BattleNetAuthenticationDefaults.AuthenticationScheme + BattleNetAuthenticationRegion.America,
        BattleNetAuthenticationDefaults.DisplayName + BattleNetAuthenticationRegion.America,
        options =>
        {
            options.Region = BattleNetAuthenticationRegion.America;
            options.ClientId = "id";
            options.ClientSecret = "secret";
            options.Scope.Add("oidc");
        })
    .AddBattleNet
```

But this doesn't work, because `BattleNetAuthenticationOptions` is still the same and leads to the same error.
Creating a proxy of both doesn't work, because `BattleNetAuthenticationHandler` inherits `OAuthHandler<BattleNetAuthenticationOptions>` and `AddOAuth` requires one to be convertible to the other.

martincostello:
Implemented based on https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/812#issuecomment-1789607922.

Unless it's because of my office's DNS, I couldn't get `oauth.battlenet.com.cn` to resolve, so I left the URLs for China as they were.

I don't actually have a BattleNet account so I haven't tested this against the actual server.

Resolves #812.

### DIFF
--- a/src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
@@ -31,6 +31,22 @@ public static class BattleNetAuthenticationDefaults
     /// </summary>
     public static readonly string CallbackPath = "/signin-battlenet";
 
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+    /// </summary>
+    public static readonly string AuthorizationEndpoint = "https://oauth.battle.net/oauth/authorize";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+    /// </summary>
+    public static readonly string TokenEndpoint = "https://oauth.battle.net/oauth/token";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public static readonly string UserInformationEndpoint = "https://oauth.battle.net/oauth/userinfo";
+
+    [Obsolete("This class is obsolete and will be removed in a future release.")]
     public static class America
     {
         /// <summary>
@@ -49,6 +65,7 @@ public static class BattleNetAuthenticationDefaults
         public static readonly string UserInformationEndpoint = "https://us.battle.net/oauth/userinfo";
     }
 
+    [Obsolete("This class is obsolete and will be removed in a future release.")]
     public static class Europe
     {
         /// <summary>
@@ -67,6 +84,7 @@ public static class BattleNetAuthenticationDefaults
         public static readonly string UserInformationEndpoint = "https://eu.battle.net/oauth/userinfo";
     }
 
+    [Obsolete("This class is obsolete and will be removed in a future release.")]
     public static class Korea
     {
         /// <summary>
@@ -85,6 +103,7 @@ public static class BattleNetAuthenticationDefaults
         public static readonly string UserInformationEndpoint = "https://kr.battle.net/oauth/userinfo";
     }
 
+    [Obsolete("This class is obsolete and will be removed in a future release.")]
     public static class Taiwan
     {
         /// <summary>
--- a/src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
+++ b/src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
@@ -32,21 +32,29 @@ public static class BattleNetAuthenticationDefaults
     public static readonly string CallbackPath = "/signin-battlenet";
 
     /// <summary>
-    /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+    /// A class containing the URLs for the unified global BattleNet OAuth servers.
     /// </summary>
-    public static readonly string AuthorizationEndpoint = "https://oauth.battle.net/oauth/authorize";
+    public static class Unified
+    {
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+        /// </summary>
+        public static readonly string AuthorizationEndpoint = "https://oauth.battle.net/oauth/authorize";
+
+        /// <summary>
+        /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+        /// </summary>
+        public static readonly string TokenEndpoint = "https://oauth.battle.net/oauth/token";
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


kevinchalet:
> Unless it's because of my office's DNS, I couldn't get `oauth.battlenet.com.cn` to resolve, so I left the URLs for China as they were.

Yeah, I noticed the same thing when testing the OpenIddict implementation. It just seems this domain name - that has a CNAME entry that points to AWS China - is not resolvable everywhere: https://www.whatsmydns.net/#CNAME/oauth.battlenet.com.cn

src/AspNet.Security.OAuth.BattleNet/BattleNetAuthenticationDefaults.cs
```diff
@@ -31,6 +31,22 @@ public static class BattleNetAuthenticationDefaults
     /// </summary>
     public static readonly string CallbackPath = "/signin-battlenet";
 
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+    /// </summary>
+    public static readonly string AuthorizationEndpoint = "https://oauth.battle.net/oauth/authorize";
```
kevinchalet: Instead of adding new "constants" and deprecating the other ones, we should just use the existing regions to use `oauth.battle.net`? (just in case they'd split that again... 😄)

martincostello: I'll just make this internal and have all the others return this then.

To be honest, I'm not a fan of the existing implementation - a set-only property is weird.

kevinchalet: > To be honest, I'm not a fan of the existing implementation - a set-only property is weird.

Yeah. If we wanted to make it look less weird, we could probably convert it to an automatic getter+setter property and move the assignation to an `IPostConfigureOptions<BattleNetOptions>` implementation.

kevinchalet (approved):
Looks good 👍🏻 

(adding a new `Unified` constants class is fine, but maybe that would have been safer to just update the existing "constants" to use the new server address, in case they'd change the rules in the future and "de-unify" some of the regions?)

martincostello:
I figured it was better to let people just switch the region so if they de-split it again, then we don't need to change anything in the code at all, people just change their `Region` value back to one of the original ones. If we change all the field values and they de-unify, then we need to do a new release to change them back again.

I figured this was a good middle ground of back-compat and not preventing the use of the unified regions at all.

kevinchalet:
Yeah, it's a reasonable compromise 👍🏻

PR #814, opened 2023-11-08
Fix duplicated query string parameters for VSO/AzDo provider

martincostello:
Fix query string parameters being duplicated if `AuthorizationEndpoint` contains any user-added query string parameters.

If this change looks good, I can look at making a similar change to the Google provider for dotnet/aspnetcore#47054.

/cc @Tratcher

### DIFF
--- a/src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
@@ -9,6 +9,7 @@ using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
 using System.Text.Json;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.VisualStudio/VisualStudioAuthenticationHandler.cs
```diff
@@ -91,12 +91,17 @@ protected override string BuildChallengeUrl([NotNull] AuthenticationProperties p
...
-        var challengeUri = new Uri(challengeUrl, UriKind.Absolute);
+        var challengeUri = new UriBuilder(challengeUrl);
         var query = QueryHelpers.ParseQuery(challengeUri.Query);
 
         query["response_type"] = "Assertion";
 
-        return QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, query);
+        // Remove the query and re-add with the edit so that the parameters are not duplicated.
+        // See https://github.com/dotnet/aspnetcore/issues/47054 for more context.
+        challengeUri.Query = string.Empty;
```
Tratcher: Something like:
```suggestion
        challengeUri.Query = QueryString.Create(query);
        return callengeUri.Uri.AbsoluteUri;
```

PR #821, opened 2023-12-29
Update feishu provider claims.

vicenteyu:
Add new claims:
- Sub;
- Name;
- Picture;
- OpenId;
- EnName;
- TenantKey;
- AvatarUrl;
- AvatarThumb;
- AvatarMiddle;
- AvatarBig;
- Email;
- UserId;
- EmployeeNo;
- Mobile;


src/AspNet.Security.OAuth.Feishu/FeishuAuthenticationConstants.cs
```diff
@@ -13,7 +13,21 @@ public static class FeishuAuthenticationConstants
 {
     public static class Claims
     {
-        public const string UnionId = "urn:feishu:unionid";
+        public const string UnionId = "urn:feishu:union_id";
```
martincostello: Was this intentionally changed? Seems quite a breaking change on their part to rename claims.

src/AspNet.Security.OAuth.Feishu/FeishuAuthenticationConstants.cs
```diff
@@ -13,7 +13,21 @@ public static class FeishuAuthenticationConstants
...
+        public const string Sub = "urn:feishu:sub";
+        public const string Name = "urn:feishu:name";
+        public const string Picture = "urn:feishu:picture";
+        public const string OpenId = "urn:feishu:open_id";
+        public const string EnName = "urn:feishu:en_name";
```
martincostello: This constants should be named properly (`EnglishName`(?), `EmployeeNumber`, etc) rather than copy-paste their abbreviations.

martincostello (commented):
Do all these claims need to be added? We typically don't exhaustively add everything as users can trivially opt-in to map additional claims through configuration.

src/AspNet.Security.OAuth.Feishu/FeishuAuthenticationConstants.cs
vicenteyu: Maybe not change is better.

vicenteyu:
Adding more claims will make it easier to use.

martincostello:
I get that it makes it "easier", but it also means mapping more things that an integration might actually use.

I'm not convinced there's a _need_ to add all these additional mappings in the common/default case of just logging in a user and knowing their identity.

vicenteyu:
> I get that it makes it "easier", but it also means mapping more things that an integration might actually use.
> 
> I'm not convinced there's a _need_ to add all these additional mappings in the common/default case of just logging in a user and knowing their identity.

Maybe making it an expansion pack is a more appropriate way, thank you.

PR #826, opened 2024-02-23
Updating the Apple Sign in docs. Undocumented needed setting.

fgauna12:
Updated the docs in the Apple ID sign in. 

I was getting an exception that `ClientSecret` was required. 

After diving through the code, I found the hack that bypasses the need for ClientSecret when using a Key. 

`GenerateClientSecret` defaults to `false`.

### DIFF
--- a/docs/sign-in-with-apple.md
+++ b/docs/sign-in-with-apple.md
@@ -16,6 +16,7 @@ The provider comes with a built-in extension method `UsePrivateKey(string)` to g
 services.AddAuthentication(options => /* Auth configuration */)
         .AddApple(options =>
         {
+            options.GenerateClientSecret = true;
             options.ClientId = Configuration["AppleClientId"];
             options.KeyId = Configuration["AppleKeyId"];
             options.TeamId = Configuration["AppleTeamId"];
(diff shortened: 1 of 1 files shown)


martincostello:
Thanks for the changes.

It might make more sense to put the setting next to setting the key [like in this code here](https://github.com/martincostello/SignInWithAppleSample/blob/69c92a11f2431951cab442117bbc5871737c62ce/src/SignInWithApple/AppleAuthenticationOptionsExtensions.cs#L30-L35) so it's clearer they're related.

martincostello:
Closing due to inactivity.

piotrek-k:
Hi, could you please consider adding it to the docs?

I have also tried to follow this example:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/e145ee3b713ccdae8c5ffbad260e90aaf3b3081b/samples/Mvc.Client/Startup.cs#L51

But it didn't work since ClientSecret was not specified and adding `GenerateClientSecret` fixed it. Would be helpful to review those examples

martincostello:
A pull request with a suggestion of what to update is welcome.

PR #834, opened 2024-03-01
Support for the new LinkedIn API version format

### LINKED ISSUE #781: Add support of the new LinkedIn API version format
### Provider name

AspNet.Security.OAuth.LinkedIn

### Expected behavior

The LinkedIn provider is updated to support LinkedIn's new API version format that is introduced in June 2022. The old version (which is used now by the provider) will be shut down on 30 June 2023. Here is an official page: https://learn.microsoft.com/en-us/linkedin/marketing/versioning?trk=eml-mktg-cust-202305-global-api-migrations-june&src=e-drip&mcid=7057115242148806657&view=li-lms-2023-04

### Actual behavior

The LinkedIn provided uses the old API version (or I haven't found a new one in the repository). 

### Additional information

Accordingly to the documentation, it shouldn't take more than 15 minutes. I can also update it manually in my project by overriding Defatuls but it would be great to have a working NuGet package after 30 June 2023.

### LINKED ISSUE #900: LinkedIn Authentication Fails
### Describe the bug

Created a new LinkedIn App. The default OAuth 2.0 scopes are "openid", "profile", "email", "w_member_social".
AspNet.Security.OAuth.LinkedIn requests "r_emailaddress" and "r_liteprofile" which results in an LinkedIn server error.

### Steps To reproduce

Create a new LinkedIn app and try using it.

### Expected behaviour

To be authenticated

### Actual behaviour

LinkedIn error "Bummer".

### System information

AspNet.Security.OAuth.LinkedIn 8.0.0

softwareolatomiwa:
New Apps created on LinkedIn returns with Error 403 on getting user details using the /me endpoint. The endpoint has been changed and scope also has been changed to include openid, profile and email

ALso the format of user data returned has changed.

https://www.linkedin.com/developers/news/featured-updates/openid-connect-authentication

- Resolves #781
- Resolves #900


global.json
```diff
@@ -1,12 +1,12 @@
 {
   "sdk": {
-    "version": "8.0.201",
+    "version": "8.0.200",
```
martincostello: Please revert the changes in this file.

samples/Mvc.Client/Controllers/AuthenticationController.cs
```diff
@@ -33,7 +34,14 @@ public async Task<IActionResult> SignIn([FromForm] string provider)
         // Instruct the middleware corresponding to the requested external identity
         // provider to redirect the user agent to its own authorization endpoint.
         // Note: the authenticationScheme parameter must match the value configured in Startup.cs
-        return Challenge(new AuthenticationProperties { RedirectUri = "/" }, provider);
+        return Challenge(new AuthenticationProperties { RedirectUri = $"/signinAuth-{provider}" }, provider);
+    }
+
+    [HttpGet("/signinAuth-{authProvider}")]
+    public async Task<IActionResult> MyCallback(string authProvider)
```
martincostello: Why are these changes necessary?

samples/Mvc.Client/Startup.cs
```diff
@@ -46,16 +46,14 @@ public void ConfigureServices(IServiceCollection services)
             options.Scope.Add("user:email");
         })
 
-        /*
-        .AddApple(options =>
+        /*.AddApple(options =>
```
martincostello: The changes in this file look like leftovers from your local testing. Please revert them.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,9 +13,19 @@ public static class LinkedInAuthenticationConstants
 {
     public static class Claims
     {
-        public const string PictureUrl = "urn:linkedin:pictureurl";
+        public const string Picture = "picture";
```
martincostello, line 22: Renaming public/protected members is a breaking change. If they are no longer used/needed, they should be marked as obsolete instead.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -58,47 +48,10 @@ public partial class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthen
...
-            {
-                identity.AddClaim(new Claim(ClaimTypes.Email, email, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
-        }
-
         await Events.CreatingTicket(context);
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
-    protected virtual async Task<string?> GetEmailAsync([NotNull] OAuthTokenResponse tokens)
```
martincostello, line 75: Same here - removing this is a breaking change.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -25,19 +26,18 @@ public LinkedInAuthenticationOptions()
...
-        ClaimActions.MapCustomJson(Claims.PictureUrl, user => GetPictureUrls(user)?.LastOrDefault());
-        ClaimActions.MapCustomJson(Claims.PictureUrls, user =>
-        {
-            var urls = GetPictureUrls(user);
-            return urls == null ? null : string.Join(',', urls);
-        });
+        ClaimActions.MapJsonKey(ClaimTypes.Name, ProfileFields.Name);
+        ClaimActions.MapJsonKey(ClaimTypes.Email, ProfileFields.Email);
+        ClaimActions.MapJsonKey(ClaimTypes.GivenName, ProfileFields.GivenName);
+        ClaimActions.MapJsonKey(ClaimTypes.Surname, ProfileFields.FamilyName);
```
martincostello, line 37: The comments on various new constants you've added are noted as being multilocale strings, but then you're not using the code to handle them (in fact, you've completely deleted it).

Either the comments are wrong, or you're not handling the values correctly.

test/AspNet.Security.OAuth.Providers.Tests/Infrastructure/ApplicationFactory.cs
```diff
@@ -92,7 +92,7 @@ private static void ConfigureApplication<TOptions>(IApplicationBuilder app, OAut
            .UseEndpoints(endpoints =>
            {
                endpoints.MapGet(
-                   "/me",
+                   "/userinfo",
```
martincostello: Please revert this change, it is nothing specific to LinkedIn, the fact it is the same as their old URL is a coincidence.

test/AspNet.Security.OAuth.Providers.Tests/LinkedIn/LinkedInTests.cs
```diff
@@ -10,7 +10,7 @@ namespace AspNet.Security.OAuth.LinkedIn;
 
 public class LinkedInTests(ITestOutputHelper outputHelper) : OAuthTests<LinkedInAuthenticationOptions>(outputHelper)
 {
-    private Action<LinkedInAuthenticationOptions>? additionalConfiguration;
+    /*private Action<LinkedInAuthenticationOptions>? additionalConfiguration;*/
```
martincostello: Delete redundant code like this, don't leave it lying around in commented-out blocks.

test/AspNet.Security.OAuth.Providers.Tests/OAuthTests`1.cs
```diff
@@ -253,7 +253,7 @@ void ConfigureServices(IServiceCollection services)
         using (var client = server.CreateDefaultClient(LoopbackRedirectHandler))
         {
             // Act
-            using var result = await client.GetAsync("/me");
+            using var result = await client.GetAsync("/userinfo");
```
martincostello: Please revert the changes in this file. They are unnecessary as the URL exposed by the test fixture does not need to be changed.

martincostello (commented):
Thanks for your contribution.

This PR needs work before we can accept it due to the breaking nature of the changes to the library's public API surface and other changes to the test code that were not necessary.

test/AspNet.Security.OAuth.Providers.Tests/OAuthTests`1.cs
softwareolatomiwa: Thanks for the update.
The recommended changes have been made.

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -30,24 +46,47 @@ public static class ProfileFields
         /// The unique identifier for the given member. May also be referenced as the <c>personId</c> within a Person URN (<c>urn:li:person:{personId}</c>).
         /// The <c>id</c> is unique to your specific developer application. Any attempts to use the <c>id</c> with other developer applications will not succeed.
```
martincostello: Is this comment still accurate?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationHandler.cs
```diff
@@ -58,20 +48,11 @@ public partial class LinkedInAuthenticationHandler : OAuthHandler<LinkedInAuthen
...
-            {
-                identity.AddClaim(new Claim(ClaimTypes.Email, email, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
-        }
-
         await Events.CreatingTicket(context);
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
+    [Obsolete("GetEmailAsync is deprecated, no longer needed in the recent LinkedIn API changes.", false)]
```
martincostello:
```suggestion
    [Obsolete("This method is no longer used and will be removed in a future version.", false)]
```

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -13,9 +13,25 @@ public static class LinkedInAuthenticationConstants
 {
     public static class Claims
     {
+        [Obsolete("PictureUrls is deprecated, no longer conforms to data returned by recent API, please use Picture instead.", false)]
```
martincostello: Please change these messages to something like:

```
$"This constant is obsolete and is no longer used. It will be removed in a future version. Use {nameof(ReplacementName)} instead."
```

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -52,8 +51,8 @@ public LinkedInAuthenticationOptions()
     public ISet<string> Fields { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
martincostello, line 52: Looks like this property is now also unused?

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationConstants.cs
```diff
@@ -27,27 +43,54 @@ public static class Claims
...
+        [Obsolete($"This constant is obsolete and is no longer used. It will be removed in a future version. Use {nameof(Picture)} instead.", false)]
         public const string PictureUrl = "profilePicture(displayImage~:playableStreams)";
+
+        /// <summary>
+        /// Full name of the member.
+        /// </summary>
+        public const string Name = "name";
+
+        /// <summary>
+        /// Picture Url of the member.
```
martincostello:
```suggestion
        /// Picture URL of the member.
```

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -25,153 +26,21 @@ public LinkedInAuthenticationOptions()
...
-    /// <summary>
-    /// Gets or sets a <c>MultiLocaleString</c> resolver, a function which takes all localized values
-    /// and an eventual preferred locale from the member and returns the selected localized value.
-    /// The default implementation resolve it in this order:
-    /// 1. Returns the <c>preferredLocale</c> value if it is set and has a value.
-    /// 2. Returns the value corresponding to the <see cref="Thread.CurrentUICulture"/> if it exists.
-    /// 3. Returns the first value.
-    /// </summary>
-    /// <see cref="DefaultMultiLocaleStringResolver(IReadOnlyDictionary{string, string}, string?)"/>
-    public Func<IReadOnlyDictionary<string, string?>, string?, string?> MultiLocaleStringResolver { get; set; } = DefaultMultiLocaleStringResolver;
```
martincostello: As these are public, removing them is a breaking change - they should be marked as obsolete instead.

We should also retain the private `DefaultMultiLocaleStringResolver` method, as otherwise it's a breaking change of the behaviour of the `MultiLocaleStringResolver` property if someone is using it for another purpose.

softwareolatomiwa: Implemented the changes as advised

src/AspNet.Security.OAuth.LinkedIn/LinkedInAuthenticationOptions.cs
```diff
@@ -6,6 +6,7 @@
 
 using System.Security.Claims;
 using System.Text.Json;
+using Microsoft.Extensions.Options;
```
martincostello, line 9: Doesn't look like this is needed for anything?

martincostello (approved):

martincostello:
Can you also confirm that this all works as expected with the real live LinkedIn API?

kevinchalet:
Given it's technically a behavior breaking change - that requires subscribing to the new OIDC-based service - it would probably make more sense to bump the minor version (i.e 8.1 instead of 8.0.1). We'll also need to mention that clearly in the release notes.

softwareolatomiwa:
@martincostello I confirm it works as expected with the real live LinkedIn API

thang-d:
Hi @martincostello 
I think we should complete it ASAP; we are having trouble with LinkedIn authentication

martincostello:
> It would probably make more sense to bump the minor version (i.e 8.1 instead of 8.0.1).

@kevinchalet So do you want to re-version everything as 8.1.0 for releasing this?

kevinchalet:
> @kevinchalet So do you want to re-version everything as 8.1.0 for releasing this?

Yeah, that would be better.

In the same vein, we should probably bump the minor version instead of the patch version when introducing new features/providers: only bumping the patch version is something we had to do when ASP.NET Core had minor versions (as we wanted to have matching major+minor versions), but it's no longer necessary now that .NET never increases its minor version.

martincostello:
OK - once I merge the open PRs, I'll do a follow up to update `8.0.1` to `8.1.0` before doing the release, and then I'll do patch or minor going forwards depending on what's in the release.

martincostello:
Thanks for your contribution @softwareolatomiwa - these changes are now available in version 8.1.0.

PR #838, opened 2024-03-04
Update README.md to recommend using OpenIddict for new applications

kevinchalet:
With most of the pain points addressed in previous OpenIddict releases (e.g WS-Federation claims or authentication schemes mapping) and a lot of services now being supported (78 at the time of writing), the OpenIddict client and its web providers are a good alternative to the packages we maintain here.

As such, I'd like to update the README to encourage users to adopt them for new applications (while still making clear the aspnet-contrib providers are still supported, of course).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,16 @@
 
 # AspNet.Security.OAuth.Providers
 
-**AspNet.Security.OAuth.Providers** is a **collection of security middleware** that you can use in your **ASP.NET Core** application to support social authentication providers like **[GitHub](https://github.com/)**, **[Foursquare](https://foursquare.com/)** or **[Dropbox](https://www.dropbox.com/)**. It is directly inspired by **[Jerrie Pelser](https://github.com/jerriep)**'s initiative, **[Owin.Security.Providers](https://github.com/RockstarLabs/OwinOAuthProviders)**.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,42 @@
 
 # AspNet.Security.OAuth.Providers
 
-**AspNet.Security.OAuth.Providers** is a **collection of security middleware** that you can use in your **ASP.NET Core** application to support social authentication providers like **[GitHub](https://github.com/)**, **[Foursquare](https://foursquare.com/)** or **[Dropbox](https://www.dropbox.com/)**. It is directly inspired by **[Jerrie Pelser](https://github.com/jerriep)**'s initiative, **[Owin.Security.Providers](https://github.com/RockstarLabs/OwinOAuthProviders)**.
+**AspNet.Security.OAuth.Providers** is a **collection of security middleware** that you can use in your **ASP.NET Core** application to support social authentication providers like **[GitHub](https://github.com/)**, **[Twitter/X](https://twitter.com/)** or **[Dropbox](https://www.dropbox.com/)**. It is directly inspired by **[Jerrie Pelser](https://github.com/jerriep)**'s initiative, **[Owin.Security.Providers](https://github.com/RockstarLabs/OwinOAuthProviders)**.
```
kevinchalet, line 4: Note: I replaced Foursquare by Twitter because their new API no longer supports OAuth 2.0 (they now use API keys).

martincostello (approved):
My only comment is that it's _quite big_ for a "tip" - on my screen setup at least, it's the almost the entire height of the screen.

Maybe just have a smaller tip at the top, that then links to a section further down with the additional details?

kevinchalet:
> Maybe just have a smaller tip at the top, that then links to a section further down with the additional details?

Good idea! 👍🏻

martincostello:
Looks good 👍

kevinchalet:
Thanks!

PR #872, opened 2024-05-12
 Attest artifacts

martincostello:
Attest the binaries and packages from the build artifacts.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,6 +77,20 @@ jobs:
       if: ${{ runner.os != 'Windows' }}
       run: ./eng/common/cibuild.sh -configuration Release -prepareMachine
 
+    - name: Attest artifacts
+      uses: actions/attest-build-provenance@951c0c5f8e375ad4efad33405ab77f7ded2358e4 # v1.1.1
+      if: |
+        runner.os == 'Windows' &&
+        github.event.repository.fork == false &&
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -77,6 +77,20 @@ jobs:
...
+      if: |
+        runner.os == 'Windows' &&
+        github.event.repository.fork == false &&
+        (github.ref_name == github.event.repository.default_branch ||
+         startsWith(github.ref, 'refs/heads/dev') ||
+         startsWith(github.ref, 'refs/heads/rel/') ||
+         startsWith(github.ref, 'refs/tags/'))
+      with:
+        subject-path: |
+          ./artifacts/bin/AspNet.Security.OAuth.*/Release/**/AspNet.Security.OAuth.*.dll
```
kevinchalet, line 91: Do we need to be ultra specific here or a more generic filter like `./artifacts/bin/**/Release/**/*.dll` could work too?

martincostello: This avoids attesting anything that's a dependency that we didn't build ourselves and avoids the sample site too.

kevinchalet: Makes sense 👍🏻

martincostello:
I just discovered the hard way that there's a 64 file limit on attestations, so even if we only did the NuGet packages, that's still to many 😓

PR #895, opened 2024-06-12
Implement Airtable OAuth provider

denis-goncharenko:
This feature allows users to authenticate using their Airtable credentials.


denis-goncharenko:
Also, I've tested the Airtable provider against the production service and can confirm that it works fine.

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        [NotNull] ClaimsIdentity identity,
+        [NotNull] AuthenticationProperties properties,
+        [NotNull] OAuthTokenResponse tokens)
+    {
+        using var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+
+        using var response =
+            await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
```
martincostello:
```suggestion
        using var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
```

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        requestMessage.Headers.Accept.Add(
+            new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.FormUrlEncoded));
+        requestMessage.Content = new FormUrlEncodedContent(tokenRequestParameters);
+        var credentials = Convert.ToBase64String(Encoding.ASCII.GetBytes(
+            $"{Options.ClientId}:{Options.ClientSecret}"));
```
martincostello:
```suggestion
        requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.FormUrlEncoded));
        requestMessage.Content = new FormUrlEncodedContent(tokenRequestParameters);
        var credentials = Convert.ToBase64String(Encoding.ASCII.GetBytes($"{Options.ClientId}:{Options.ClientSecret}"));
```

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Basic", credentials);
+
+        requestMessage.Version = Backchannel.DefaultRequestVersion;
+
+        var response = await Backchannel.SendAsync(requestMessage, Context.RequestAborted);
+        var body = await response.Content.ReadAsStringAsync(Context.RequestAborted);
+
+        return response.IsSuccessStatusCode switch
+        {
+            true => OAuthTokenResponse.Success(JsonDocument.Parse(body)),
```
martincostello, line 84: Just for this provider (not all the others), would this benefit from using `ReadAsStream()` and `ParseAsync()` too (similar to the changes from #890)?

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        [LoggerMessage(3, LogLevel.Error, "An error occurred while retrieving an access token: the remote server returned a {Status} response with the following payload: {Headers} {Body}.")]
+        private static partial void ExchangeCodeError(
+            ILogger logger,
+            HttpStatusCode status,
+            string headers,
+            string body);
+
+        [LoggerMessage(1, LogLevel.Error,
+            "An error occurred while retrieving the user profile: the remote server returned a {Status} response with the following payload: {Headers} {Body}.")]
```
martincostello: Copy-paste artifact probably, but the IDs are 3 and then 1. Would make more sense to be 1 and 2.

martincostello:
```suggestion
        [LoggerMessage(1, LogLevel.Error, "An error occurred while retrieving the user profile: the remote server returned a {Status} response with the following payload: {Headers} {Body}.")]
```

martincostello (commented):
Just a few minor comments?

Are you planning on any other PRs to add more providers other than these 2, as if so I'll hold any release until they're all completed and merged.

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
kevinchalet: Note: it's not the standard flavor of the `client_secret_basic` client authentication, that expects the identifier and the secret to be percent-encoded first. Unless Airtable is known to implement the flawed version, we should probably use the standard version: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/cc5acd8e9838bd8a48f2403106b268d973bd3bc1/src/AspNet.Security.OAuth.Ebay/EbayAuthenticationHandler.cs#L92-L100

kevinchalet: Also, it seems weird to add formurl-encoded to the list of accepted content types when a JSON serializer is used to process the response. Does Airtable require doing that type of sorcery?

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
kevinchalet: Hum, that was a risky change: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/890#issuecomment-2163441183

denis-goncharenko:
> Just a few minor comments?
> 
> Are you planning on any other PRs to add more providers other than these 2, as if so I'll hold any release until they're all completed and merged.

I am going to add 2 more providers in one-two days.

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
denis-goncharenko: You are right, thanks. It was my fault for adding formurl-encoded to the list of accepted content types.

martincostello (approved):

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            tokenRequestParameters.Add(OAuthConstants.CodeVerifierKey, codeVerifier!);
+            context.Properties.Items.Remove(OAuthConstants.CodeVerifierKey);
+        }
+
+        using var requestMessage = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Json));
+        requestMessage.Content = new FormUrlEncodedContent(tokenRequestParameters);
+        requestMessage.Headers.Authorization = CreateAuthorizationHeader();
+        requestMessage.Version = Backchannel.DefaultRequestVersion;
```
kevinchalet, line 77: It's a good idea (we could also set the `VersionPolicy` property), but it looks like we haven't been very consistent regarding that as only 4 providers seem to set it.

@denis-goncharenko would you be interested in sending a PR to fix the inconsistency? (either by setting it for all the providers or by removing it for the 4 that currently use it, which would be less work)

kevinchalet (approved):
Thanks for your contribution! 👍🏻

src/AspNet.Security.OAuth.Airtable/AirtableAuthenticationHandler.cs
denis-goncharenko: I can do it. I will set it for all providers next week in a separate PR.

PR #907, opened 2024-06-15
Implement Docusign OAuth provider

denis-goncharenko:
This feature allows users to authenticate using their Docusign credentials. Tested in production service and works perfectly.


src/AspNet.Security.OAuth.Docusign/AspNet.Security.OAuth.Docusign.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageValidationBaselineVersion>8.1.0</PackageValidationBaselineVersion>
```
martincostello:
```suggestion
    <PackageValidationBaselineVersion>8.2.0</PackageValidationBaselineVersion>
```

kevinchalet:
> Tested in production service and works perfectly.

Isn't https://account-d.docusign.com/ DocuSign's dev environment?

We'll need to add an `Environment` option to allow users to change the environment without having to replace each endpoint manually to use https://account.docusign.com/ instead of https://account-d.docusign.com/.

denis-goncharenko:
> > Tested in production service and works perfectly.
> 
> Isn't https://account-d.docusign.com/ DocuSign's dev environment?
> 
> We'll need to add an `Environment` option to allow users to change the environment without having to replace each endpoint manually to use https://account.docusign.com/ instead of https://account-d.docusign.com/.

You are right. It's my fault I've just tested in prod and then in sandbox and forgot to change it. 

I'll add the 'Environment' option.

denis-goncharenko:
I've included the 'Environment' option to facilitate work with development and production environments. I have tested both options, and they work as expected.

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationConstants.cs
```diff
@@ -0,0 +1,22 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+namespace AspNet.Security.OAuth.Docusign;
+
+public static class DocusignAuthenticationConstants
```
martincostello: I would put this with the defaults, and also use readonly fields like we do everywhere else for the URLs. We've had providers change things in the past which can make constants problematic for endpoint URLs.

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+/// <summary>
+/// Defines a set of options used by <see cref="DocusignAuthenticationHandler"/>.
+/// </summary>
+public class DocusignAuthenticationOptions : OAuthOptions
+{
+    /// <summary>
+    /// Sets or retrieves a value that determines whether development or production endpoints are used.
+    /// The default value of this property is <see cref="DocusignAuthenticationEnvironment.Development"/>.
+    /// </summary>
+    public DocusignAuthenticationEnvironment Environment
```
martincostello: This could be a simple property, then we can use a post configure options to assign the properties (or throw if it's not one of the two expected values) for the URIs if not explicitly set. I forget which providers but there are existing examples of this pattern in the codebase.

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationDefaults.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public static readonly string DevelopmentUserInformationEndpoint = "https://account-d.docusign.com/oauth/userinfo";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public static readonly string ProductionUserInformationEndpoint = "https://account.docusign.com/oauth/userinfo";
```
martincostello: I wonder if it would be less copy-paste-y to have these just be the paths, and then use the post-configure to just switch the domain?

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace AspNet.Security.OAuth.Docusign;
+
+/// <summary>
+/// Defines a set of options used by <see cref="DocusignAuthenticationHandler"/>.
+/// </summary>
+public class DocusignAuthenticationOptions : OAuthOptions
+{
+    /// <summary>
+    /// Sets or retrieves a value that determines whether development or production endpoints are used.
```
martincostello:
```suggestion
    /// Gets or sets a value that determines whether development or production endpoints are used.
```

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// The default value of this property is <see cref="DocusignAuthenticationEnvironment.Development"/>.
```
martincostello: This feels like a foot-gun to me - I would have thought production being the default would be less-surprising/safer.

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public DocusignAuthenticationEnvironment Environment { get; set; }
```
martincostello, line 20: As this provider now has custom properties, it should have an entry in the docs folder to document it.

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationPostConfigureOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using Microsoft.Extensions.Options;
+
+namespace AspNet.Security.OAuth.Docusign;
+
+/// <summary>
+/// Used to setup defaults for all <see cref="DocusignAuthenticationOptions"/>.
```
martincostello:
```suggestion
/// Used to configure <see cref="DocusignAuthenticationOptions"/> instances.
```

test/AspNet.Security.OAuth.Providers.Tests/Docusign/DocusignTests.cs
martincostello, line 1: Could you add a test that validates the environment being changed uses the right URLs please?

src/AspNet.Security.OAuth.Docusign/DocusignAuthenticationDefaults.cs
martincostello: Similar example: [GitHub provider](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubPostConfigureOptions.cs)

martincostello:
Could you update the version for 8.2.0 here please as we're adding a new provider?

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/5319292ae8606ae2a9ae58afb56386c5e2257d82/eng/Versions.props#L5-L6

Otherwise LGTM.

martincostello (approved):

denis-goncharenko:
Hello @martincostello! When are you planning to create NuGet packages for version 8.2.0?

martincostello:
#910 is still WIP so assumed we'd do the release when that was finished too.

PR #910, opened 2024-06-24
Implement Zoho OAuth provider

denis-goncharenko:
The implementation allows users to authenticate using their Zoho account credentials.


src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationPostConfigureOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    private static void ConfigureEndpoints(ZohoAuthenticationOptions options)
+    {
+        var domain = GetDomain(options.Region);
+
+        options.AuthorizationEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.AuthorizationPath);
+        options.TokenEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.TokenPath);
+        options.UserInformationEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.UserInformationPath);
+    }
```
martincostello: I would change this to only override the values if they aren't already set - that way if a new region is added in the future, a user will be able to use it without us changing the provider and releasing a new version that supports it.

docs/zoho.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Example
+
+```csharp
+services.AddAuthentication(options => /* Auth configuration */)
+        .AddZoho(options =>
+        {
+            options.ClientId = "my-client-id";
+            options.ClientSecret = "my-client-secret";
+            options.Region = ZohoAuthenticationRegion.Global;
```
martincostello: If the default is `Global`, then this isn't actually required.

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationPostConfigureOptions.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return region switch
+        {
+            ZohoAuthenticationRegion.Global => "accounts.zoho.com",
+            ZohoAuthenticationRegion.Europe => "accounts.zoho.eu",
+            ZohoAuthenticationRegion.India => "accounts.zoho.in",
+            ZohoAuthenticationRegion.Australia => "accounts.zoho.com.au",
+            ZohoAuthenticationRegion.Japan => "accounts.zoho.jp",
+            ZohoAuthenticationRegion.Canada => "accounts.zohocloud.ca",
+            ZohoAuthenticationRegion.SaudiArabia => "accounts.zoho.sa",
```
martincostello: Nit: for easier readability, can you sort these alphabetically please?

test/AspNet.Security.OAuth.Providers.Tests/Zoho/ZohoAuthenticationPostConfigureOptionsTests.cs
```diff
@@ -0,0 +1,63 @@
...
+public static class ZohoAuthenticationPostConfigureOptionsTests
+{
+    [Theory]
+    [InlineData(ZohoAuthenticationRegion.Global, "accounts.zoho.com")]
+    [InlineData(ZohoAuthenticationRegion.Europe, "accounts.zoho.eu")]
+    [InlineData(ZohoAuthenticationRegion.India, "accounts.zoho.in")]
+    [InlineData(ZohoAuthenticationRegion.Australia, "accounts.zoho.com.au")]
+    [InlineData(ZohoAuthenticationRegion.Japan, "accounts.zoho.jp")]
+    [InlineData(ZohoAuthenticationRegion.Canada, "accounts.zohocloud.ca")]
+    [InlineData(ZohoAuthenticationRegion.SaudiArabia, "accounts.zoho.sa")]
```
martincostello: Same here for sorting.

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationDefaults.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+    /// </summary>
+    public static readonly string TokenPath = "/oauth/v2/token";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public static readonly string UserInformationPath = "/oauth/user/info";
```
martincostello, line 52: I can't find this in the documentation - where is it documented?

denis-goncharenko: I found it here https://www.zoho.com/accounts/protocol/oauth/use-access-token.html

martincostello: Is there any documentation for the response's format?

denis-goncharenko: I couldn't find the response format in the documentation, but I can provide you with an example.

docs/zoho.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Required Additional Settings
+
+| Property Name | Property Type                                                                                                                                                                                                       | Description                                       | Default Value                     |
+|:--------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:--------------------------------------------------|:----------------------------------|
+| `Region`      | [`ZohoAuthenticationRegion`](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationRegion.cs "ZohoAuthenticationRegion enumeration") | The target online region for Zoho authentication. | `ZohoAuthenticationRegion.Global` |
+
+## Optional Settings
+
+_None._
```
martincostello: Needs reworking to be optional not required.

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationPostConfigureOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        ConfigureEndpoints(options);
+    }
+
+    private static void ConfigureEndpoints(ZohoAuthenticationOptions options)
+    {
+        if (AreEndpointsInitialized(options))
+        {
+            return;
+        }
```
martincostello: Test each one on an individual basis, rather than as a group.

denis-goncharenko: What do you think if I add one more Region 'Custom'? And use it like:

```
        if (options.Region == ZohoAuthenticationRegion.Custom)
        {
            return;
        }

        var domain = GetDomain(options.Region);

        options.AuthorizationEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.AuthorizationPath);
        options.TokenEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.TokenPath);
        options.UserInformationEndpoint = CreateUrl(domain, ZohoAuthenticationDefaults.UserInformationPath);
```

then the user can use any future regions/endpoints.

martincostello: I would just guard the property assignments with a null check and not set them if they already have a value.

kevinchalet:
Did you try with a user account created in a non-US region?

I just added this provider to OpenIddict and it was fairly painful to support that, as you need to make the token and userinfo endpoints dynamic (based on the `location` parameter returned from the authorization endpoint). Given it doesn't seem to be implemented here, it's likely it requires a bit more work to really work.

denis-goncharenko:
> Did you try with a user account created in a non-US region?
> 
> I just added this provider to OpenIddict and it was fairly painful to support that, as you need to make the token and userinfo endpoints dynamic (based on the `location` parameter returned from the authorization endpoint). Given it doesn't seem to be implemented here, it's likely it requires a bit more work to really work.

Yes, I did. I have tested with almost all regions.

Each region that supports Zoho has its own domain (https://www.zoho.com/accounts/protocol/oauth/multi-dc.html). In my implementation, I configure all necessary endpoints in ZohoAuthenticationPostConfigureOptions.cs file and use hardcoded endpoints for selected region. Do you mean to use this endpoint (https://accounts.zoho.com/oauth/serverinfo) to receive locations or smth else?

kevinchalet:
> In my implementation, I configure all necessary endpoints in ZohoAuthenticationPostConfigureOptions.cs file and use hardcoded endpoints for selected region.

But they are configured statically/globally and are used for all users. Yet, if your application is configured to use the US region and you try to log in with a EU account, you'll get an error during the token request since the client was expected to send that token request to the EU token endpoint and not to the US token endpoint.

kevinchalet:
That (stupid) logic is explained here: https://www.zoho.com/accounts/protocol/oauth/multi-dc/client-authorization.html

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
```diff
@@ -51,6 +51,30 @@ public partial class ZohoAuthenticationHandler : OAuthHandler<ZohoAuthentication
...
+        if (!response.IsSuccessStatusCode)
+        {
+            await Log.ServerInfoErrorAsync(Logger, response, Context.RequestAborted);
+            throw new HttpRequestException("An error occurred while retrieving the server info.");
+        }
+
+        using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+
+        var domain = payload.RootElement.GetProperty("locations").GetProperty(location.ToString());
+        Options.UserInformationEndpoint = $"{domain}{ZohoAuthenticationDefaults.UserInformationPath}";
```
kevinchalet: You can't do that: the `Options` object is shared globally and `HandleRemoteAuthenticateAsync()` might be called concurrently (e.g if multiple Zoho authentication are handled concurrently).

(feel free to hardcode the endpoints to avoid an extra query: they are unlikely to change anyway)

denis-goncharenko:
Hey, @kevinchalet ! When you get a chance, could you take a look at this pull request?

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationDefaults.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="RemoteAuthenticationOptions.CallbackPath"/>.
+    /// </summary>
+    public static readonly string CallbackPath = "/signin-zoho";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.AuthorizationEndpoint"/>.
+    /// </summary>
+    public static readonly string AuthorizeEndpoint = "https://accounts.zoho.com/oauth/v2/auth";
```
kevinchalet: FYI, we generally name it `AuthorizationEndpoint` instead of `AuthorizeEndpoint`

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
```diff
@@ -0,0 +1,166 @@
...
+    private string CreateEndpoint(string path)
+    {
+        var location = Context.Request.Query["location"];
+
+        var domain = GetDomain(location.ToString());
+
+        return $"{domain}{path}";
+    }
+
+    private static string GetDomain(string location)
```
kevinchalet: Looks like `CreateEndpoint` is the only method that calls it. Maybe we can simplify the code to merge the two helpers?

Please also add a code comment indicating that we deliberately don't use the `accounts-server` parameter and that we fall back to the US region when the region isn't recognized for security reasons (blindly using `accounts-server` without checking it would allow stealing the client secret very easily). Just in case future us would forget this little detail 🤣

kevinchalet (approved):
> Hey, @kevinchalet ! When you get a chance, could you take a look at this pull request?

Hey,

Sorry for the late reply, busy weekend 🤣

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationDefaults.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public static readonly string UserInformationPath = "/oauth/user/info";
+
+    /// <summary>
+    /// Default value for Zoho Server Info.
+    /// </summary>
+    public static readonly string ServerInfoEndpoint = "https://accounts.zoho.com/oauth/serverinfo";
```
kevinchalet: Since we're now hardcoding the endpoints, I guess we can remove this "constant"?

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
```diff
@@ -0,0 +1,168 @@
...
+            "eu" => "https://accounts.zoho.eu",
+            "us" => "https://accounts.zoho.com",
+            "in" => "https://accounts.zoho.in",
+            "jp" => "https://accounts.zoho.jp",
+            "sa" => "https://accounts.zoho.sa",
+            "uk" => "https://accounts.zoho.uk",
+            _ => "https://accounts.zoho.com"
+        };
+
+        return $"{domain}{path}";
```
martincostello: Use a `UriBuilder` to do this not string concatenation. Then it will correctly deal with things like if the path does/doesn't start with a `/`.

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationDefaults.cs
martincostello: Could you provide that example?

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+
+        return OAuthTokenResponse.Success(payload);
+    }
+
+    /// <summary>
+    /// Creates the endpoint for the Zoho API using the location parameter.
+    /// If the location parameter doesn't match any of the supported locations, the default location (US) is used.
+    /// We don't use the accounts-server parameter due to security reasons.
```
martincostello: Could you elaborate (just here in GitHub is fine, not in the code) why that is?

```suggestion
    /// We don't use the <c>accounts-server</c> parameter for security reasons.
```

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
```diff
@@ -0,0 +1,175 @@
...
+        var domain = location.ToString().ToLowerInvariant() switch
+        {
+            "au" => "https://accounts.zoho.com.au",
+            "ca" => "https://accounts.zohocloud.ca",
+            "eu" => "https://accounts.zoho.eu",
+            "us" => "https://accounts.zoho.com",
+            "in" => "https://accounts.zoho.in",
+            "jp" => "https://accounts.zoho.jp",
+            "sa" => "https://accounts.zoho.sa",
+            "uk" => "https://accounts.zoho.uk",
```
martincostello, line 111: `gb` is often the ISO country code for the United Kingdom (yay, consistency). Is this definitely correct? It _may_ be useful to support both.

test/AspNet.Security.OAuth.Providers.Tests/Zoho/bundle.json
```diff
@@ -0,0 +1,44 @@
...
+          "au": "https://accounts.zoho.com.au",
+          "in": "https://accounts.zoho.in",
+          "jp": "https://accounts.zoho.jp",
+          "uk": "https://accounts.zoho.uk",
+          "us": "https://accounts.zoho.com",
+          "ca": "https://accounts.zohocloud.ca",
+          "sa": "https://accounts.zoho.sa"
+        }
+      }
+    }
```
martincostello: Is this used by anything?

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
denis-goncharenko: You are correct, but Zoho only uses these [codes](https://accounts.zoho.com/oauth/serverinfo). We don't manually set it up.

test/AspNet.Security.OAuth.Providers.Tests/Zoho/bundle.json
denis-goncharenko: No, it was used in a previous version of the Zoho implementation. I will remove it.

src/AspNet.Security.OAuth.Zoho/ZohoAuthenticationHandler.cs
denis-goncharenko: To prevent client's secrets from being stolen, this summary serves as a reminder of our purpose.

martincostello (approved):

martincostello:
Just to confirm, have you tested this works with the real service?

denis-goncharenko:
> Just to confirm, have you tested this works with the real service?

yep, I just did one more test and it works as expected.

martincostello:
Thanks again for your contribution. I'll look to do a 8.2.0 release for this and Docusign at some point this weekend.

martincostello:
The Docusign and Zoho providers are now both available from NuGet.org as part of the 8.2.0 release - thanks again!

PR #941, opened 2024-09-09
Update Arcade

martincostello:
- Annual update of .NET Arcade for .NET 9.
- Fix new analysis warnings.


Directory.Build.props
```diff
@@ -9,7 +9,6 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>portable</DebugType>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
kevinchalet, line 12: We may want to revert that as the lack of `snupkg` support in Arcade is specific to their assets publishing process and shouldn't affect us (and honestly, `snupkg` is better than any other format).

We can easily ignore the warning using `<SuppressSymbolPackageFormatValidation>true</SuppressSymbolPackageFormatValidation>`: https://github.com/openiddict/openiddict-core/blob/d4c05e2e236c9fe6ecc2c97570b1cde7fac1a8ea/Directory.Build.props#L16-L17

kevinchalet (approved):
Thanks! ❤️

Directory.Build.props
martincostello: I agree - I just changed it because it broke the build.

PR #962, opened 2024-10-20
Implement VK ID OAuth2 provider

hscokies:
Implement [VK ID](https://id.vk.com/about/business/go/docs/en/vkid/latest/vk-id/connection/start-integration/auth-without-sdk-web) OAuth2.


src/AspNet.Security.OAuth.VkId/VkIdAuthenticationErrors.cs
```diff
@@ -0,0 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+*/
+
+namespace AspNet.Security.OAuth.VkId;
+
+public static class VkIdAuthenticationErrors
```
martincostello: There's no need to expose these as constants - they can just be specified in the code where they're used.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace AspNet.Security.OAuth.VkId;
+
+/// <summary>
+/// Extension methods to add VK ID authentication capabilities to an HTTP application pipeline.
+/// </summary>
+public static class VkIdAuthenticationExtensions
+{
+    /// <summary>
+    /// Adds <see cref="VkIdAuthenticationHandler"/> to the specified
+    /// <see cref="AuthenticationBuilder"/>, which enables Vkontakte authentication capabilities.
```
martincostello: Should all these references to "Vkontakte" be "VK ID"?

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+
+namespace AspNet.Security.OAuth.VkId;
+
+public sealed class VkIdAuthenticationHandler : OAuthHandler<VkIdAuthenticationOptions>
+{
+    [Obsolete("Obsolete")]
+    public VkIdAuthenticationHandler(IOptionsMonitor<VkIdAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock)
+        : base(options, logger, encoder, clock)
+    {
+    }
```
martincostello: There's no point adding new members that are immediately obsolete.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+
+    protected override string BuildChallengeUrl(AuthenticationProperties properties, string redirectUri)
+    {
+        var parameter = Options.Scope;
+        var scopes = FormatScope(parameter);
+
+        var data = new byte[32];
+        RandomNumberGenerator.Fill((Span<byte>)data);
```
martincostello:
```suggestion
        var data  = RandomNumberGenerator.GetBytes(32);
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        var query = new Dictionary<string, string?>
+        {
+            { "response_type", "code" },
+            { "client_id", Options.ClientId },
+            { "scope", scopes },
+            { "redirect_uri", redirectUri },
+            { "state", Options.StateDataFormat.Protect(properties) },
+            { "code_challenge", WebEncoders.Base64UrlEncode(SHA256.HashData(Encoding.UTF8.GetBytes(codeVerifierKey))) },
+            { "code_challenge_method", OAuthConstants.CodeChallengeMethodS256 },
```
martincostello: `["foo"] = "bar"` style preferred for new code please.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+            }
+
+            if (!string.IsNullOrEmpty(tokens.TokenType))
+            {
+                tokensToStore.Add(new AuthenticationToken { Name = "token_type", Value = tokens.TokenType });
+            }
+
+            if (int.TryParse(tokens.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out var expiresIn))
+            {
+                var expiresAt = TimeProvider.GetUtcNow() + TimeSpan.FromSeconds(expiresIn);
```
martincostello:
```suggestion
                var expiresAt = TimeProvider.GetUtcNow().AddSeconds(expiresIn);
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        var identity = new ClaimsIdentity(ClaimsIssuer);
+        var ticket = await CreateTicketAsync(identity, properties, tokens);
+        return HandleRequestResult.Success(ticket);
+    }
+
+    protected override async Task<OAuthTokenResponse> ExchangeCodeAsync(OAuthCodeExchangeContext context)
+    {
+        if (!context.Properties.Items.TryGetValue(VkIdAuthenticationConstants.AuthenticationProperties.DeviceId,
+                out var deviceId) ||
```
martincostello:
```suggestion
        if (!context.Properties.Items.TryGetValue(VkIdAuthenticationConstants.AuthenticationProperties.DeviceId, out var deviceId) ||
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+            { "state", Options.StateDataFormat.Protect(context.Properties) },
+        };
+
+        using var request = new HttpRequestMessage(HttpMethod.Post, Options.TokenEndpoint);
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        request.Content = new FormUrlEncodedContent(query);
+        request.Version = Backchannel.DefaultRequestVersion;
+
+        var response = await Backchannel.SendAsync(request, Context.RequestAborted);
+        var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
```
martincostello:
```suggestion
        var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+        var errorCode = errorElement.GetString()!;
+        var errorDescription = errorElement.GetProperty("error_description").GetString()!;
+        return OAuthTokenResponse.Failed(new Exception($"{errorCode}: {errorDescription}"));
+    }
+
+    protected override async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
+    {
+        var query = new Dictionary<string, string>
+        {
+            { "access_token", tokens.AccessToken! }, { "client_id", Options.ClientId }
```
martincostello: +1 on style here, also one key assignment per line.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,221 @@
...
+        };
+        using var request = new HttpRequestMessage(HttpMethod.Post, Options.UserInformationEndpoint);
+        request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        request.Content = new FormUrlEncodedContent(query);
+        request.Version = Backchannel.DefaultRequestVersion;
+
+        var response = await Backchannel.SendAsync(request, Context.RequestAborted);
+        var content = await response.Content.ReadAsStringAsync();
```
martincostello:
```suggestion
        var content = await response.Content.ReadAsStringAsync(Context.RequestAborted);
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationScopes.cs
```diff
@@ -0,0 +1,98 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
+ * See https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+namespace AspNet.Security.OAuth.VkId;
+
+/// <summary>
+/// List of <see href="https://id.vk.com/about/business/go/docs/vkid/latest/vk-id/connection/work-with-user-info/scopes"> scopes </see>
```
martincostello:
```suggestion
/// List of <see href="https://id.vk.com/about/business/go/docs/vkid/latest/vk-id/connection/work-with-user-info/scopes">scopes</see>.
```

src/AspNet.Security.OAuth.VkId/AspNet.Security.OAuth.VkId.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```
martincostello, line 5:
```suggestion
    <PackageValidationBaselineVersion>8.3.0</PackageValidationBaselineVersion>
    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```

martincostello (commented):
Provider needs adding to README.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationExtensions.cs
hscokies: Not sure 😅
VK ID = OAuth2.1 vkontakte implementation

test/AspNet.Security.OAuth.Providers.Tests/VkId/VkIdTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public override string DefaultScheme => VkIdAuthenticationDefaults.AuthenticationScheme;
+
+    protected internal override void RegisterAuthentication(AuthenticationBuilder builder)
+    {
+        builder.AddVkId(options => ConfigureDefaults(builder, options));
+    }
+
+    [Theory]
+
```
martincostello:
```suggestion
```

martincostello (commented):
(I see the PR changes, but I'm on vacation until next Wednesday so my ability to properly review/merge will be sporadic until then)

README.md
```diff
@@ -152,107 +152,108 @@ Documentation for the providers' settings can be found [here](docs/README.md "Pr
 
 If a provider you're looking for does not exist, consider making a PR to add one.
 
-| Provider | Stable | Nightly | Documentation |
```
martincostello: Please revert all these formatting changes and just add the new line at the appropriate point in the A-Z order. Otherwise it's just ongoing maintenance to adjust for alignment as new providers are added.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationDefaults.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.TokenEndpoint"/>.
+    /// </summary>
+    public const string TokenEndpoint = "https://id.vk.com/oauth2/auth";
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/>.
+    /// </summary>
+    public const string UserInformationEndpoint = "https://id.vk.com/oauth2/user_info";
```
martincostello: These should be `static readonly` like in our other providers so that if they ever need to change (there is past precedent for other providers), the values aren't "burned in" to any call sites:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/82d20e3a415be2203257ecc1222edde5766961bb/src/AspNet.Security.OAuth.Amazon/AmazonAuthenticationDefaults.cs#L34-L47

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationScopes.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace AspNet.Security.OAuth.VkId;
+
+/// <summary>
+/// List of <see href="https://id.vk.com/about/business/go/docs/vkid/latest/vk-id/connection/work-with-user-info/scopes"> scopes </see>.
+/// </summary>
+public static class VkIdAuthenticationScopes
+{
+    /// <summary>
+    /// Grants access to first and last name, gender, profile photo, date of birth.
```
martincostello: If we make it less specific, then if new fields are added to their API our documentation isn't wrong.

```suggestion
    /// Grants access to personal information.
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationExtensions.cs
martincostello: The reason I ask is because it's referred to as "VK ID" elsewhere in the XML documentation comments.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,247 @@
...
+using Microsoft.Extensions.Primitives;
+using Base64UrlEncoder = Microsoft.AspNetCore.Authentication.Base64UrlTextEncoder;
+
+namespace AspNet.Security.OAuth.VkId;
+
+public sealed class VkIdAuthenticationHandler(
+    IOptionsMonitor<VkIdAuthenticationOptions> options,
+    ILoggerFactory logger,
+    UrlEncoder encoder)
+    : OAuthHandler<VkIdAuthenticationOptions>(options, logger, encoder)
```
martincostello: I can't find the comment where it was originally stated, but please use a non-primary constructor for consistency with the rest of the codebase (adopting them was suggested once, but rejected).

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationExtensions.cs
hscokies: Yea, changed it to VK ID to not confuse people

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -19,12 +19,16 @@
...
-    IOptionsMonitor<VkIdAuthenticationOptions> options,
-    ILoggerFactory logger,
-    UrlEncoder encoder)
-    : OAuthHandler<VkIdAuthenticationOptions>(options, logger, encoder)
+public sealed class VkIdAuthenticationHandler : OAuthHandler<VkIdAuthenticationOptions>
 {
+    public VkIdAuthenticationHandler(
+        IOptionsMonitor<VkIdAuthenticationOptions> options,
+        ILoggerFactory logger,
+        UrlEncoder encoder)
```
martincostello: To match the other implementations.

```suggestion
        IOptionsMonitor<VkIdAuthenticationOptions> options,
        ILoggerFactory logger,
        UrlEncoder encoder)
```
```suggestion
        [NotNull] IOptionsMonitor<VkIdAuthenticationOptions> options,
        [NotNull] ILoggerFactory logger,
        [NotNull] UrlEncoder encoder)
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationOptions.cs
martincostello, line 1: Looking at the documentation, PKCE is supported, so `UsePkce` should be defaulted to `true` and the appropriate changes made to `ExchangeCodeAsync()` in the handler (search PKCE in the existing handlers to see the sort of changes needed).

hscokies: Looks like it's mandatory to use PKCE
image
image
Should I still specify UsePkce default value?

martincostello: Yes, as the base class doesn't enable it by default.

hscokies: Added UsePkce default value, AuthenticationProperties check already at [L166](https://github.com/HScokies/AspNet.Security.OAuth.Providers.Vkid/blob/7c69c627aaec7d5ae251471861cbae4119883dfb/src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs#L166-L166)

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        if (!payload.RootElement.TryGetProperty("error", out var errorElement))
+        {
+            return OAuthTokenResponse.Success(payload);
+        }
+
+        var errorCode = errorElement.GetString()!;
+        var errorDescription = errorElement.GetProperty("error_description").GetString()!;
+        return OAuthTokenResponse.Failed(new Exception($"{errorCode}: {errorDescription}"));
```
martincostello: Can you factor this to match the pattern used in the other providers similar to this please? For example, the error and success cases are other way around and the HTTP status code isn't being explicitly checked, so a JSON response without an `error` property is always treated as success, even if it returns a non-success response.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/82d20e3a415be2203257ecc1222edde5766961bb/src/AspNet.Security.OAuth.Deezer/DeezerAuthenticationHandler.cs#L47-L59

Similarly for other methods in this handler class in terms of layout and style - that makes the providers easier to maintain if we have to make a similar change in a future version to all 90+ providers.

Not being similar is partly why I didn't realise this already handled PKCE at a glance, because it was structured differently and the accompanying comment like you find in the other providers.

hscokies: Got it! gonna refactor it later this week.
Also, should I add logger to this handler? If yes, are there any docs about event IDs?

martincostello: Just whatever logging makes sense in the same style as the other providers. The event IDs are just arbitrary numbers starting from 1 as the provider is implemented.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,335 @@
...
+        var properties = Options.StateDataFormat.Unprotect(query["state"]);
+        if (properties is null)
+        {
+            return HandleRequestResult.Fail("The oauth state was missing or invalid.");
+        }
+
+        if (ValidateCorrelationId(properties) is false)
+        {
+            return HandleRequestResult.Fail("Correlation failed.");
+        }
```
martincostello:
```suggestion
        // OAuth2 10.12 CSRF
        if (!ValidateCorrelationId(properties))
        {
            return HandleRequestResult.Fail("Correlation failed.", properties);
        }
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        // According to docs query cannot contain errors but VK documentation tends to lie so debug log here
+        Log.CodeResponse(Logger, query);
+
+        var code = Request.Query["code"];
+        if (StringValues.IsNullOrEmpty(code))
+        {
+            return HandleRequestResult.Fail("Code was not found.");
```
martincostello:
```suggestion
            return HandleRequestResult.Fail("Code was not found.", properties);
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        var deviceId = Request.Query["device_id"];
+        if (StringValues.IsNullOrEmpty(deviceId))
+        {
+            return HandleRequestResult.Fail("Device ID was not found.");
```
martincostello:
```suggestion
            return HandleRequestResult.Fail("Device ID was not found.", properties);
```

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        if (!payload.RootElement.TryGetProperty("error", out var errorElement))
+        {
+            return OAuthTokenResponse.Success(payload);
+        }
+
+        var errorCode = errorElement.GetString()!;
+        var errorDescription = errorElement.GetProperty("error_description").GetString()!;
+        return OAuthTokenResponse.Failed(new Exception($"{errorCode}: {errorDescription}"));
```
martincostello: Could you invert this so the failure case is in the `if` and the success case is the end of the method please?

test/AspNet.Security.OAuth.Providers.Tests/VkId/VkIdTests.cs
```diff
@@ -19,7 +19,14 @@ public VkIdTests(ITestOutputHelper outputHelper)
...
     {
-        builder.AddVkId(options => ConfigureDefaults(builder, options));
+        builder.AddVkId(
+            options =>
+            {
+                options.ClientId = "my-client-id";
+                options.ClientSecret = "my-client-secret";
+                options.Backchannel = CreateBackchannel(builder);
+                options.UseExchangeCodeAsyncStateValidation = false;
+            });
```
martincostello: Do something like this, then you only need to set one extra property:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/34f683a19d855fdc133dbd176cc0ca9766f422b0/test/AspNet.Security.OAuth.Providers.Tests/Okta/OktaTests.cs#L15-L19

Why is it that you need to turn this off?

hscokies: I'm out of ideas how to mock IDataProtector to pass tests failing on this [line](https://github.com/HScokies/AspNet.Security.OAuth.Providers.Vkid/blob/366b9092112d7a08a4ca28f4cf39921973d99d63/src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs#L199) :(

martincostello: You'll need to configure the HTTP request interceptor to echo back the state in the response body in some way. For all the other providers this is handled [here](https://github.com/HScokies/AspNet.Security.OAuth.Providers.Vkid/blob/366b9092112d7a08a4ca28f4cf39921973d99d63/test/AspNet.Security.OAuth.Providers.Tests/Infrastructure/LoopbackRedirectHandler.cs#L83-L121) by adjusting the callback URL.

We shouldn't really be building in an option there's no real reason to ever turn off from a consumer's point of view _just_ to get the tests to pass.

hscokies: phew, looks like I've finally got it :D

test/AspNet.Security.OAuth.Providers.Tests/VkId/VkIdTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        var fakeDataProtector = Substitute.For<IDataProtector>();
+
+        fakeDataProtector.Protect(Arg.Any<byte[]>())
+            .Returns(x => dataProtector.Protect(x.Arg<byte[]>()));
+        fakeDataProtector.Unprotect(Arg.Is<byte[]>(x => !x.SequenceEqual(Array.Empty<byte>())))
+            .Returns(x => dataProtector.Unprotect(x.Arg<byte[]>()));
+
+        // Use fake DP with empty AuthenticationProperties for ExchangeCodeAsync state validation
+        fakeDataProtector.Unprotect(Arg.Is<byte[]>(x => x.SequenceEqual(Array.Empty<byte>())))
+            .Returns([1, 0, 0, 0, 0, 0, 0, 0]);
```
martincostello: Could you make this use a slightly more realistic value that matches a non-empty `scope` value in the payload?

martincostello (approved):
Just one minor comment regarding the tests.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,337 @@
...
+    {
+        var query = Request.Query;
+        var properties = Options.StateDataFormat.Unprotect(query["state"]);
+        if (properties is null)
+        {
+            return HandleRequestResult.Fail("The oauth state was missing or invalid.");
+        }
+
+        // OAuth2 10.12 CSRF
+        if (ValidateCorrelationId(properties) is false)
```
kevinchalet: It's not technically wrong, but `!ValidateCorrelationId(properties)` would be easier to read (and consistent with the other providers).

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,337 @@
...
+        if (response.StatusCode is not HttpStatusCode.OK)
+        {
+            await Log.ExchangeCodeErrorAsync(Logger, response, Context.RequestAborted);
+            return OAuthTokenResponse.Failed(new Exception("Invalid remote server response during code exchange."));
+        }
+
+        var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+
+        // Code exchange response should always contain state parameter.
+        if (!payload.RootElement.TryGetProperty("state", out var state) ||
```
kevinchalet: `state` in token responses is not really a standard thing and it's not clear what security benefits validating the `state` here provides. I'd just remove that part to simplify things.

src/AspNet.Security.OAuth.VkId/VkIdAuthenticationHandler.cs
```diff
@@ -0,0 +1,337 @@
...
+        [NotNull] UrlEncoder encoder)
+        : base(options, logger, encoder)
+    {
+    }
+
+    protected override string BuildChallengeUrl(
+        [NotNull] AuthenticationProperties properties,
+        [NotNull] string redirectUri)
+    {
+        var parameter = Options.Scope;
```
kevinchalet: Looks like these temp variables aren't very useful: feel free to remove them and do the formatting directly in the dictionary initializer to save a few lines.

kevinchalet (commented):
Thanks for your PR!

kevinchalet:
(I find it funny that VK claims they are implementing "OAuth 2.1" - that hasn't been officially adopted yet - when all they're doing is implementing yet another non-standard version of OAuth with non-standard concepts like HTTP 200 errors and `device_id` 😄)

hscokies:
yea, VK is a funny company...
image

kevinchalet (approved):
Looks good, thanks a lot for your contribution! 👍🏻 

I'll let @martincostello merge it in case he'd have additional feedback 😃

martincostello:
Before merging this, can you confirm you've tested the provider against the production VK ID service and that you were able to successfully authenticate?

martincostello:
Oh, and could you update this to `3` and `0` please?

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/8bff342db12c404ae6604158c7b5be63cec23471/eng/Versions.props#L5-L6

hscokies:
> Before merging this, can you confirm you've tested the provider against the production VK ID service and that you were able to successfully authenticate?

Everything seems to work fine, yes
image
image

martincostello:
Thanks again for your contribution - the new provider 📦 is now available from NuGet.org: https://www.nuget.org/packages/AspNet.Security.OAuth.VkId/8.3.0

PR #1063, opened 2025-05-01
Adds Contentful provider

jerriep:


docs/contentful.md
martincostello, line 1: Doesn't look like there's any Contentful-specific settings here, so we could just not bother adding the docs for this provider.

martincostello (approved):
Could you bump these for 9.3.0, otherwise LGTM and I can push this out after merging:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/97600060bf30456537d54275b1ae53862f783291/eng/Versions.props#L5-L6

jerriep:
@martincostello Made updates as requested

martincostello:
@jerriep Package is now in NuGet.org.

PR #1071, opened 2025-05-21
Perf optimization for Bilibili provider

LeaFrock:
As stated in the title.

### DIFF
--- a/src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
+++ b/src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
@@ -90,17 +90,16 @@ public partial class BilibiliAuthenticationHandler : OAuthHandler<BilibiliAuthen
         return Convert.ToHexStringLower(hash);
     }
 
-    private static string BuildSignatureString(HttpRequestMessage request, string appSecret)
+    private static string BuildSignatureString(SortedList<string, string> xBiliHeaders, string appSecret)
     {
-        var headers = request.Headers
-            .Where(h => h.Key.StartsWith("x-bili-", StringComparison.OrdinalIgnoreCase))
-            .OrderBy(h => h.Key)
-            .Select(h => $"{h.Key}:{string.Join(",", h.Value)}")
-            .ToList();
+        StringBuilder sb = new(256); // 256 is an estimated size for the plain text
+        foreach (var xbh in xBiliHeaders)
+        {
+            sb.Append(xbh.Key).Append(':').Append(xbh.Value).Append('\n');
+        }
 
-        var signature = string.Join('\n', headers);
-
-        return ComputeHmacSHA256(appSecret, signature);
+        var signSrcText = sb.ToString(0, sb.Length - 1); // Ignore the last '\n'
+        return ComputeHmacSHA256(appSecret, signSrcText);
     }
 
     protected override async Task<AuthenticationTicket> CreateTicketAsync(
@@ -110,14 +109,14 @@ public partial class BilibiliAuthenticationHandler : OAuthHandler<BilibiliAuthen
     {
         using var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
         request.Headers.Add("access-token", tokens.AccessToken);
-        request.Headers.Add("x-bili-accesskeyid", Options.ClientId);
-        request.Headers.Add("x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e"); // It's a GET request so there's no content, so we send the MD5 hash of an empty string
-        request.Headers.Add("x-bili-signature-method", "HMAC-SHA256");
-        request.Headers.Add("x-bili-signature-nonce", Base64Url.EncodeToString(RandomNumberGenerator.GetBytes(256 / 8)));
-        request.Headers.Add("x-bili-signature-version", "2.0");
-        request.Headers.Add("x-bili-timestamp", TimeProvider.GetUtcNow().ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture));
 
-        var signature = BuildSignatureString(request, Options.ClientSecret);
+        var xBiliHeaders = BuildXBiliHeaders();
+        foreach (var xbh in xBiliHeaders)
+        {
+            request.Headers.Add(xbh.Key, xbh.Value);
+        }
+
+        var signature = BuildSignatureString(xBiliHeaders, Options.ClientSecret);
         request.Headers.Add("Authorization", signature);
 
         request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
@@ -145,6 +144,27 @@ public partial class BilibiliAuthenticationHandler : OAuthHandler<BilibiliAuthen
         return new AuthenticationTicket(context.Principal!, context.Properties, Scheme.Name);
     }
 
+    private SortedList<string, string> BuildXBiliHeaders()
+    {
+        var list = new SortedList<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            { "x-bili-accesskeyid", Options.ClientId },
+            { "x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e" }, // It's a GET request so there's no content, so we send the MD5 hash of an empty string
+            { "x-bili-signature-method", "HMAC-SHA256" },
+            { "x-bili-signature-nonce", GenerateNonce() },
+            { "x-bili-signature-version", "2.0" },
+            { "x-bili-timestamp", TimeProvider.GetUtcNow().ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture) }
+        };
+        return list;
+    }
+
+    private static string GenerateNonce()
+    {
+        Span<byte> bytes = stackalloc byte[256 / 8];
+        RandomNumberGenerator.Fill(bytes);
+        return Base64Url.EncodeToString(bytes);
+    }
+
     /// <summary>
     /// Check the code sent back by server for potential server errors.
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -145,6 +144,27 @@ protected override async Task<AuthenticationTicket> CreateTicketAsync(
...
+        var list = new SortedList<string, string>(StringComparer.OrdinalIgnoreCase)
+        {
+            { "x-bili-accesskeyid", Options.ClientId },
+            { "x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e" }, // It's a GET request so there's no content, so we send the MD5 hash of an empty string
+            { "x-bili-signature-method", "HMAC-SHA256" },
+            { "x-bili-signature-nonce", GenerateNonce() },
+            { "x-bili-signature-version", "2.0" },
+            { "x-bili-timestamp", TimeProvider.GetUtcNow().ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture) }
+        };
+        return list;
```
martincostello: Can just return it and remove the local, and if we're optimising then should pass the capacity.

```suggestion
        return new SortedList<string, string>(6, StringComparer.OrdinalIgnoreCase)
        {
            { "x-bili-accesskeyid", Options.ClientId },
            { "x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e" }, // It's a GET request so there's no content, so we send the MD5 hash of an empty string
            { "x-bili-signature-method", "HMAC-SHA256" },
            { "x-bili-signature-nonce", GenerateNonce() },
            { "x-bili-signature-version", "2.0" },
            { "x-bili-timestamp", TimeProvider.GetUtcNow().ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture) }
        };
```

src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -110,14 +109,14 @@ protected override async Task<AuthenticationTicket> CreateTicketAsync(
...
-        request.Headers.Add("x-bili-accesskeyid", Options.ClientId);
-        request.Headers.Add("x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e"); // It's a GET request so there's no content, so we send the MD5 hash of an empty string
-        request.Headers.Add("x-bili-signature-method", "HMAC-SHA256");
-        request.Headers.Add("x-bili-signature-nonce", Base64Url.EncodeToString(RandomNumberGenerator.GetBytes(256 / 8)));
-        request.Headers.Add("x-bili-signature-version", "2.0");
-        request.Headers.Add("x-bili-timestamp", TimeProvider.GetUtcNow().ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture));
-
-        var signature = BuildSignatureString(request, Options.ClientSecret);
+
+        var xBiliHeaders = BuildXBiliHeaders();
```
martincostello: Not a fan of the `x` prefix, so let's call it `xbiliHeaders`.

src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -110,14 +109,14 @@ protected override async Task<AuthenticationTicket> CreateTicketAsync(
...
+        foreach (var xbh in xBiliHeaders)
+        {
+            request.Headers.Add(xbh.Key, xbh.Value);
```
martincostello: Can this be deconstructed?
```suggestion
        foreach ((var name, var value) in xBiliHeaders)
        {
            request.Headers.Add(name, value);
```

src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -90,17 +90,16 @@ private static string ComputeHmacSHA256(string key, string data)
...
+    private static string BuildSignatureString(SortedList<string, string> xBiliHeaders, string appSecret)
     {
-        var headers = request.Headers
-            .Where(h => h.Key.StartsWith("x-bili-", StringComparison.OrdinalIgnoreCase))
-            .OrderBy(h => h.Key)
-            .Select(h => $"{h.Key}:{string.Join(",", h.Value)}")
-            .ToList();
-
-        var signature = string.Join('\n', headers);
+        StringBuilder sb = new(256); // 256 is an estimated size for the plain text
```
martincostello: Style preference:

```suggestion
        var builder = new StringBuilder(256); // 256 is an estimated size for the plain text
```

src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -90,17 +90,16 @@ private static string ComputeHmacSHA256(string key, string data)
...
+        foreach (var xbh in xBiliHeaders)
+        {
+            sb.Append(xbh.Key).Append(':').Append(xbh.Value).Append('\n');
```
martincostello: Style preference for readability:

```suggestion
            builder.Append(xbh.Key)
                   .Append(':')
                   .Append(xbh.Value)
                   .Append('\n');
```

src/AspNet.Security.OAuth.Bilibili/BilibiliAuthenticationHandler.cs
```diff
@@ -90,17 +90,16 @@ private static string ComputeHmacSHA256(string key, string data)
...
     {
-        var headers = request.Headers
-            .Where(h => h.Key.StartsWith("x-bili-", StringComparison.OrdinalIgnoreCase))
-            .OrderBy(h => h.Key)
-            .Select(h => $"{h.Key}:{string.Join(",", h.Value)}")
-            .ToList();
-
-        var signature = string.Join('\n', headers);
+        StringBuilder sb = new(256); // 256 is an estimated size for the plain text
+        foreach (var xbh in xBiliHeaders)
```
martincostello: Deconstruct?

martincostello (commented):
Just curious, do you have any numbers showing the actual impact of the changes?

martincostello:
> Just curious, do you have any numbers showing the actual impact of the changes?

LeaFrock:
> Just curious, do you have any numbers showing the actual impact of the changes?

It is not easy to write a precise microbenchmark. I tested a simplified and somewhat crude version.

The result is the following:

```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.4061)
AMD Ryzen 7 3700X, 1 CPU, 16 logical and 8 physical cores
.NET SDK 9.0.300
  [Host]     : .NET 9.0.5 (9.0.525.21509), X64 RyuJIT AVX2
  DefaultJob : .NET 9.0.5 (9.0.525.21509), X64 RyuJIT AVX2
```
| Method   | Mean       | Error    | StdDev   | Median     | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|--------- |-----------:|---------:|---------:|-----------:|------:|--------:|-------:|----------:|------------:|
| DictLinq | 1,209.9 ns | 24.13 ns | 48.19 ns | 1,188.7 ns |  1.00 |    0.06 | 0.2823 |   2.31 KB |        1.00 |
| SortList |   570.7 ns | 11.38 ns | 21.64 ns |   565.2 ns |  0.47 |    0.03 | 0.1764 |   1.45 KB |        0.62 |


The src codes:

```csharp
[MemoryDiagnoser]
public class DictLinqVsSortList
{
    private const string ClientId = "Bilibili_Cheers";
    private readonly string Timestamp = DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString();

    [Benchmark(Baseline = true)]
    public string DictLinq()
    {
        Dictionary<string, string> d = new(7)
        {
            { "access-token", "1234567890" },
            { "x-bili-accesskeyid", ClientId },
            { "x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e" }, // It's a GET request so there's no content, so we send the MD5 hash of an empty string
            { "x-bili-signature-method", "HMAC-SHA256" },
            { "x-bili-signature-nonce", GenerateNonce() },
            { "x-bili-signature-version", "2.0" },
            { "x-bili-timestamp", Timestamp }
        };
        var headers = d
           .Where(h => h.Key.StartsWith("x-bili-", StringComparison.OrdinalIgnoreCase))
           .OrderBy(h => h.Key)
           .Select(h => $"{h.Key}:{string.Join(",", h.Value)}")
           .ToList();
        return string.Join('\n', headers);
    }

    [Benchmark]
    public string SortList()
    {
        SortedList<string, string> s = new(6, StringComparer.OrdinalIgnoreCase)
        {
            { "x-bili-accesskeyid", ClientId },
            { "x-bili-content-md5", "d41d8cd98f00b204e9800998ecf8427e" }, // It's a GET request so there's no content, so we send the MD5 hash of an empty string
            { "x-bili-signature-method", "HMAC-SHA256" },
            { "x-bili-signature-nonce", GenerateNonce() },
            { "x-bili-signature-version", "2.0" },
            { "x-bili-timestamp", Timestamp }
        };
        return BuildSignatureString(s);

        static string BuildSignatureString(SortedList<string, string> xbiliHeaders)
        {
            var sb = new StringBuilder(256); // 256 is an estimated size for the plain text
            foreach ((var name, var value) in xbiliHeaders)
            {
                sb.Append(name)
                  .Append(':')
                  .Append(value)
                  .Append('\n');
            }

            return sb.ToString(0, sb.Length - 1); // Ignore the last '\n'
        }
    }

    private static string GenerateNonce()
    {
        Span<byte> bytes = stackalloc byte[256 / 8];
        RandomNumberGenerator.Fill(bytes);
        return Base64Url.EncodeToString(bytes);
    }
}
```

I do allocate an extra `SortedList<string, string>` but it's worthy to replace the LINQ and temp strings' selecting.

The fastest way should be generating the sorted `x-bili-` headers codes during compiling time with SG, which avoids the runtime sorting anymore,  but that would be a little bit complex to maintain and not quite meaningful.

Also, the `GenerateNonce` is a better version as it avoids a `byte[32]` heap allocation each time. Perhaps the JIT of .NET 10 can eliminate such differences since the byte array is small(32 bytes), but it is better to optimize at the code level.

martincostello:
Thanks for putting together the benchmark - ~2x throughput and memory improvement is a nice win.

I'll merge this now, but I won't release it today so there's not two releases in one day 😄

martincostello (approved):

LeaFrock:
Thanks. There's no rush for the next release.

I just think that library-level code should be as performance-friendly as possible, and I'm willing to make a contribution 😄  .

PR #1110, opened 2025-09-23
Use NuGet Trusted Publishing

martincostello:
Switch to using GitHub OIDC for pushing packages to NuGet.org with [Trusted Publishing](https://devblogs.microsoft.com/dotnet/enhanced-security-is-here-with-the-new-trust-publishing-on-nuget-org/).

@kevinchalet Could you update the repository configuration so this will work please?

Steps:

1. Create an environment called `NuGet.org`
2. Add a secret to the `NuGet.org` environment named `NUGET_USER` whose value is my NuGet.org username (it's not a secret, but I think it's better not to be hard-coded in the workflow 😃)
3. (Optionally) delete the `NUGET_API_KEY` secret

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,6 +172,11 @@ jobs:
 
   publish-nuget:
     needs: [ build, validate-packages ]
+    environment:
+      name: NuGet.org
+      url: https://www.nuget.org/profiles/aspnet-contrib
+    permissions:
+      id-token: write
     runs-on: ubuntu-latest
     if: |
       github.event.repository.fork == false &&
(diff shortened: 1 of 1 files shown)


kevinchalet:
> @kevinchalet Could you update the repository configuration so this will work please?

Done on both repos. Let me know if you need anything else 😃 

> (Optionally) delete the `NUGET_API_KEY` secret

Once we're 100% sure NTP works as intended, I'll take care of that 👍🏻

martincostello:
> Once we're 100% sure NTP works as intended

I hope it does, as I used it to publish [this](https://github.com/martincostello/openapi-extensions/releases/tag/v2.0.0-preview.2) 😅

martincostello:
> Once we're 100% sure NTP works as intended, I'll take care of that 👍🏻

Looks like [it's working](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/actions/runs/18123796968/job/51574577431).

kevinchalet:
> Looks like [it's working](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/actions/runs/18123796968/job/51574577431).

Great! I deleted the API key on NuGet.org and removed the corresponding secret here 😃

PR #1112, opened 2025-09-27
feat(Keycloak): add "family_name" mapping to ClaimTypes.Surname

XB-RS:
This commit will add support for the default ASP NET Core ClaimTypes.Surname. Default key for the surname is "family_name" in Keycloak.


martincostello (approved):

martincostello:
The [bundle file](https://github.com/XB-RS/AspNet.Security.OAuth.Providers/blob/dev/test/AspNet.Security.OAuth.Providers.Tests/Keycloak/bundle.json) needs updating so the tests will pass.

PR #1126, opened 2025-11-07
Add Etsy OAuth provider

### LINKED ISSUE #1082: Add Etsy oauth2
### Which provider would you like support added for?

Etsy oauth2 for api v3 (latest)

### Provider Documentation

https://developers.etsy.com/documentation/

https://developers.etsy.com/documentation/reference/

### Are you willing to implement this provider yourself?

I would like to participate but I am missing the appropriate knowledge to do it my own so if anyone would be willing to help...?

### Additional context

If you might read their GitHub repo Which is only used for issues or more using just discussion, you will see there are also others that are struggling with the docs and guidance they are providing, which made them already create python scripts just for doing that auth flow. I did not see any dotnet or asp net core code for this so far. That's why I would think this could be valuable for not just me

DevTKSS:
This pull request adds initial support for Etsy OAuth authentication to the codebase. It introduces the core implementation for an ASP.NET Core authentication provider, including configuration, constants, and documentation. The changes provide all the necessary building blocks and guidance for integrating Etsy login into an ASP.NET Core application.

### Etsy OAuth Provider Implementation

* Added the main project file `AspNet.Security.OAuth.Etsy.csproj` to define the package and its dependencies for the Etsy authentication provider.
* Introduced `EtsyAuthenticationAccessType` enum to distinguish between personal and commercial access types for Etsy apps.
* Added `EtsyAuthenticationConstants` class with claim and scope constants for mapping Etsy user data and permissions.
* Defined `EtsyAuthenticationDefaults` with all the required default values for endpoints, scheme names, and callback paths used by the Etsy authentication middleware.
* Implemented extension methods in `EtsyAuthenticationExtensions` to make it easy to add Etsy authentication to an ASP.NET Core application via the authentication builder.

### Documentation

* Added a comprehensive setup and usage guide in `docs/etsy.md`, detailing configuration, required/optional settings, claims mapping, and sample code for integrating Etsy OAuth authentication.

> [!NOTE]
> The AccessType enum does so far only have one member `Personal` because I only have this level of App registration and can not see or tell if the commercial Access Apps do somehow support confidential clients against the guidelines of the Etsy Authentication API. I asked their support some while ago which did tell that the `shared secret` would be used at refresh, but there is no note at all for this in their docs, so this provider is adhering to the existing docs and oAuth standards [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) and does treat it as [Public Client](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1).
> Recognizing the #610 Issue in the past, I directly implemented the same fix up from the start, because even if the user sets the value, we don't have to require it, so the provider should also not fail because of not setting this.

Closes: #1082


AspNet.Security.OAuth.Providers.sln
```diff
@@ -328,450 +329,1341 @@ EndProject
...
+EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AspNet.Security.OAuth.Contentful", "src\AspNet.Security.OAuth.Contentful\AspNet.Security.OAuth.Contentful.csproj", "{B1F6EA42-7B1B-469E-B304-6B2E6FE39852}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AspNet.Security.OAuth.Etsy", "src\AspNet.Security.OAuth.Etsy\AspNet.Security.OAuth.Etsy.csproj", "{53B5B8F0-023E-4D2D-84F0-5B68610682A4}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
```
martincostello: We don't build CPU architecture-specific assemblies so all these x64 and x86 configurations should be removed.

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```
martincostello, line 5: The version here is just a placeholder - given we'll be shipping 10.0.0 next week, it's likely going to be 10.1.0.

```suggestion
    <PackageValidationBaselineVersion>9.5.0</PackageValidationBaselineVersion>
    <TargetFrameworks>$(DefaultNetCoreTargetFramework)</TargetFrameworks>
    <!-- TODO Remove once published to NuGet.org -->
    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
```

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
```diff
@@ -0,0 +1,18 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>ASP.NET Core security middleware enabling Etsy authentication.</Description>
+    <Authors>Sonja</Authors>
```
martincostello: Your full name or NuGet user name would be preferred. I can see from your LinkedIn URL in your GitHub profile that it isn't a secret.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationAccessType.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Public client access type aka 'private usage access' in Etsy App Registration.
+    /// </summary>
+    Personal,
+
+    //// <summary>
+    //// Confidential client access type aka 'commercial usage access' in Etsy App Registration. // TODO: Uncomment if someone can verify that commercial usage access supports confidential clients.
+    //// </summary>
+    // Commercial
```
martincostello: I'd be reticent to actually ship this without this being validated as working. It would be better to just remove it entirely and add it later when needed and can be tested.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationConstants.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace AspNet.Security.OAuth.Etsy;
+
+/// <summary>
+/// Contains constants specific to the <see cref="EtsyAuthenticationHandler"/>.
+/// </summary>
+public static class EtsyAuthenticationConstants
+{
+    public static class Claims
+    {
+        public const string UserId = "urn:etsy:user_id";
```
martincostello: Make all the `const string`s `static readonly string` please.

We've been affected in the past by providers making breaking changes to providers with changes to values, and `const` values get compiled into assemblies consuming them so can cause issues, so as this is a new provider I think it's better we avoid them being `const`.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Default value for <see cref="AuthenticationSchemeOptions.ClaimsIssuer"/>.
+    /// </summary>
+    public static readonly string Issuer = "Etsy";
+
+    /// <summary>
+    /// Default value for <see cref="RemoteAuthenticationOptions.CallbackPath"/>.
+    /// </summary>
+    /// <value>
+    /// "/signin-etsy"
+    /// </value>
```
martincostello:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/> <see href="https://developers.etsy.com/documentation/reference/#operation/getMe">Etsy getMe Endpoint</see>.
+    /// </summary>
+    public static readonly string UserInformationEndpoint = "https://openapi.etsy.com/v3/application/users/me";
+
+    /// <summary>
+    /// Default value for the Etsy v3 API Base URI
+    /// </summary>
+    /// <value>
+    /// https://openapi.etsy.com/v3/application/
+    /// </value>
```
martincostello:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
```diff
@@ -0,0 +1,67 @@
...
+    public static readonly string EtsyBaseUri = "https://openapi.etsy.com/v3/application/";
+
+    /// <summary>
+    /// Default value for Etsy user details endpoint path <see href="https://developers.etsy.com/documentation/reference/#operation/getUser">getUser</see>.
+    /// </summary>
+    /// <value>
+    /// "/users/{user_id}"
+    /// </value>
```
martincostello:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+    }
+
+    protected override async Task<AuthenticationTicket> CreateTicketAsync(
+        [NotNull] ClaimsIdentity identity,
+        [NotNull] AuthenticationProperties properties,
+        [NotNull] OAuthTokenResponse tokens)
+    {
+        // Get the basic user info (user_id and shop_id)
+        using var meRequest = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
```
martincostello:
```suggestion
        using var request = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        meRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Json));
+        meRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+        meRequest.Headers.Add("x-api-key", Options.ClientId);
```
martincostello: Just checking - is it correct that the request is authorized with two different headers, or is their header just poorly named and its an ID not a key?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        using var meResponse = await Backchannel.SendAsync(meRequest, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
```
martincostello:
```suggestion
        using var response = await Backchannel.SendAsync(meRequest, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        if (!meResponse.IsSuccessStatusCode)
+        {
+            await Log.UserProfileErrorAsync(Logger, meResponse, Context.RequestAborted);
+            throw new HttpRequestException("An error occurred while retrieving basic user info from Etsy.");
+        }
+
+        using var mePayload = JsonDocument.Parse(await meResponse.Content.ReadAsStringAsync(Context.RequestAborted));
+        var meRoot = mePayload.RootElement;
```
martincostello: Same here - remove the `me` prefixes - there's no other requests/responses/payloads etc, so we don't need to worry about making it more specific.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        using var meRequest = new HttpRequestMessage(HttpMethod.Get, Options.UserInformationEndpoint);
+        meRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Json));
+        meRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+        meRequest.Headers.Add("x-api-key", Options.ClientId);
+
+        using var meResponse = await Backchannel.SendAsync(meRequest, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+        if (!meResponse.IsSuccessStatusCode)
+        {
+            await Log.UserProfileErrorAsync(Logger, meResponse, Context.RequestAborted);
+            throw new HttpRequestException("An error occurred while retrieving basic user info from Etsy.");
```
martincostello:
```suggestion
            throw new HttpRequestException("An error occurred while retrieving basic user information from Etsy.");
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /// <summary>
+    /// Retrieves detailed user information from Etsy.
+    /// </summary>
+    /// <param name="tokens">The OAuth token response.</param>
+    /// <returns>A JSON document containing the detailed user information.</returns>
+    protected virtual async Task<JsonDocument> GetDetailedUserInfoAsync([NotNull] OAuthTokenResponse tokens)
+    {
+        using var userRequest = new HttpRequestMessage(HttpMethod.Get, EtsyAuthenticationDefaults.EtsyBaseUri + EtsyAuthenticationDefaults.UserDetailsPath);
```
martincostello: `request` and `response`.

martincostello: These URLs should be configurable in the options, not hard-coded. See other providers for examples. This also isn't a safe way to concatenate URLs, e.g. in the case where the base URI ends in a slash as you'll get `//` in the path.

martincostello: Is the URL genuinely `https://openapi.etsy.com/v3/application//users/{user_id}`? It looks like it's a placeholder, but this code isn't replacing the placeholder. Notice also the double `//` from the concatenation as previously noted.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        userRequest.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Json));
+        userRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
+        userRequest.Headers.Add("x-api-key", Options.ClientId);
+
+        using var userResponse = await Backchannel.SendAsync(userRequest, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
+        if (!userResponse.IsSuccessStatusCode)
+        {
+            await Log.UserProfileErrorAsync(Logger, userResponse, Context.RequestAborted);
```
martincostello: Re-using the same log event for two separate API requests might create user confusion as to which one actually failed without them inspecting the stack trace.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+using static AspNet.Security.OAuth.Etsy.EtsyAuthenticationConstants;
+
+namespace AspNet.Security.OAuth.Etsy;
+
+/// <summary>
+/// Defines a set of options used by <see cref="EtsyAuthenticationHandler"/>.
+/// </summary>
+public class EtsyAuthenticationOptions : OAuthOptions
+{
+    public bool IncludeDetailedUserInfo { get; set; }
```
martincostello: Place after the constructor and add `///` documentation.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        AuthorizationEndpoint = EtsyAuthenticationDefaults.AuthorizationEndpoint;
+        TokenEndpoint = EtsyAuthenticationDefaults.TokenEndpoint;
+        UserInformationEndpoint = EtsyAuthenticationDefaults.UserInformationEndpoint;
+
+        // Enable PKCE by default (required by Etsy)
+        UsePkce = true;
+
+        // Enable refresh token support
+        SaveTokens = true;
```
martincostello: These comments don't really add any value here.

```suggestion
        UsePkce = true;
        SaveTokens = true;
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        // Ensure PKCE is enabled (required by Etsy)
+        if (!UsePkce)
+        {
+            throw new ArgumentException("PKCE is required by Etsy Authentication and must be enabled.", nameof(UsePkce));
+        }
+
+        if (!SaveTokens)
+        {
+            throw new ArgumentException("Saving tokens is required by Etsy Authentication and must be enabled.", nameof(SaveTokens));
+        }
```
martincostello: I'm not sure we need all this double-checking for Etsy-specific options here. We don't do this for most of the other providers.

test/AspNet.Security.OAuth.Providers.Tests/Etsy/EtsyAuthenticationOptionsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+namespace AspNet.Security.OAuth.Etsy;
+
+public static class EtsyAuthenticationOptionsTests
+{
+    public static TheoryData<EtsyAuthenticationAccessType> AccessTypes => new()
+    {
+        { EtsyAuthenticationAccessType.Personal }, // Private Etsy API access does not use client secret (aka 'shared secret' in Etsy App Registration) https://developers.etsy.com/documentation/essentials/authentication
+
+        // { EtsyAuthenticationAccessType.Commercial } // TODO: Verify commercial access app registration Authentication does support confidential clients. Etsy docs do not indicate this to be used at all but support stated this would be required for token refresh
```
martincostello: As noted previously, I think we should deleted all these untested and skipped combinations.

test/AspNet.Security.OAuth.Providers.Tests/Etsy/bundle.json
```diff
@@ -0,0 +1,38 @@
...
+      "uri": "https://openapi.etsy.com/v3/application/users/me",
+      "contentFormat": "json",
+      "contentJson": {
+        "user_id": 123456,
+        "shop_id": 789012
+      }
+    },
+    {
+      "comment": "Etsy /v3/application/users/{user_id} endpoint - returns detailed user information",
+      "uri": "https://openapi.etsy.com/v3/application/users/123456",
```
martincostello, line 27: Based on a previous comment, I don't see how this URL would match and thus the test would fail.

AspNet.Security.OAuth.Providers.sln
DevTKSS: oh, not noticed it, had some AI help since I am kinda new with auth and asp net core 😅

test/AspNet.Security.OAuth.Providers.Tests/Etsy/bundle.json
DevTKSS: @martincostello reviewing it currently along the comment of copilot: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1126#discussion_r2502947548

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
DevTKSS: @martincostello okay, do I need to install 10.0.0 then too for validation locally? do you know of inteferencing breaking changes that would cause possibly issus if only build with 9?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        UsePkce = true;
+
+        // Enable refresh token support
+        SaveTokens = true;
+
+        // Default scopes - Etsy requires at least one scope and this is the one for basic user info
+        Scope.Add(Scopes.ShopsRead);
+
+        // Map basic user claims
+        ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "user_id");
```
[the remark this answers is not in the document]
DevTKSS: the identifyer should be the `user_id` in my opinion reading the usage in auth docs of them and their reference, so the solution should be opposite of what copilot assumed 👍 
But now I have the small problem, that it errors me with duplicate mappings, so I checked how its done in the existing providers, where there is no duplicate adding (removed them in my local code already) but what if the user not notices we did already add them by default? The Etsy "UserInfoEndpoint" is not like in other API's, where you would directly call the `getUser` Endpoint and not have the `getMe` inbetween🤔 

@martincostello can I have your opinion to this?
My solution idea would be, to provide a `DetailedUserInfoClaims` `IReadOnlyList<string>` in the Defaults class maybe, so the caller could use a quick & clean loop to get them all and I can also use it for testing or similar 🤔 
Similar means like, I could also assume that if he sets the `IncludeDetailedUserInfo` option to `true`, I would add those claims in my handler with loop, but this would have the downside (if this would be the default behaviour) that the caller might not want/need the first/last name or the image, then just the email 🤷 then the cookie would maybe get bigger than required.
How critical is this, so what should we do?

- [Linking assumed `UserInfoEndpoint` `getMe` at reference with only `shop_id` and `user_id`](https://developers.etsy.com/documentation/reference#operation/getMe)
- [Linking assumed `DetailedUserInfoEndpoint` `getUser` at reference with `first_name`, `last_name`, `primary_email`, `image_url_75x75` and the identifier `user_id`](https://developers.etsy.com/documentation/reference#operation/getUser)
  But I dont think this should be a **always done** request, because it's requiring the `email_r` Scope and the endpoint description lets me assume we are not guaranteed that the client aggrees with letting us get it 🤔 

against this, the call to [`getUserAdress`](https://developers.etsy.com/documentation/reference#operation/getUserAddress) is defintly caller task to do, not from the provider in my opinion.

test/AspNet.Security.OAuth.Providers.Tests/Etsy/EtsyAuthenticationOptionsTests.cs
DevTKSS: @martincostello as the docs do not explicitly tell about the shared secret is used at all, maybe just assume public client to be always the case? 🤔 
my only reason for considering the opposite is:
- the answer of the etsy support
  not public visible they using a ticket chat, but here is a screenshot of what I am reffering to:
  image
  image
  (Link url that was used from him: https://datatracker.ietf.org/doc/html/rfc6749)
  from their docs I would assume, the refresh token flow he meant to link is this [rcf 6749 Section-1.5](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5)
- https://github.com/etsy/open-api/discussions/1443 <- asked in their discussion-based "report issues here" repo and have been told, that this would be used. 🤔 

I think, you have way more expirience with this, I would be greatfull if you could check if we need to still require the `shared secret` to be provided, even if its never stated to be used in their docs or the oAuth standard for the `Authorization Code Grant` flow...

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
DevTKSS: 😄 indeed this wasnt even on purpose, but thanks for giving me a heads up, will update it 👍

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
DevTKSS: But it's required from them... or do you mean to just remove the validation, state in the docs of it to not disable it and then setting it in the ctor by default to true like it is already?

martincostello: The default configuration should work, and if they change it manually they'll find it won't work. I don't think we need to worry about the additional code to maintain to detect that and break it if users try to change it for some reason.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello I think it's correct that we need both, referring to the docs of it: https://developers.etsy.com/documentation/reference#operation/getMe
Except from maximum a hand full of endpoints (knowing of the ping endpoint for example) the Etsy API requires the `api_key` aka our `client_id` / `keystring` in the app registration and the `oauth2` token with correct scope for successfull request

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: leftovers from initial approach where I had both calls in one function. 👍

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello so I would aggree with the fact of the slashes 👍
But not sure about the concatenating 🤔 first idea was just string.Format, which I know from the non web sdk like on Desktop apps, but then I indeed got this interesting [CA1863](https://learn.microsoft.com/de-de/dotnet/fundamentals/code-analysis/quality-rules/ca1863) which holds also a Sample recommendation that looked like it could work. 👀 
It's Using static field of `CompositeFormat`...

But this would not be possible in Options then.
So either:
- we get the perf of CompositeFormat and could add a Option Property of `UserId` that could get evaluated instead of fetching it again before get the `shop_id`
- I can re-add the path that could be added with replace and so on, but from my POV this sounds unnessesary complicated if the first option would work 🤔 

I will add the first option first and you can review and tell if you still want the second one with string and extension methods like `.Replace(...,userId)`  and I could still implement it then. 👍

AspNet.Security.OAuth.Providers.sln
DevTKSS: 🤔 checking this right now and I think that this was caused by vs 2022/vs code. I started in vs code, which doest need the csproj to be in the solution, but adding the line manually in vs2022 then switching over to vscode and back...
I do remember it told me, that it would have a problem (was not saying which of course) and that I would have to reload this project.
never seen the red X before on a project and it wasnt the banner I know from other dotnet based sdk's...

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: removed the duplicated slash 👍

test/AspNet.Security.OAuth.Providers.Tests/Etsy/bundle.json
DevTKSS: @martincostello I never worked with such .json file using tests before, but seeing the test results, it seems that my code gets the `/123456` as it should, but setting the uri in this bundle.json to the formatting string does result in the error:

```
Microsoft.AspNetCore.Authentication.AuthenticationFailureException : An error was encountered while handling the remote login.
---- JustEat.HttpClientInterception.HttpRequestNotInterceptedException : No HTTP response is configured for GET https://openapi.etsy.com/v3/application/users/123456.
at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
at Microsoft.AspNetCore.TestHost.HttpContextBuilder.<>c__DisplayClass23_0.<<SendAsync>g__RunRequestAsync|0>d.MoveNext()
at Microsoft.AspNetCore.TestHost.ClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
at AspNet.Security.OAuth.Infrastructure.LoopbackRedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) (file:///c%3A/Users/Sonja/source/repos/AspNet.Security.OAuth.Providers/test/AspNet.Security.OAuth.Providers.Tests/Infrastructure/LoopbackRedirectHandler.cs#L77,0)
at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
at AspNet.Security.OAuth.OAuthTests`1.AuthenticateUserAsync(WebApplicationFactory`1 server) (file:///c%3A/Users/Sonja/source/repos/AspNet.Security.OAuth.Providers/test/AspNet.Security.OAuth.Providers.Tests/OAuthTests%601.cs#L256,0)
at AspNet.Security.OAuth.Providers.Tests.Etsy.EtsyTests.Includes_Detailed_Claims_When_IncludeDetailedUserInfo_Is_True() (file:///c%3A/Users/Sonja/source/repos/AspNet.Security.OAuth.Providers/test/AspNet.Security.OAuth.Providers.Tests/Etsy/EtsyTests.cs#L80,0)
at JustEat.HttpClientInterception.InterceptingHttpMessageHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
at Microsoft.Extensions.Http.Logging.LoggingScopeHttpMessageHandler.<SendCoreAsync>g__Core|4_0(HttpRequestMessage request, Boolean useAsync, CancellationToken cancellationToken)
at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
at AspNet.Security.OAuth.Etsy.EtsyAuthenticationHandler.GetDetailedUserInfoAsync(OAuthTokenResponse tokens, Int64 userId) (file:///c%3A/Users/Sonja/source/repos/AspNet.Security.OAuth.Providers/src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs#L117,0)
at AspNet.Security.OAuth.Etsy.EtsyAuthenticationHandler.CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens) (file:///c%3A/Users/Sonja/source/repos/AspNet.Security.OAuth.Providers/src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs#L70,0)
at Microsoft.AspNetCore.Authentication.OAuth.OAuthHandler`1.HandleRemoteAuthenticateAsync()
at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
```

as soon as I did just insert the number back in there it was fine 🤔 do you know more about this maybe?
I would assume this value in bundle.json needs to be this way, reading this report.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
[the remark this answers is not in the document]
DevTKSS: @martincostello this comment should not longer apply, what do you think?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello should now be solved, could you check?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        string userDetailsUrl;
+        if (!string.IsNullOrWhiteSpace(Options.DetailedUserInfoEndpoint))
+        {
+            userDetailsUrl = string.Format(CultureInfo.InvariantCulture, Options.DetailedUserInfoEndpoint, userId);
+        }
+        else
+        {
+            userDetailsUrl = string.Format(CultureInfo.InvariantCulture, EtsyAuthenticationDefaults.DetailedUserInfoEndpoint, userId);
+        }
```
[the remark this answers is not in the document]
DevTKSS: tried this before... At least when the parameter were `var` this caused IDE to tell me its not clear what is returned CompositeFormat or string. Which does not make sense because both returning string but🤷 its the compiler... Maybe it likes it better when its now not var then string declated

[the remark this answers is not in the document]
DevTKSS: tested, not possible to use `??` or `?` with CompositeFormat + string.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: for reference, added screenshots of the Etsy developer support conversation about client ID, x-api-key, keystring and so on here: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1126/files/2c84ed859a5d62015348ef22ea90f4ba40a1fca6#r2503440727
by this, I would understand this setup as correct like this.
Using the `Options.ClientId` is meeting their `client_id` json naming and I tried to tell this detail that its the `keystring` in the App Registration in the [`docs/etsy.md`](https://github.com/DevTKSS/AspNet.Security.OAuth.Providers/blob/add-etsy-oauth-provider/docs/etsy.md#required-additional-settings)

martincostello:
Thanks for responding to the various comments - I'll take a look at this again when I get the time. Shipping the release for .NET 10 was the main priority this week.

DevTKSS:
@martincostello thanks for your response and time!
Until you find time, do you think from what you seen and the tests implemented by now, I could use the produced .dll locally in a ASP.NET Core API project to do an actual test call? 🤔 

I am not sure in this, because I have had lots of trouble with implementing oAuth in there, because of **Antiforgery** and **Cors** 😓

the Official MS Docs for oAuth2 are horribly lacking on guidiance when there is no exiting provider like Google or something OIDC that supports discovery endpoint, which the Etsy API doesn't have 👀

We do have the Auth Code Exchange implemented here, which is against the Cross Origin Forgery Attacking 🤔
I would currently miss the information, if we then still need specific setup content for Antiforgery additionally to what's told in our:
- [Basic Provider Readme](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/docs/README.md)
- [Repo root Readme](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/README.md)

And in the Authentication/Authorization docs for minimal api here:
- https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?view=aspnetcore-9.0

Its not told in our Readme files, if we do need Cookie to be added, or if the oAuth Base Extension our Providers here are building on do this already for us by default.

In case we need the dev user (like me then) to add this for e.g. have something like the Refreshing of the Token implemented (which I seen in none of the existing Providers here (?), so did not add it in the Etsy Provider too) eventually it would be usefull to add a Link to the [Cookie Auth ms docs](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/cookie?view=aspnetcore-9.0) eventually to the [OIDC docs](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/configure-oidc-web-authentication?view=aspnetcore-9.0) which contains a auth flow diagram with `code` and `code_verifyer` or at least to [Social Providers Auth](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-9.0) ?

**BUT** while on the social login providers docs do link to the External Providers Page, which also links to this repo (thats how I found it 😉) this would be missing information about Refreshing Tokens too 🤷 

If you would like me to open another issue / PR for this to improve these spots, let me know 👍

martincostello:
You _should_ be able to use our [MVC sample App](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/tree/dev/samples/Mvc.Client) to test the changes locally.

DevTKSS:
@martincostello Hey, I am on trying the provider with the `samples/Mvc.Client` but a bit confused how and where to set the Callback Url, as the Etsy API requires me to pre-provide one or more Callback Url's, no dynamic Url, no dynamic Port etc.
The Options we configure in the IOptions loaded appsettings or User Secrets for the ClientId, we can provide the CallbackPath, but seems like the Mvc.Client is overwriting it? 👀 

```csharp
    [HttpPost("~/signin")]
    public async Task<IActionResult> SignIn([FromForm] string provider)
    {
        // Note: the "provider" parameter corresponds to the external
        // authentication provider chosen by the user agent.
        if (string.IsNullOrWhiteSpace(provider))
        {
            return BadRequest();
        }

        if (!await HttpContext.IsProviderSupportedAsync(provider))
        {
            return BadRequest();
        }

        // Instruct the middleware corresponding to the requested external identity
        // provider to redirect the user agent to its own authorization endpoint.
        // Note: the authenticationScheme parameter must match the value configured in Startup.cs
        return Challenge(new AuthenticationProperties { RedirectUri = "/" }, provider);
    }
```
Could you maybe take a quick look and let me know if I just misunderstood this and the CallbackPath set in Options can be used as expected?
I am not getting the exact difference between them I guess...
- CallbackPath -> EtsyAuthenticationOptions
- RedirectUri -> Mvc.Client AuthenticationProperties

the Value I would like to set is e.g. `https://localhost:8080/etsy/callback`

martincostello:
`CallbackPath` is where the OAuth provider will redirect the user after authenticating so that an authenticated session can be created for your application by the authentication middleware. For your example, you would set this to `"etsy/callback"`.

`RedirectUri` is the URL the user is directed to in your website once they have successfully authenticated and been signed in.

DevTKSS:
@martincostello 🎉 🥳 Worked!!! just that the `CallbackPath` needed to be `/callback/etsy` with the `/` before
image

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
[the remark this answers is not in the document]
DevTKSS: @martincostello While validating the claim output, I identified two potential enhancements that might improve alignment with Etsy’s API conventions.
Since I’m still new to the claims-mapping aspects of ASP.NET Core, I’d appreciate your assessment on whether either approach could introduce complications in claim processing.

The `getMe` endpoint returns:

```json
{
  "user_id": 1,
  "shop_id": 1
}
```

### **Option 1 — Expose `urn:etsy:user_id` as an additional optional claim**

Map it to the same value currently used for `nameIdentifier`. This would mirror Etsy’s terminology without altering existing logic.

### **Option 2 — Derive `nameIdentifier` from a combination of `user_id` and `shop_id`**

Since both values are always present, even with our minimal configuration setup using `IncludeDetailedUserInfo = false` a combined identifier might avoid ambiguity between the two identifiers.

The current implementation behaves correctly based on tests and `Mvc.Client`. These ideas are non-breaking suggestions, provided they don’t conflict with the claims pipeline.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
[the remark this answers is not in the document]
DevTKSS: I would assume, as tests are passing and the Auth flow did work with the Mvc.Client on the "living object" (real API) this suggestion is not needed like this and fine with the current setup.
@martincostello would you agree with this?

DevTKSS: @martincostello Since everything works correctly with the current setup, can we mark this as resolved?

*I didn’t override the `UserInfoEndpoint` in the `Mvc.Client` test configuration, because that setup wouldn’t be able to call the `DetailedUserInfoEndpoint` on the Etsy API when using a different endpoint.*

README.md
```diff
@@ -92,6 +92,7 @@ We would love it if you could help contributing to this repository.
 * [Sinan](https://github.com/SH2015)
 * [Stefan](https://github.com/Schlurcher)
 * [Steffen Wenz](https://github.com/swenz)
+* [Sonja Schweitzer](https://github.com/DevTKSS)
```
martincostello: Nit: for sorting, this should be two lines above after Sinan.

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageValidationBaselineVersion>9.5.0</PackageValidationBaselineVersion>
```
martincostello:
```suggestion
    <PackageValidationBaselineVersion>10.1.0</PackageValidationBaselineVersion>
```

src/AspNet.Security.OAuth.Etsy/ClaimActionCollectionExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+/// Provides extension methods for <see cref="ClaimActionCollection"/> to map Etsy API specific user claims.
+/// </summary>
+public static class ClaimActionCollectionExtensions
+{
+    /// <summary>
+    /// Maps the Etsy user's profile image URL (75x75) to the <see cref="EtsyAuthenticationConstants.Claims.ImageUrl"/> claim.
+    /// </summary>
+    public static void MapImageClaim(this ClaimActionCollection collection)
+    {
+        collection.MapJsonKey(EtsyAuthenticationConstants.Claims.ImageUrl, "image_url_75x75");
```
martincostello, line 22: For the sake of one line of code in a method used in one place, let's just put this in the place it's used to avoid additional public API surface area we'd need to maintain.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <summary>
+    /// Default value for <see cref="OAuthOptions.UserInformationEndpoint"/> <see href="https://developers.etsy.com/documentation/reference/#operation/getMe">Etsy getMe Endpoint</see>.
+    /// </summary>
+    public static readonly string UserInformationEndpoint = "https://openapi.etsy.com/v3/application/users/me";
+
+    /// <summary>
+    /// Default value for receiving the user profile based upon a unique user ID <see href="https://developers.etsy.com/documentation/reference/#operation/getUser">getUser</see>.
+    /// </summary>
+    public static readonly CompositeFormat DetailedUserInfoEndpoint = CompositeFormat.Parse("https://openapi.etsy.com/v3/application/users/{0}");
```
martincostello: Feels a bit weird to type this as a `CompositeFormat` in the public API surface. If there's an analyser suggesting this is used, I would make this a string and then have the `CompositeFormat` for re-use be a private implementation detail.

src/AspNet.Security.OAuth.Etsy/EtsyPostConfigureOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+    public void PostConfigure(string? name, EtsyAuthenticationOptions options)
+    {
+        // Auto-add the email_r scope if detailed user info was requested but the scope not explicitly supplied.
+        if (options.IncludeDetailedUserInfo && !options.Scope.Contains(EtsyAuthenticationConstants.Scopes.EmailRead))
+        {
+            options.Scope.Add(EtsyAuthenticationConstants.Scopes.EmailRead);
+
+            options.ClaimActions.MapJsonKey(ClaimTypes.Email, "primary_email");
+            options.ClaimActions.MapJsonKey(ClaimTypes.GivenName, "first_name");
+            options.ClaimActions.MapJsonKey(ClaimTypes.Surname, "last_name");
```
martincostello: I _think_ you can just map these actions anyway - if the values aren't present in the payload, then the claims should just be absent. If that's the case, then we can avoid the need for this class entirely.

test/AspNet.Security.OAuth.Providers.Tests/Etsy/bundle.json
martincostello: I think you've updated the PR since Copilot started it now that the expectations match.

martincostello (commented):
I haven't fully reviewed this PR yet, or thought about responding to the specific questions yet, but I've done a quick pass. I'll come back to this another time next week.

src/AspNet.Security.OAuth.Etsy/EtsyPostConfigureOptions.cs
DevTKSS: @martincostello and what about the Scope EmailRead? I thought about adding them into the `EtsyAuthenticationOptions.Validate` to avoid this class to, but this would be against the task of the Validation Method and would potentially not be executed at all if e.g. somone uses not Microsoft Extensions Options Validation way with DataAnnotations then FluentValdiation 🤔 or is this handled by the build package automatically? On Desktop targets it's not, but is potentially different with ASP NET Core Authentication 👍

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
DevTKSS: yes, indeed this is caused by [CA1863](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1126#discussion_r2505651204) if I attempt to use string.Format 🫤 should I instead add a NoWarn in the csproj for this?

src/AspNet.Security.OAuth.Etsy/ClaimActionCollectionExtensions.cs
DevTKSS: @martincostello sorry, but I am not sure what your suggestion wants me to do exactly `let's just put this in the place it's used to avoid additional public API surface area we'd need to maintain`?
1. Delete the Extension and just add full, non extension claim into the Etsy.md docs?
2. EtsyAuthenticationExtensions (extension method I guess?)
3. something else?

martincostello:
Please resolve merge conflicts - the `.sln` file has been removed and replaced with an `.slnx` file.

DevTKSS:
@martincostello Resolved Merge Conflicts + cleaned commits up a bit
I would have squashed some more/reworded, but the sln->slnx migration crashes `git rebase -i` so hope it's okay like this?

AspNet.Security.OAuth.Providers.slnx
```diff
@@ -39,6 +39,9 @@
     <File Path=".vscode/launch.json" />
     <File Path=".vscode/tasks.json" />
   </Folder>
+  <Folder Name="/Solution Items/docs/assets/">
+    <File Path="docs/assets/Etsy-find-your-client_id.png" />
```
martincostello: Could you rename the file to the below please (lowercase, no `_`)?

```suggestion
    <File Path="docs/assets/etsy-find-your-client-id.png" />
```

docs/etsy.md
martincostello, line 1: Thanks for being so comprehensive, but I think there's a little too much detail here from a maintenance perspective.

A lot of the content here should be evident from reading ASP.NET Core OAuth documentation and Etsy's own documentation (e.g. the various setup steps needed in the Etsy developer portal).

Typically we just document the bare minimum required to set things up with a code snippet, highlighting any optional code ([example](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/docs/github.md)), point out anything that's unusual compared to other providers, and document the purpose of any non-optional settings that aren't part of the underlying `OAuthHandler` from ASP.NET Core (for example, we don't typically document `UsePkce`, `SaveTokens` or `CallbackPath` as they aren't our properties).

For example, mapping additional claims isn't specify to Etsy, it can be done with any of the providers.

docs/etsy.md
```diff
@@ -0,0 +1,430 @@
...
+  }
+}
+```
+
+> [!NOTE]
+> We recommend saving tokens (`SaveTokens = true`) to facilitate token refresh, so the user does not need to re-authenticate frequently.
+> [!NOTE]
+> If `IncludeDetailedUserInfo` is set to `true` and the scopes `shops_r` and `email_r` scopes are sufficient, you don't need to set additional scopes in `appsettings.json`, they are added automatically.
+> [!TIP]
+> We recommend using the `EtsyAuthenticationDefaults` class in your `.AddEtsy` call which contains the default endpoint URLs.
```
martincostello: I don't think we need these recommendations - people can make their own decisions on how they way to integrate things.

samples/Mvc.Client/Startup.cs
martincostello, line 1: Please revert these changes from your testing - we don't typically update the sample for every new provider we add.

samples/Mvc.Client/appsettings.json
```diff
@@ -14,5 +14,11 @@
   "Twitter": {
     "ClientId": "6XaCTaLbMqfj6ww3zvZ5g",
     "ClientSecret": "Il2eFzGIrYhz6BWjYhVXBPQSfZuS4xoHpSSyD9PI"
+  },
+  "Etsy": {
+    "ClientId": "client-id-from-user-secrets",
+    "ClientSecret": "client-secret-from-user-secrets",
+    "CallbackPath": "/etsy-signin",
+    "IncludeDetailedUserInfo": true
```
martincostello:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/AspNet.Security.OAuth.Etsy.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <Authors>Sonja Schweitzer</Authors>
+    <PackageTags>aspnetcore;authentication;etsy;oauth;security</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="JetBrains.Annotations" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
```
martincostello: Add new line to get rid of the ⛔

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationConstants.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        /// <summary>The claim type for the user's Etsy shop ID.</summary>
+        public static readonly string ShopId = "urn:etsy:shop_id";
+
+        /// <summary>The claim type for the user's profile image URL.</summary>
+        public static readonly string ImageUrl = "urn:etsy:image_url";
+    }
+
+    /// <summary>
+    /// Contains the <see href="https://developers.etsy.com/documentation/reference#section/Authentication/oauth2">Etsy OAuth Scopes</see> constants for Etsy authentication.
```
martincostello: As new ones might be added to the API in the future that aren't in the provider:

```suggestion
    /// Contains <see href="https://developers.etsy.com/documentation/reference#section/Authentication/oauth2">Etsy OAuth Scopes</see> constants for Etsy authentication.
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+        if (!response.IsSuccessStatusCode)
+        {
+            await Log.BasicUserInfoErrorAsync(Logger, response, Context.RequestAborted);
+            throw new HttpRequestException("An error occurred while retrieving basic user information from Etsy.");
+        }
+
+        using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+        var meRoot = payload.RootElement;
+
+        // Extract user_id and shop_id from the /me response
```
martincostello: `shop_id` is referenced in the comments but doesn't seem to be used anymore.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
martincostello: `DetailedUserInfoEndpoint` should have a value, so we can remove the condition and just use `Options.DetailedUserInfoEndpoint` - if someone messes with it and the request fails, that's on them.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+            // shops_r scope is required to access basic user info.
+            throw new ArgumentOutOfRangeException(nameof(Scope), string.Join(',', Scope), $"The '{Scopes.ShopsRead}' scope must be specified.");
+        }
+
+        if (IncludeDetailedUserInfo && !Scope.Contains(Scopes.EmailRead))
+        {
+            // EmailRead scope is required to access detailed user info when IncludeDetailedUserInfo is enabled.
+            // The post configure action should have added it at this stage, so we need to ensure it's present.
+            throw new ArgumentOutOfRangeException(nameof(Scope), string.Join(',', Scope), $"The '{Scopes.EmailRead}' scope must be specified when '{nameof(IncludeDetailedUserInfo)}' is enabled.");
+        }
```
martincostello: I think it's better to not have this validation - if Etsy changes some requirements on their side, then our logic won't match. The user can just inspect any error logs if they misconfigure things.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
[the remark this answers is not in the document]
martincostello: I think it would be better to have the extra claim, rather than trying to second-guess their back-end.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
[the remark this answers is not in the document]
martincostello: If it's redundant, then yep let's remove it.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
martincostello: Thanks - I think we can skip the image, and just describe it.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
martincostello: Rather than format it, it should just be a string.

As the URL needs to be computed at runtime from the options, I would have thought the analyzer shouldn't fire.

src/AspNet.Security.OAuth.Etsy/EtsyPostConfigureOptions.cs
martincostello: The OAuth handlers have built-in validation, and the claims mapping is unrelated to options validation. It just runs over the actions and maps what it finds.

docs/etsy.md
DevTKSS: @martincostello the etsy api has not very much information as they are only telling about JavaScript way to do this and even this is known to be outdated + incomplete. They did at least add the `/users/me` endpoint in the past, so the users will any longer have to inspect the raw html to get their shop ID 😅
What I am not sure about, can you tell me, if you know any reference, what the default paths in Cookie Auth are? I did not find them honestly. The `MVC.Client` Sample **does** specify them, but the ASP.NET Core Docs are not clear about this and never mapping them 🤔

Refering to the [`Minimal API quick Reference"](https://learn.microsoft.com/de-de/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-10.0#authorization) or [`Use cookie authentication without ASP.NET Core Identity`](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/cookie?view=aspnetcore-10.0), I did not find this information, so while I was trying to set the provider up in my own API, which uses Minimal API, I have no idea, if I now need to explicitly register it or if not need to 😅

Same for implementing Refresh Token 🤔
1. **Do** we need in ASP.NET Core to have some kind of [`TokenRefreshHandler` over the here shown claims actions](https://learn.microsoft.com/de-de/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-10.0) in the Consumer WebApi from your knowledge or does the Framework handle this automatically?
2. If it does not, but I potentially just didn't find what the Name of the Event or Method is, we would need to call and the [social Auth guide in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-10.0) (which would be those providers here too?) only tells about:
  1. "Apply Authorization" (using MVC, no minimal API) by only adding the attribute, no sample content at all so assuming we might be fine with calling this `ChallengeAsync` like the Mvc client does?
  2. Then [`Save Tokens` by defining a route `OnGetAsync()`](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-10.0#save-the-access-token) which I dont see in our sample (any additionals to be known for our providers repo I missed?) 🤔
  3. And then it's just telling about [Sign Out](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-10.0#sign-out) which endpoint we also have in our sample, **BUT** if I click on the Logout Button in it the comment would tell me, the cookie with the `access_token` or `refresh_token` we eventually automatically stored, does get deleted (?) so I should be asked to approve sign in of this app again if I click sign in, and not have to delete the cookies for Etsy before it prompts me again 🤔
  4. **BUT** their [`SocialSample` does define such route](https://github.com/dotnet/aspnetcore/blob/13524176e22eaba6bc74d9d052ed91f629f6613b/src/Security/Authentication/samples/SocialSample/Startup.cs#L273) 👀

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello yes it got linted as variable that was assigned but never used. Will adjust the comment, if we even need one 👍

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationDefaults.cs
DevTKSS: @martincostello okay, adjusting it to just be a string and concatinate it with the gotten user_id in the Handler then 👍

docs/etsy.md
DevTKSS: @martincostello updated guide, would you check if this fits now?

martincostello: Could you update the headings to match the other documents please? (Example, Required Additional Settings, Optional Settings).

docs/etsy.md
```diff
@@ -0,0 +1,53 @@
...
+
+- You can obtain the Client ID (`keystring`) for your app by registering your application on [Etsy's developer portal](https://www.etsy.com/developers/your-apps).
+- The ClientSecret (`shared secret` in the Etsy app details) is optional for public clients using PKCE.
+
+## Provider Features
+
+| Feature | Description | Default |
+|:--|:--|:--|
+| `EtsyAuthenticationConstants.Scopes.*` | Contains [Etsy OAuth Scopes](https://developers.etsy.com/documentation/reference#section/Authentication/oauth2) constants for Etsy authentication. | N/A |
+| `EtsyAuthenticationConstants.Claims.*` | Contains claim type constants for both Etsy authentication endpoints. `getMe` and `getUser`. | N/A |
```
martincostello:
```suggestion
| `EtsyAuthenticationConstants.Claims.*` | Contains claim type constants for both Etsy authentication endpoints, `getMe` and `getUser`. | N/A |
```

docs/etsy.md
```diff
@@ -0,0 +1,53 @@
...
+| `EtsyAuthenticationExtensions.MapImageClaim()` | Extension method to map the `image_url_75x75` claim to `EtsyAuthenticationConstants.Claims.ImageUrl`. | N/A |
+| `EtsyAuthenticationDefaults.IncludeDetailedUserInfo` | Fetch extended profile data with auto-mapped claims (Email, GivenName, Surname). | `false` |
```
martincostello:
```suggestion
| `EtsyAuthenticationDefaults.IncludeDetailedUserInfo` | Fetch extended profile data with auto-mapped claims. | `false` |
```

docs/etsy.md
```diff
@@ -0,0 +1,53 @@
...
+| `EtsyAuthenticationDefaults.CallbackPath` | The default callback path used by the Etsy authentication handler. | `/signin-etsy` |
```
martincostello: This isn't our setting, it's inherited, so we don't document them.

docs/etsy.md
```diff
@@ -0,0 +1,53 @@
...
+    options.ClaimActions.MapImageClaim(); // Optional Extension to map the image_url_75x75 claim, will not be mapped automatically
+  });
+```
+
+- You can obtain the Client ID (`keystring`) for your app by registering your application on [Etsy's developer portal](https://www.etsy.com/developers/your-apps).
+- The ClientSecret (`shared secret` in the Etsy app details) is optional for public clients using PKCE.
+
+## Provider Features
+
+| Feature | Description | Default |
```
martincostello: Please use the same columns as the other documents.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            await Log.BasicUserInfoErrorAsync(Logger, response, Context.RequestAborted);
+            throw new HttpRequestException("An error occurred while retrieving basic user information from Etsy.");
+        }
+
+        using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+        var meRoot = payload.RootElement;
+
+        // Extract user_id from the /me response required to get detailed user info. shop_id is mapped later via ClaimActions
+        var userId = meRoot.GetProperty("user_id").GetInt64();
```
martincostello: Seems like this only needs to be done if `Options.IncludeDetailedUserInfo` is `true`.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// Retrieves detailed user information from Etsy.
+    /// </summary>
+    /// <param name="tokens">The OAuth token response.</param>
+    /// <param name="userId">The user ID to retrieve details for.</param>
+    /// <returns>A <see cref="JsonDocument"/> containing the detailed user information.</returns>
+    protected virtual async Task<JsonDocument> GetDetailedUserInfoAsync([NotNull] OAuthTokenResponse tokens, long userId)
+    {
+        var userDetailsUrl = $"{Options.DetailedUserInfoEndpoint}{userId}";
```
martincostello: Feels prudent to handle if `DetailedUserInfoEndpoint` doesn't end in a `/` here.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// from the <see href="https://developers.etsy.com/documentation/reference#operation/getUser">getUser</see> Endpoint.
+    /// </summary>
+    public bool IncludeDetailedUserInfo { get; set; }
+
+    /// <summary>
+    /// Gets or sets the endpoint used to retrieve detailed user information.
+    /// </summary>
+    /// <remarks>
+    /// The placeholder for <c>user_id</c> needs to be <c>"{0}"</c> and will be replaced with the authenticated user's ID.
+    /// </remarks>
```
martincostello: This is outdated.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+    public string? DetailedUserInfoEndpoint { get; set; }
```
martincostello: Initialize to `EtsyAuthenticationDefaults.DetailedUserInfoEndpoint` by default?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello not sure what you mean by "prudent" sorry 😅 do you mean that we need to add a validation like `.EndsWith('/')` or what do you suggest?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
DevTKSS: @martincostello yes it is, but if we need to run the ClaimActions in Line 64 before the conditional (could you check if we need to?) I would not exactly know how to do this, except from you want me to make another if statement for it?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
martincostello: If `DetailedUserInfoEndpoint` was overridden and the user forgot to add the trailing `/`, then the URL would be malformed.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
martincostello: Sorry I don't follow - the value is read and stored into a local variable and only used within that block, so if it isn't entered it's a redundant read.

Are you saying that if it wasn't read here but later that would cause a bug? If so, then that makes it seem like there's "magic" happening, which is itself a problem if the way the provider function isn't immediately clear from inspecting the code in the method.

DevTKSS: 😅 no it isn't doing magic from what I understood from the mappings. The `userId` is used in the detailedUserInfo block some lines below, to fetch this if requested. 🤔 
You remember the `/users/{user_id}`? This is what we need it for in the additional User details.

But it is possible that there is outdated code and the checks for the ClaimActions Types below is missing the userId which we do duplicate store in nameIdentity + Constant Map to `urn:Etsy:user_id` what I don't know is, if the `RunClaimActions` here before this is needed 🤔 if not than we might be able to only ask for the userId value once *inside* of the if statement

DevTKSS: 😅 no it isn't doing magic from what I understood from the mappings. The `userId` is used in the detailedUserInfo block some lines below, to fetch this if requested. 🤔 
You remember the `/users/{user_id}`? This is what we need it for in the additional User details.

But it is possible that there is outdated code and the checks for the ClaimActions Types below is missing the userId which we do duplicate store in nameIdentity + Constant Map to `urn:Etsy:user_id` what I don't know is, if the `RunClaimActions` here before this is needed 🤔 if not than we might be able to only ask for the userId value once *inside* of the if statement

martincostello: Sorry, I'm still not following your reply here.

The local variable `userId` is only used on line 69, which is after an `if`. Therefore the declaration and assignment of the variable can be moved from here to line 68.

```csharp
if (Options.IncludeDetailedUserInfo)
{
    var userId = meRoot.GetProperty("user_id").GetInt64();
    using var detailedPayload = await GetDetailedUserInfoAsync(tokens, userId);
}
```

DevTKSS: ah I think I found what you mean. 👍
moved this line into the if statement.
@martincostello Could you check, if we still need `context.RunClaimActions()` already before the if statement or if afterwards is also enough? I dont know how this behaves 🤔
I had initially a return inside of the if statement, but now the only return is at the end, so I would assume, it's enough at the end?

DevTKSS:
Hello @martincostello, hope you had great Christmas Days! Would you mind to check the last leftover Comment and let me know if there are any future open tasks to do in this PR?

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
martincostello: I don't remember off the top of my head. I would suggest stepping through it in a debugger to see what happens, but the code in your PR when I looked at it didn't appear to depend on any side effects from `RunClaimActions()` that may have run anyway, so the question is likely moot.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        ClaimsIssuer = EtsyAuthenticationDefaults.Issuer;
+        CallbackPath = EtsyAuthenticationDefaults.CallbackPath;
+
+        AuthorizationEndpoint = EtsyAuthenticationDefaults.AuthorizationEndpoint;
+        TokenEndpoint = EtsyAuthenticationDefaults.TokenEndpoint;
+        UserInformationEndpoint = EtsyAuthenticationDefaults.UserInformationEndpoint;
+
+        UsePkce = true;
+        SaveTokens = true;
```
martincostello, line 27: None of our existing providers enable this by default, so we shouldn't here either.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// </summary>
+    public string DetailedUserInfoEndpoint { get; set; } = EtsyAuthenticationDefaults.DetailedUserInfoEndpoint;
+
+    /// <inheritdoc />
+    public override void Validate()
+    {
+        if (IncludeDetailedUserInfo && !Scope.Contains(Scopes.EmailRead))
+        {
+            Scope.Add(Scopes.EmailRead);
+        }
```
martincostello, line 57: This isn't validation - it's "magic" turning on a misconfiguration.

This should just be removed, otherwise we're trying to duplicate Etsy's server-side validation, which could change at any time.

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// <param name="properties">The authentication properties.</param>
+    /// <param name="tokens">The OAuth token response containing the access token.</param>
+    /// <returns>An <see cref="AuthenticationTicket"/> containing the user claims and properties.</returns>
+    /// <exception cref="HttpRequestException">Thrown when an error occurs while retrieving user information from Etsy.</exception>
+    protected override async Task<AuthenticationTicket> CreateTicketAsync(
+        [NotNull] ClaimsIdentity identity,
+        [NotNull] AuthenticationProperties properties,
+        [NotNull] OAuthTokenResponse tokens)
+    {
+        // Get the basic user info (user_id and shop_id)
```
martincostello, line 41: I think the code is self-explanatory.

```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+            throw new HttpRequestException("An error occurred while retrieving basic user information from Etsy.");
+        }
+
+        using var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync(Context.RequestAborted));
+        var meRoot = payload.RootElement;
+
+        var principal = new ClaimsPrincipal(identity);
+        var context = new OAuthCreatingTicketContext(principal, properties, Context, Scheme, Options, Backchannel, tokens, meRoot);
+
+        // Map claims from the basic payload first
```
martincostello, line 60:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+        context.RunClaimActions();
+
+        // Optionally enrich with detailed user info if requested
```
martincostello, line 63:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+        if (Options.IncludeDetailedUserInfo)
+        {
+            // Extract user_id from the /me response
```
martincostello, line 66:
```suggestion
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+                // Skip the action if it's a JsonKeyClaimAction for user_id or shop_id
+                if (action is Microsoft.AspNetCore.Authentication.OAuth.Claims.JsonKeyClaimAction { ClaimType: var t } &&
+                                    (t == ClaimTypes.NameIdentifier
+                                  || t == EtsyAuthenticationConstants.Claims.ShopId))
+                {
+                    continue;
+                }
+
+                action.Run(detailedRoot, identity, Options.ClaimsIssuer ?? ClaimsIssuer);
+            }
```
martincostello, line 86: I wonder if there's a better way to deal with this as a claim action configured in the options, rather than relying on a implementation details (and possibly ignoring users' own claim actions for JSON keys they may have configured).

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationHandler.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// Retrieves detailed user information from Etsy.
+    /// </summary>
+    /// <param name="tokens">The OAuth token response.</param>
+    /// <param name="userId">The user ID to retrieve details for.</param>
+    /// <returns>A <see cref="JsonDocument"/> containing the detailed user information.</returns>
+    protected virtual async Task<JsonDocument> GetDetailedUserInfoAsync([NotNull] OAuthTokenResponse tokens, long userId)
+    {
+        var userDetailsUrl = Options.DetailedUserInfoEndpoint.EndsWith('/') ? $"{Options.DetailedUserInfoEndpoint}{userId}" : $"{Options.DetailedUserInfoEndpoint}/{userId}";
```
martincostello, line 101:
```suggestion
        var userDetailsUrl = $"{Options.DetailedUserInfoEndpoint.TrimEnd('/')}/{userId}";
```

src/AspNet.Security.OAuth.Etsy/EtsyAuthenticationOptions.cs
DevTKSS: @martincostello for the `SaveTokens = true`  I have no problem with removing this🤔 but if you say, that the Provider must be able to work with the Settings we are setting internally, we need to set `UsePkce = true`. Pkce is required for http Request for the Authorization Code in Etsy Auth flow.

martincostello: The comment only applies to that one line.

PR #968, opened 2024-10-30
Add support for Alipay certificate signing and openid.

BuknSS:
I need to call Alipay's "fund expenditure interface", and I have to change the key signature method to the certificate signature method, which will make the current Alipay Provider unavailable, so I refer to the [Alipay official SDK code](https://github.com/alipay/alipay-sdk-net-all/blob/1b7b73909954b107bddb6476dec68aafcc3f16e9/v2/AlipaySDKNet.Standard/Util/AntCertificationUtil.cs#L71), add some code, just add two certificate-related fields in the request content:
- app_cert_sn
- alipay_root_cert_sn

When options _AppCertPath_  and _RootCertPath_  are configured with the certificate file path, the certificate signing mode will be enabled

Alipay's certificate signature document:
- https://opendocs.alipay.com/common/057k53?pathHash=7b14a0af#%E8%AF%81%E4%B9%A6%E6%96%B9%E5%BC%8F

----

Alipay now only provides open_id to new merchants, without user_id. In this case, the following code will cause an error.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/1aaadb1fe28f3b9b3bc8e0405bf6746e900e4350/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs#L137

Alipay's openid document:
- https://opendocs.alipay.com/mini/0ai2i6

_Alipay documents are in Chinese, so they need to be translated._

----

Screenshot of network packet capture:
TokenEndpoint
Calling TokenEndpoint

UserInformationEndpoint
Calling UserInformationEndpoint

PR #1131, opened 2025-11-18
Add support for Alipay certificate signing

AigioL:
improve https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/968


src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -25,5 +25,16 @@ public static class Claims
         /// The user's gender. F: Female; M: Male.
         /// </summary>
         public const string Gender = "urn:alipay:gender";
+
+        /// <summary>
+        /// OpenID is the unique identifier of Alipay users in the application dimension.
```
martincostello: What is an _"application dimension"_?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -25,5 +25,16 @@ public static class Claims
...
+        /// See https://opendocs.alipay.com/mini/0ai2i6
+        /// </summary>
+        public const string OpenId = "urn:alipay:open_id";
+
+        /// <summary>
+        /// Alipay user system internal identifier, will no longer be independently open in the future, and will be replaced by OpenID.
```
martincostello: > will no longer be independently open in the future

I don't understand what this means.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -44,6 +45,21 @@ protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
...
+    private const string SignType = "RSA2";
+
+    private async Task AddCertSignatureParametersAsync(SortedDictionary<string, string?> parameters)
+    {
+        ArgumentNullException.ThrowIfNull(Options.PrivateKey);
+        ArgumentNullException.ThrowIfNull(Options.AppCertSNKeyId);
+        ArgumentNullException.ThrowIfNull(Options.RootCertSNKeyId);
+
+        var app_cert_sn = await Options.PrivateKey(Options.AppCertSNKeyId, Context.RequestAborted);
+        var alipay_root_cert_sn = await Options.PrivateKey(Options.RootCertSNKeyId, Context.RequestAborted);
```
martincostello: Use `camelCase`.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -44,6 +45,21 @@ protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
         return base.HandleRemoteAuthenticateAsync();
     }
 
+    private const string SignType = "RSA2";
+
+    private async Task AddCertSignatureParametersAsync(SortedDictionary<string, string?> parameters)
```
martincostello:
```suggestion
    private async Task AddCertificateSignatureParametersAsync(SortedDictionary<string, string?> parameters)
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
         ClaimActions.MapJsonKey(Claims.Gender, "gender");
         ClaimActions.MapJsonKey(Claims.Nickname, "nick_name");
         ClaimActions.MapJsonKey(Claims.Province, "province");
+        ClaimActions.MapJsonKey(Claims.OpenId, "open_id");
+        ClaimActions.MapJsonKey(Claims.UserId, "user_id");
+    }
+
+    /// <summary>
+    /// Get or set a value indicating whether to use certificate mode for signature implementation.
```
martincostello:
```suggestion
    /// Gets or sets a value indicating whether to use certificate mode for signature implementation.
```

martincostello: "generation" rather than "implementation"?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
...
+    /// <para>https://opendocs.alipay.com/common/057k53?pathHash=e18d6f77#%E8%AF%81%E4%B9%A6%E6%A8%A1%E5%BC%8F</para>
+    /// </summary>
+    public bool EnableCertSignature { get; set; }
```
martincostello:
```suggestion
    public bool UseCertificateSignatures { get; set; }
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
...
+
+    /// <summary>
+    /// Gets or sets the optional ID for your Sign in with app_cert_sn.
+    /// </summary>
+    public string? AppCertSNKeyId { get; set; }
```
martincostello: Assuming that's what `SN` means here:

```suggestion
    public string? ApplicationCertificateSubjectNameKeyId { get; set; }
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
...
+    }
+
+    /// <summary>
+    /// Get or set a value indicating whether to use certificate mode for signature implementation.
+    /// <para>https://opendocs.alipay.com/common/057k53?pathHash=e18d6f77#%E8%AF%81%E4%B9%A6%E6%A8%A1%E5%BC%8F</para>
+    /// </summary>
+    public bool EnableCertSignature { get; set; }
+
+    /// <summary>
+    /// Gets or sets the optional ID for your Sign in with app_cert_sn.
```
martincostello: "Sign in with app_cert_sn" - this needs a better description.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
...
+    /// </summary>
+    public string? AppCertSNKeyId { get; set; }
+
+    /// <summary>
+    /// Gets or sets the optional ID for your Sign in with alipay_root_cert_sn.
```
martincostello: Same here.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,52 @@ public AlipayAuthenticationOptions()
...
+    /// </summary>
+    public string? RootCertSNKeyId { get; set; }
```
martincostello:
```suggestion
    public string? RootCertificateSubjectNameKeyId { get; set; }
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Extension methods to configure Sign in with Alipay authentication capabilities for an HTTP application pipeline.
+/// </summary>
+public static class AlipayAuthenticationOptionsExtensions
+{
+    /// <summary>
+    /// Configures the application to use a specified private to generate a client secret for the provider.
```
martincostello:
```suggestion
    /// Configures the application to use a specified private key to generate a client secret for the provider.
```

src/AspNet.Security.OAuth.Alipay/AntCertificationUtil.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Globalization;
+using System.Numerics;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace AspNet.Security.OAuth.Alipay;
+
+/// <summary>
+/// https://github.com/alipay/alipay-sdk-net-all/blob/master/v2/AlipaySDKNet.Standard/Util/AntCertificationUtil.cs
```
martincostello: Please update this to a permalink that points to a specific Git SHA.

src/AspNet.Security.OAuth.Alipay/AntCertificationUtil.cs
```diff
@@ -0,0 +1,140 @@
...
+/// </summary>
+internal static class AntCertificationUtil
```
martincostello: What is an `Ant`?

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,140 @@
...
+/// </summary>
+internal static class AntCertificationUtil
+{
+    public static string GetCertSN(ReadOnlySpan<char> certContent)
+    {
+        using var cert = X509Certificate2.CreateFromPem(certContent);
+        return GetCertSN(cert);
+    }
+
+    public static string GetCertSN(X509Certificate2 cert)
```
martincostello: Does this code need to be this performance critical (e.g. with `stackalloc`)? Feels like it could be implemented more simply to make it easier to read and maintain.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
AigioL: This refers to the fact that for a single app, the OpenId is unique. If you have multiple distinct app services, duplicate values may occur. In such cases, you need to use the unionid.
Additionally, newly registered apps no longer include a UserId. The official documentation recommends using OpenId as a replacement.

src/AspNet.Security.OAuth.Alipay/AntCertificationUtil.cs
AigioL: Ant is the abbreviation for Ant Financial, the company behind the Alipay app. Since the code for this class originates from Alipay's official repository, I believe it's best to retain the original name.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: This section of code rarely requires modification and generally remains unchanged unless issues arise. Naturally, higher performance is preferable, and it has passed unit testing in my production projects.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
martincostello: Please update the comment so that a future reader won't ask the same question. It should be apparent from the code what this means without requiring reading the documentation.

src/AspNet.Security.OAuth.Alipay/AntCertificationUtil.cs
martincostello: If we're just copying code and have a link to it's provenance (which should be a permalink not a link to a branch), and have already edited it to remove things not needed, then keeping the original name, particularly when "Ant" is meaningless to the code as it's just a company name, we can give it a better more meaningful name that matches its intended function.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
martincostello: > unchanged unless issues arise

Otherwise known as maintenance 😄

Simpler code is better - we have over a hundred providers to maintain here and I'd rather the code was easy to read.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
```diff
@@ -25,5 +25,16 @@ public static class Claims
...
+        /// <summary>
+        /// OpenID is the unique identifier for Alipay users at the application level.
+        /// See https://opendocs.alipay.com/mini/0ai2i6
+        /// </summary>
+        public const string OpenId = "urn:alipay:open_id";
+
+        /// <summary>
+        /// The internal identifier for Alipay users will no longer be independently available going forward and will be replaced by OpenID.
+        /// </summary>
+        public const string UserId = "urn:alipay:user_id";
```
martincostello, line 39: On reflection, if this is going to be deprecated, it would be better not to expose it as a `public` member. Otherwise it feels like we're adding a new API that's immediately obsolete.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -44,6 +44,21 @@ protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
...
+    {
+        ArgumentNullException.ThrowIfNull(Options.PrivateKey);
+        ArgumentNullException.ThrowIfNull(Options.ApplicationCertificateSnKeyId);
+        ArgumentNullException.ThrowIfNull(Options.RootCertificateSnKeyId);
+
+        var app_cert_sn = await Options.PrivateKey(Options.ApplicationCertificateSnKeyId, Context.RequestAborted);
+        var alipay_root_cert_sn = await Options.PrivateKey(Options.RootCertificateSnKeyId, Context.RequestAborted);
+
+        parameters["app_cert_sn"] = AlipayCertificationUtil.GetCertSN(app_cert_sn.Span);
+        parameters["alipay_root_cert_sn"] = AlipayCertificationUtil.GetRootCertSN(alipay_root_cert_sn.Span, SignType);
```
martincostello: `camelCase`.

```suggestion
        var appPrivateKey = await Options.PrivateKey(Options.ApplicationCertificateSnKeyId, Context.RequestAborted);
        var alipayPrivateKey = await Options.PrivateKey(Options.RootCertificateSnKeyId, Context.RequestAborted);

        parameters["app_cert_sn"] = AlipayCertificationUtil.GetCertSN(appPrivateKey.Span);
        parameters["alipay_root_cert_sn"] = AlipayCertificationUtil.GetRootCertSN(alipayPrivateKey.Span, SignType);
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -44,6 +44,21 @@ protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
...
+    private const string SignType = "RSA2";
+
+    private async Task AddCertificateSignatureParametersAsync(SortedDictionary<string, string?> parameters)
+    {
+        ArgumentNullException.ThrowIfNull(Options.PrivateKey);
+        ArgumentNullException.ThrowIfNull(Options.ApplicationCertificateSnKeyId);
+        ArgumentNullException.ThrowIfNull(Options.RootCertificateSnKeyId);
+
+        var app_cert_sn = await Options.PrivateKey(Options.ApplicationCertificateSnKeyId, Context.RequestAborted);
+        var alipay_root_cert_sn = await Options.PrivateKey(Options.RootCertificateSnKeyId, Context.RequestAborted);
```
martincostello: Is this actually the private key for a root certificate belonging to Alipay? That would be highly insecure in the hands of users, so maybe the naming just needs improving?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+ * for more information concerning the license and the contributors participating to this project.
+ */
+
+using AspNet.Security.OAuth.Alipay;
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Extension methods to configure Sign in with Alipay authentication capabilities for an HTTP application pipeline.
```
martincostello: Is it actually called _"Sign in with Alipay"_, or is this just copy-paste where "Apple" has been changed to "Alipay"?

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+/// </summary>
+public static class AlipayAuthenticationOptionsExtensions
+{
+    /// <summary>
+    /// Configures the application to use a specified private key to generate a client secret for the provider.
+    /// </summary>
+    /// <param name="options">The Apple authentication options to configure.</param>
```
martincostello:
```suggestion
    /// <param name="options">The Alipay authentication options to configure.</param>
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Extension methods to configure Sign in with Alipay authentication capabilities for an HTTP application pipeline.
+/// </summary>
+public static class AlipayAuthenticationOptionsExtensions
+{
+    /// <summary>
+    /// Configures the application to use a specified private key to generate a client secret for the provider.
```
martincostello:
```suggestion
    /// Configures the application to use a specified private key to generate a client secret for the provider when using certificate signatures.
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Globalization;
+using System.Numerics;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace AspNet.Security.OAuth.Alipay;
+
+/// <summary>
+/// https://github.com/alipay/alipay-sdk-net-all/blob/b482d75d322e740760f9230d2a3859090af642a7/v2/AlipaySDKNet.Standard/Util/AntCertificationUtil.cs
```
martincostello:
```suggestion
/// Based on https://github.com/alipay/alipay-sdk-net-all/blob/b482d75d322e740760f9230d2a3859090af642a7/v2/AlipaySDKNet.Standard/Util/AntCertificationUtil.cs.
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+/// </summary>
+internal static class AlipayCertificationUtil
+{
+    public static string GetCertSN(ReadOnlySpan<char> certContent)
```
martincostello: What actually _is_ the return value? `SN` suggests a serial number, but it actually appears to be the MD5 hash of the serial number.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        using var cert = X509Certificate2.CreateFromPem(certContent);
+        return GetCertSN(cert);
+    }
+
+    public static string GetCertSN(X509Certificate2 cert)
+    {
+        var issuerDN = cert.Issuer.Replace(", ", ",", StringComparison.InvariantCulture);
```
martincostello:
```suggestion
        var issuerDN = cert.Issuer.Replace(", ", ",", StringComparison.Ordinal);
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+/// </summary>
+internal static class AlipayCertificationUtil
+{
+    public static string GetCertSN(ReadOnlySpan<char> certContent)
+    {
+        using var cert = X509Certificate2.CreateFromPem(certContent);
+        return GetCertSN(cert);
+    }
+
+    public static string GetCertSN(X509Certificate2 cert)
```
martincostello:
```suggestion
    private static string GetCertSN(X509Certificate2 cert)
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private static IEnumerable<string> GetRootCertSNCore(X509Certificate2Collection x509Certificates, string signType)
+    {
+        foreach (X509Certificate2 cert in x509Certificates)
+        {
+            var signatureAlgorithm = cert.SignatureAlgorithm.Value;
+            if (signatureAlgorithm != null)
+            {
+                if ((signType.StartsWith("RSA", StringComparison.InvariantCultureIgnoreCase) &&
+                    signatureAlgorithm.StartsWith("1.2.840.113549.1.1", StringComparison.InvariantCultureIgnoreCase)) ||
```
martincostello: Based on [this search](https://github.com/search?q=repo%3Adotnet%2Fruntime%20%221.2.840.113549.1.1%22&type=code), `"1.2.840.113549.1.1."` seems more correct?

Otherwise would it pick up values in `11-19`, `100-199`, `1000-1999` etc. that it shouldn't?

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+                    (signType.StartsWith("SM2", StringComparison.InvariantCultureIgnoreCase) &&
+                    signatureAlgorithm.StartsWith("1.2.156.10197.1.501", StringComparison.InvariantCultureIgnoreCase)))
```
martincostello: Similarly, `"1.2.156.10197.1.501."` for a value from the "SM2 Signing with SM3" family?

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
martincostello:
```suggestion
                    signatureAlgorithm.StartsWith("1.2.840.113549.1.1", StringComparison.OrdinalIgnoreCase)) ||
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
martincostello:
```suggestion
                    signatureAlgorithm.StartsWith("1.2.156.10197.1.501", StringComparison.OrdinalCultureIgnoreCase)))
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    public static string GetRootCertSN(ReadOnlySpan<char> rootCertContent, string signType = "RSA2")
+    {
+        var rootCertSN = string.Join('_', GetRootCertSNCore(rootCertContent, signType));
+        return rootCertSN;
+    }
+
+    private static IEnumerable<string> GetRootCertSNCore(X509Certificate2Collection x509Certificates, string signType)
+    {
+        foreach (X509Certificate2 cert in x509Certificates)
```
martincostello:
```suggestion
    private static IEnumerable<string> GetRootCertSNCore(X509Certificate2Collection certificates, string signType)
    {
        foreach (X509Certificate2 cert in certificates)
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+    }
+
+    private static IEnumerable<string> GetRootCertSNCore(ReadOnlySpan<char> rootCertContent, string signType)
+    {
+        X509Certificate2Collection x509Certificates = [];
+        x509Certificates.ImportFromPem(rootCertContent);
+        return GetRootCertSNCore(x509Certificates, signType);
```
martincostello:
```suggestion
        var certificates = new X509Certificate2Collection();
        x509Certificates.ImportFromPem(certificates);
        return GetRootCertSNCore(certificates, signType);
```

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+        return CalculateMd5(string.Join(',', attributes) + serialNumber);
+    }
+
+    public static string GetRootCertSN(ReadOnlySpan<char> rootCertContent, string signType = "RSA2")
+    {
+        var rootCertSN = string.Join('_', GetRootCertSNCore(rootCertContent, signType));
+        return rootCertSN;
+    }
+
+    private static IEnumerable<string> GetRootCertSNCore(X509Certificate2Collection x509Certificates, string signType)
```
martincostello: The `Core` suffix feels redundant.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// <summary>
+    /// https://github.com/dotnet/runtime/blob/v9.0.8/src/libraries/System.Text.Json/Common/JsonConstants.cs#L12
+    /// </summary>
+    private const int StackallocByteThreshold = 256;
+
+    private static string CalculateMd5(ReadOnlySpan<char> chars)
+    {
+        var lenU8 = Encoding.UTF8.GetMaxByteCount(chars.Length);
```
martincostello: `length`

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+            stackalloc byte[StackallocByteThreshold] :
+            (array = ArrayPool<byte>.Shared.Rent(lenU8));
+        try
+        {
+            Encoding.UTF8.TryGetBytes(chars, bytes, out var bytesWritten);
+            bytes = bytes[..bytesWritten];
+
+            Span<byte> hash = stackalloc byte[MD5.HashSizeInBytes];
+#pragma warning disable CA5351
+            MD5.HashData(bytes, hash);
```
martincostello: The return value should be checked.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
```diff
@@ -0,0 +1,105 @@
...
+        x509Certificates.ImportFromPem(rootCertContent);
+        return GetRootCertSNCore(x509Certificates, signType);
+    }
+
+    /// <summary>
+    /// https://github.com/dotnet/runtime/blob/v9.0.8/src/libraries/System.Text.Json/Common/JsonConstants.cs#L12
+    /// </summary>
+    private const int StackallocByteThreshold = 256;
+
+    private static string CalculateMd5(ReadOnlySpan<char> chars)
```
martincostello: I think this is still overcomplicated really. This isn't a hot path.

The whole thing could just be:

```csharp
var buffer = Encoding.UTF8.GetBytes(chars);
return MD5.HashData(buffer);
```

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
AigioL: UserId cannot be modified to ```private``` because applications created before OpenId or legacy code still require it. For newly created applications, OpenId should be used unless a support ticket is submitted to request the platform to enable legacy UserId. I have updated the code and added the ```ObsoleteAttribute```.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
AigioL: The two certificate PEM files here are both public keys

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
AigioL: is correct, see https://auth.alipay.com/login/global.htm
image

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: ```SN``` is not merely a serial number, but rather the MD5 value of Issuer + SerialNumber or Reverse. The final output is a lowercase hex MD5 string. Official documentation provides no additional explanation; the naming convention ```SN``` in Java code originates from the official C# code, hence the retention of the original name.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: Yes, the certificate signature algorithm OID 1.2.156.10197.1.501 indicates that this certificate uses the SM2 signature algorithm based on SM3. SM2 certificate data, similar to RSA algorithm certificates, includes the certificate version, serial number, issuer, subject information, subject public key, validity period, certificate extensions, etc. However, the public key algorithm for SM2 certificates uses the OID identifier for the ECC algorithm, while the public key parameters use the OID identifier for the SM2 national cryptographic algorithm.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: Theoretically, that's correct, but I suspect no other numbers would match. The certificate file is downloaded from Alipay's backend platform, and its Java code operates in the same manner.

https://github.com/alipay/alipay-sdk-java-all/blob/0d2915e22649e4fc178ba204e106b57e78b33316/v3/src/main/java/com/alipay/v3/util/AntCertificationUtil.java#L270-L272

https://github.com/alipay/alipay-sdk-java-all/blob/02f65a40c5c5a4b317a4da303585cdca4a6d8f3e/v2/src/main/java/com/alipay/api/internal/util/AntCertificationUtil.java#L286-L288

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: The MD5 algorithm has a fixed length, so there is no need to check the returned result. If the input array is too short, it will throw an exception.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: I have balanced high performance with code conciseness. The return value here should be a lowercase hex string. The updated code uses `stackalloc` instead of `byte[]` to avoid fragmented memory.

martincostello: _Technically_ yes, but I don't think it's of importance here against a simpler implementation.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
martincostello: Adding a new public API that is immediately obsolete serves no useful purpose to the user and adds maintenance burden for us to remove it in the future (which is then itself a breaking change).

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
martincostello: So it's called "Sign In Alipay", not "Sign In **with** Alipay"?

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
martincostello: So the name is wrong then, because it doesn't reflect what it is. I don't think the original Alipay code is particularly good C#, so saying "the official code is like that" is not a justification to push back on any of the comments I'm giving on the code here, as ultimately if this PR is accepted and merged it becomes ours to maintain.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
martincostello: That just sounds to me like both their C# and Java implementations are wrong and over-eagerly matching.

martincostello:
Please don't resolve comments where no code changes have been made - otherwise it makes things harder to review as I have to uncollapse the comments in the UI, and if I disagree I just have to unresolve them anyway.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,56 @@ public AlipayAuthenticationOptions()
         ClaimActions.MapJsonKey(Claims.Gender, "gender");
         ClaimActions.MapJsonKey(Claims.Nickname, "nick_name");
         ClaimActions.MapJsonKey(Claims.Province, "province");
+        ClaimActions.MapJsonKey(Claims.OpenId, "open_id");
+#pragma warning disable CS0618
+        ClaimActions.MapJsonKey(Claims.UserId, "user_id");
+#pragma warning restore CS0618
```
martincostello: If it's going to be obsolete, just hard-code the value here without the need for the constant. Then we can just delete it in the future without having to also change the public API surface.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
```diff
@@ -29,5 +29,56 @@ public AlipayAuthenticationOptions()
...
+    /// the value of the <see cref="ApplicationCertificateSnKeyId"/> or <see cref="RootCertificateSnKeyId"/> property and the <see cref="CancellationToken"/>
+    /// associated with the current HTTP request.
+    /// </summary>
+    /// <remarks>
+    /// The public key should be in PKCS #8 (<c>.p8</c>) format.
+    /// </remarks>
+    public Func<string, CancellationToken, Task<ReadOnlyMemory<char>>>? PublicKey { get; set; }
+
+    /// <inheritdoc />
+    public override void Validate()
```
martincostello, line 57: Shouldn't this also validate that `PublicKey` is not null?

aicodeguard:
### Code Review – Key Findings

**Summary:**
The review identified two significant issues: a resource leak in the certificate utility helper and a performance bottleneck in the request handling logic. Both issues affect the operational stability and efficiency of the provider under load.

## Issues

1. **[Performance]** **Redundant certificate parsing on every request**
   - **File:** `src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs`
   - **Location:** [Lines 49-60](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/59de1615f13774ec5c94ac85d997b345b09184e3/src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs#L49-L60)
   - **Details:** The `AddCertificateSignatureParametersAsync` method invokes the `PublicKey` delegate (which performs file I/O in the default extension) and parses the PEM content on every authentication attempt. Since these certificates are static configuration, this creates unnecessary latency and CPU usage.
   - **Fix:** Compute and cache the `app_cert_sn` and `alipay_root_cert_sn` values once (e.g., during options validation or lazily on the first request) rather than re-computing them for every transaction.

2. **[Maintainability]** **Resource leak in certificate collection handling**
   - **File:** `src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs`
   - **Location:** [Lines 43-46](https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/59de1615f13774ec5c94ac85d997b345b09184e3/src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs#L43-L46)
   - **Details:** The `GetRootCertSN` method creates an `X509Certificate2Collection` and populates it via `ImportFromPem`. While the collection itself is not disposable, the `X509Certificate2` instances it contains hold unmanaged resources and must be disposed. Failing to do so causes a handle leak.
   - **Fix:** Ensure that all certificates within the collection are disposed, for example by iterating over them in a `finally` block after the SN generation is complete.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
AigioL: This should be “Sign in with Alipay” because, like Apple, it uses a third-party platform account for login. This will redirect you to an Alipay webpage displaying a QR code. Use the Alipay mobile app to scan the code and log in.

src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs
AigioL: The latest submission has been simplified.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/13e690ee0b4ccc041551c61ed3170a9bc58a2972/src/AspNet.Security.OAuth.Alipay/AlipayCertificationUtil.cs#L67-L75

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
AigioL: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1131/commits/13e690ee0b4ccc041551c61ed3170a9bc58a2972

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptions.cs
AigioL: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1131/commits/13e690ee0b4ccc041551c61ed3170a9bc58a2972

AigioL:
Based on aicodeguard's comment, there is indeed no need to parse the PEM certificate file with every request. I removed the certificate-related content and simplified all changes by using fixed configuration values.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/1131/commits/fe4159d46116fc918799eec9b5c67561735417f5

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationConstants.cs
martincostello: My intention was to delete the property entirely, not just to remove the `[Obsolete]` attribute.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationOptionsExtensions.cs
martincostello: "Sign **I**n with Apple" is the name of the product ([docs](https://support.apple.com/en-gb/102609)), "Sign **i**n with Apple" is the description of what the user does.

"Sign In Alipay" appears to be the name, as it is not grammatically correct English to be a description of what the user is doing.

src/AspNet.Security.OAuth.Alipay/AlipayAuthenticationHandler.cs
```diff
@@ -44,6 +44,16 @@ protected override Task<HandleRequestResult> HandleRemoteAuthenticateAsync()
         return base.HandleRemoteAuthenticateAsync();
     }
 
+    private const string SignType = "RSA2";
+
+    private void AddCertificateSignatureParameters(SortedDictionary<string, string?> parameters)
+    {
+        ArgumentNullException.ThrowIfNull(Options.ApplicationCertificateSn);
+        ArgumentNullException.ThrowIfNull(Options.RootCertificateSn);
```
martincostello, line 52: These aren't arguments to the method. They were already validated in the options. If they are subsequently messed with by the integrator that's user error.

```suggestion
```

